PR #33, opened 2015-02-18
Set .gitignore and .gitattributes

ngbrown:
Set `.gitattributes` to normalize most file endings with auto (except `.csproj` and `.sln`, which are crlf).  This also includes the commit which fixes all the files that had mixed line endings or other issue.


michaelklishin:
Thank you!

PR #34, opened 2015-02-19
Clarify that ConnectionFactory.RequestedHeartbeat sets a timeout

michaelklishin:
…and not a communication interval.


michaelklishin:
For @simonmacmullen to QA.

PR #37, opened 2015-03-03
Make consumer operation dispatch concurrent

michaelklishin:
This ports `ConsumerDispatcher` from the Java client. It offers per-channel operation ordering while making
cross-channel dispatching concurrent.

It is also possible to provide a custom [TaskScheduler](https://msdn.microsoft.com/en-us/library/system.threading.tasks.taskscheduler(v=vs.100).aspx) to make dispatch sequential.

Will merge once documentation guide is updated and we have some feedback from downstream projects such as [EasyNetQ](http://easynetq.com/).


michaelklishin:
The docs are updated, this is ready to go.

Issue #32, opened 2015-02-18
Build nuget packages from this repository.


ngbrown:
For years, @joefitzgerald has been taking the official dll's and packaging them into nuget packages from a separate repository ([link](https://github.com/joefitzgerald/rabbitmq-dotnet-client-package)).  The .nuspec and build scripts should be moved to this repository so that future merges to support additional platforms (Windows Store and portable apps) can appropriately adjust the nuget generation.

michaelklishin:
Agreed.

joefitzgerald:
:+1:

ngbrown:
Is there an updated document that outlines the entire process of using the various build scripts with what tools need to be installed?  The NuGet execution needs to be somewhere in those build scripts.

For example, the public-umbrella [build instructions](https://github.com/rabbitmq/rabbitmq-public-umbrella/tree/master/release-build) mentions <http://moshi.lshift.net/RabbitMQDotnetClientReleaseProcess> but that is a dead link.

Is all the mono build switching still relevant? It seems that building on Windows and then testing on both Windows and Mono would be acceptable.

michaelklishin:
@ngbrown see http://www.rabbitmq.com/build-dotnet-client.html. Building on Windows for NuGet makes sense.

michaelklishin:
@ngbrown I've updated the pages. VS 2013, .NET 4 and `cp Local.props.example Local.props` is all you need these days. Then simply run `msbuild.exe`.

michaelklishin:
I'd really like to see NuGet scripts to be in this repo. That's the easiest way to automated the rest for us and do manual releases as needed (e.g. to publish preview releases).

joefitzgerald:
Perhaps a first easy step is to set up a CI build on AppVeyor (https://appveyor.com) against the public repo for the .NET client. From there, we start by publishing snapshot releases. Once everyone's comfortable with it, we could publish the actual releases once you're comfortable with the process.

joefitzgerald:
@michaelklishin Can you give me commit rights to this repo so I can enable the builds on AppVeyor? Not a commitment to use it, but I'll push a branch and create a PR against it to review how this would look.

michaelklishin:
@joefitzgerald you should have an invitation by now.

ngbrown:
I did get a build running in appveyor <https://ci.appveyor.com/project/ngbrown/rabbitmq-dotnet-client/history>.

[...]

PR #38, opened 2015-03-04
Changes to run tests on single-core machine and adds nuspec files.

ngbrown:
This pull request partially addresses #32. There are not any changes to the build scripts and this pull request is the basis for any direction taken with incorporating nuget outputs into the build.

I think this can be merged early (with or without the nuspec files) to allow for all public continuous integration builds and local developer tests to start taking place on all future branches.

### DIFF
--- a/RabbitMQ.Client.nuspec
+++ b/RabbitMQ.Client.nuspec
@@ -1,0 +1,25 @@
+<?xml version="1.0"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+  <metadata>
+    <id>RabbitMQ.Client</id>
+    <version>$version$</version>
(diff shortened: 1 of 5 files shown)


michaelklishin:
Thank you, we'll get to it tomorrow.

RabbitMQ.Client.nuspec
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+  <metadata>
+    <id>RabbitMQ.Client</id>
+    <version>$version$</version>
+    <title>RabbitMQ.Client</title>
+    <authors>Pivotal,joefitzgerald</authors>
```
joefitzgerald: You can remove me as an author!

PR #41, opened 2015-03-12
Execute tasks on the provided TaskSchduler

### LINKED ISSUE #40: ConsumerWorkService doesn't use the scheduler it is constructed wtih

michaelklishin:
Fixes #40.

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/impl/ConsumerWorkService.cs
+++ b/projects/client/RabbitMQ.Client/src/client/impl/ConsumerWorkService.cs
@@ -18,6 +18,7 @@ namespace RabbitMQ.Client
 
         public ConsumerWorkService(TaskScheduler scheduler)
         {
+            this.scheduler = scheduler;
             this.workPool = new BatchingWorkPool<IModel, Action>();
         }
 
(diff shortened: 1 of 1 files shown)


kae:
https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/e01343344f58bb15967472b0051a0b6878c8930e/projects/client/RabbitMQ.Client/src/client/impl/ConsumerWorkService.cs#L66

Does this Task.Start intentionally called without scheduler?

michaelklishin:
@kae sort of. I'm not sure if it would be really useful to use a custom scheduler at that particular call site,
when executing thunks (batches of operations).

In the Java client we do this using an executor but primarily because there is no "default" executor on the JVM. Do you think it would make a real difference for you? If so, feel free to submit a pull request.

kae:
@michaelklishin definetely not for me.
I kind of like concurrent dispatching.

It just looks kind of suspicious.
And if someone will want really synchronous solution, current "half-synced" solution may feel strange.

michaelklishin:
It is not "half synchronous". Individual operations are performed concurrently if task scheduler allows for that. You can profile how much time it takes to submit a batch for execution: it's nothing compared to executing consumer methods that do actual work.

"Suspicious" is not objective, backed by data or something that can be acted on. Profiling results are. Pull requests are.

PR #44, opened 2015-03-14
Revert client authentication changes in 36efc00

kingpong:
The change in 36efc00 broke the client's ability to require specific version(s) of SSL. This pull request contains a test that fails when the feature is broken, and a revert commit to fix it.

This seems to work correctly whether I am presenting a client certificate or not. I can't speak for Mono, but everything works fine this way on Windows.


michaelklishin:
This makes it impossible (again) to not provide a client certificate. I suspect we should provide an empty `X509CertificateCollection` instead of using `AuthenticateAsClient/1`.

michaelklishin:
Thank you. I had to add a couple of `using` statements to make it work on Mono.

dumbbell:
Commits fixing this issue were cherry-picked to `stable`.

PR #54, opened 2015-04-06
Ensure body is cleared after we return encoded content

### LINKED ISSUE #53: MapMessageBuilder.GetContentBody() throws NullReferenceException
Looks like the Body is being set to null before actually writing the contents.

```c#
public override byte[] GetContentBody()
{
	this.Body = null;
	MapWireFormatting.WriteMap(base.Writer, this.Body);
	return base.GetContentBody();
}
```

michaelklishin:
Fixes #53.


michaelklishin:
Submitted against the wrong branch.

PR #65, opened 2015-04-19
Added WinRT (Windows 8) compatible library, sharing code where possible.

ngbrown:
Includes NUnit driven tests.
Includes Windows Store example application.
Includes netcore45 addition to NuGet nuspec file.
Addresses GitHub issue #16.

The work I did toward #32 with a public CI ran into some troubles with WinRT Tasks usage and the single CPU test environment.  Some unit tests work fine on multi-processor environment, but deadlock on a single CPU core environment (can be simulated with CPU affinity).  I think by pushing this to the main branch, more people can work on this and it can be figured out.

PR #66, opened 2015-04-20
Added WinRT (Windows 8) compatible library, sharing code where possible

michaelklishin:
This is #65 by @ngbrown, re-submitted to follow our branch naming convention. I intend to push a few build system fixes to disable building WinRT client on Mono and make it possible to disable it on Windows, e.g. versions prior to 8.

`effort-high` because this is a substantial contribution (my contribution is `effort-tiny`).


michaelklishin:
@ngbrown I'm now getting

```
Build action 'Page' is not supported by the specific combination of the project's targets
```

What do I have to configure to get rid of it?

ngbrown:
@michaelklishin, I get the same errors with your changes.  Bisecting back to a7f371b18339a7a4a48c17c2dde1bc5478b93cb8, the merge, it doesn't have the problem.

michaelklishin:
@ngbrown yes, it was a confusing error message, the problem was unrelated to the target. However, I still could not make the solution build without re-targeting. I have Windows 8 SDK installed, tried rebooting, inspecting project files manually, and so on.

We'll go with VS 2013 CE and Windows 8.1 as a requirement for now. RabbitMQ 3.6.0 will ship in September-October. If Windows 8 support is a big enough deal, we can revisit this decision later.

michaelklishin:
Build on Mono works now. Windows 7 doesn't seem to be an option for building Windows Store apps, so I'll have to test that WinRT is properly disabled in the build system there.

Otherwise this seems to be ready to merge.

ngbrown:
@michaelklishin With the most recent changes you've pushed, `RabbitMQ.Client.Examples.WinRT.Subscriber` still has the Page errors.  Something about the way the `.csproj` files was changed is making the Visual Studio designer think it is a WPF project.

More seriously, even though it compiles, RabbitMQ.Client.Examples.WinRT.Subscriber no longer deploys and launches.  It is not building the AppX container.

Is there any way that the build process can specify the target projects to include or exclude instead of  these conditionals in the individual `.csproj` files?

michaelklishin:
We can look into adding some Make targets that perform platform-specific builds a bit later. I'll investigate what's up with the project files.

michaelklishin:
@ngbrown I've started from your revision. Re-targeted for 8.1, getting

```
Named resource 'Files\..\..\..\client\RabbitMQ.Client.WinRT\build\bin\RabbitMQ.Client.pri' is outside project root! Only resource containers are allowed outside the project root.
```

My goal is to make sure Windows stuff works first, then add condtions for Mono.

ngbrown:
@michaelklishin I see that you changed RabbitMQ.Client.Unit.WinRT.csproj to link to the non-winrt client.  See 3489419fa848dfff578d53f9f07431f941d1fe88 for the minimum changes for a re-target to win 8.1.

ngbrown:
@michaelklishin, you mentioned the Windows 8 SDK, did you install it with this screen from the Visual Studio 2013 ISO/Installer?

screenshot - 2015-04-19 10_35_30 pm

This is all I have to enable this on my Windows 8.1 computer: "Tools for Maintaining Store Apps for Windows 8".  I don't have Visual Studio 2012.

Another possibility would be to close the solution (and make sure the example client .exe is killed) and run `git clean -d -x` wipe all the temporary files and then load in Visual Studio again.

michaelklishin:
I have both Tools for Maintaining Store Apps for Windows 8 *and* Windows Phone 8.0 SDK installed.

michaelklishin:
Uninstalling Windows Phone 8.0 SDK didn't help. I ended up symlinking `v12.0` in `WindowsXaml` to `v11.0` and now things build.

Attempts to deploy the example app now fail with

```
Error : DEP0700 : Registration of the app failed. error 0x80070005: Failed to set access rights to \\?\C:\cygwin64\home\admin\git\umbrella.git\rabbitmq-dotnet-client\projects\examples\WinRT\Subscriber\build\bin\Debug\AppX. (0x80070005)
```

ngbrown:
See the comments to the answer here: http://stackoverflow.com/a/13163076/25182 This is a known cygwin issue.

Is there a way to chat or something more real-time?  I think I'm in #rabbitmq.

michaelklishin:
Changing build output directory to a random directory outside of Cygin's home indeed helps. The app deploys successfully. I'm not sure how to run it in the emulator but looks like Windows is good and we can move on to Mono.

michaelklishin:
After an IRC discussion and multiple resets we have retained Windows 8.0 compatibility. Excluding Mono  is also done, need to investigate why msbuild.exe fails to build from scratch while VS doesn't.

michaelklishin:
RabbitMQ.Client.Unit.WinRT fails to build with `msbuild.exe` (cannot find RabbitMQ client classes, suggests a missing assembly reference).

michaelklishin:
@ngbrown OK, now things build in VS, with msbuild.exe and xbuild on Mono. Unit.WinRT passes all tests.

I'm ready to merge as soon as you confirm that things work for you just as well and the example app deploys and runs.

michaelklishin:
I felt confident (and excited) enough to merge.

PR #67, opened 2015-04-22
Fixes issue with closed merge #66

ngbrown:
Change Build,Rebuild,Clean Target behavior through changing DependsOn properties instead of replacing Targets.

@michaelklishin Can you check out that this works with Mono?


michaelklishin:
This is a good idea. I'll merge and add the necessary conditionals. I'm also open to considering not supporting solution-wide builds on Mono because we only care about the "regular" client, ApiGen, and .NET test suite working there.

michaelklishin:
@ngbrown it does seem to work well for Mono/xbuild. We're looking into our nightly builds, otherwise I'm OK with keeping the build system as is in master.

Thank you!

michaelklishin:
Unfortunately this does not work on Windows 7.

PR #71, opened 2015-04-24
Rework heartbeat handling, fixes #68

### LINKED ISSUE #68: 3.4.4 -> 3.5.1 Timeout
I have a small wrapper using this library on .NET 4.5.
I've updated from 3.4.4 to 3.5.1 and now, the connection gets closed and it crashes whenver I try to awknowledge a message:
*AMQP close-reason, initiated by Library, code=0, text="End of stream"*

I've found that this only happens when processing the message takes more than the current `RequestHeartbeat`. So if I increase it, it `Ack()` fine, otherwise it closes the connection and if throws.

Has anything changed related to this or am I using `RequestHeartbeat` in the wrong way?

michaelklishin:
With this change .NET client uses the same approach as the Java
one.

Previously in 3.5.x we've used two timers that
re-scheduled themselves and used auto-reset events to be
notified about connection activity on the I/O loop.

This turned out to have some issues:

 * It was clever but fairly complicated
 * Picking the wrong heartbeat interval value led to sporadic
   failures due to race conditions [on the wire]
 * It didn't play well with socket timeouts

Our new strategy is more straightforward:
heartbeat writer uses a periodic timer that always sends
a frame, regardless of recent activity, every 1/2nd a timeout.

Given that heartbeat timeouts are often in minutes, this is not
unreasonable but possibly the most straightforward implementation
possible. It also doesn't suffer from race conditions on the wire
because we send a heartbeat frame every half the interval, so
slight inaccuracy in .NET runtime scheduling of timers won't
get in our way with reasonable (> 1 second) heartbeat intervals.

Heartbeat "reader" now relies on the socket receive timeout,
just like in the Java client. This is more straightforward
and there is no dissonance between I/O timeouts and the one we
use for incoming heartbeats.

Various cases when a SocketException could indicate a problem
other than a timeout (usually early in the connection
lifecycle) are accounted for.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -87,4 +87,10 @@ Backup*/
 UpgradeLog*.XML
 UpgradeLog*.htm
 
+# Unit tests
+
 projects/client/Unit/TestResult.xml
+
+# Development scripts
+
+*.fsx
--- a/projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs
+++ b/projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs
@@ -38,13 +38,13 @@
 //  Copyright (c) 2007-2014 GoPivotal, Inc.  All rights reserved.
 //---------------------------------------------------------------------------
 
+using RabbitMQ.Client.Exceptions;
+using RabbitMQ.Client.Framing.Impl;
+using RabbitMQ.Client.Impl;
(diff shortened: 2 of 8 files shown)


michaelklishin:
Seems to fix #68 according to my own tests as well as user feedback. Merging.

Issue #82, opened 2015-05-03
Switch heartbeat implementation back to reset events


michaelklishin:
See comments in #72. To support WinRT and not subject ourselves to x10 times the maintenance pain, we should perhaps abandon the idea of using socket read timeouts (to which we've switched recently in 9a688e92a05e25c43674d3a94532e607366b7a84).

michaelklishin:
Fixed for `stable`, needs to be ported to master.

michaelklishin:
Merged into master.

PR #99, opened 2015-06-18
Move heartbeat implementation back to timers and reset events

michaelklishin:
References #82, only for `stable` at the moment.


michaelklishin:
Multiple rounds of testing demonstrate that this implementation works well, and when heartbeats are missed, automatic connection recovery kicks in as expected.

michaelklishin:
Note: this cannot be automatically merged into master, will re-apply manually there soon.

Issue #94, opened 2015-06-04
Provide a logging abstraction


rjrizzuto:
It might be helpful to log exception/error cases from the .Net library to help in problem resolution.  I think Common.Logging is a good candidate for this.  It is a logging facade rather than a logging framework, so the user of RabbitMQ.Client can choose to use NLog, Log4net, etc. If nothing is configured specifically, Console.Logging logs NoOpLoggerFactoryAdapter.  

Here's a perfect case in point from AutorecoveringConnection.cs:

```
                catch (Exception e)
                {
                    // TODO: logging
                    Console.WriteLine("BeginAutomaticRecovery() failed: {0}", e);
                }
```

With common.logging:

```
                catch (Exception e)
                {
                    logger.Warn("BeginAutomaticRecovery() failed", e);
                }
```

The logger would log all the details, including stack trace (configurable in the logging config file) about the exception.

cjbhaines:
I think it would be a bad idea to add dependencies to this library. Every single consumer of it would then be forced to reference Common.Logging which is not a desirable situation. You only really want it to log on bugs in the internal code but the test coverage should be good enough to catch those before it's released. I think an event on the AutorecoveringConnection to notify of RecoveryFailed would be more applicable.

michaelklishin:
I'm not sure I agree with this. Exceptions do happen from time to time and it's not a great idea from the operations perspective to not have any logging.

We can make the client depend on Common.Logging, and if no logging is configured, it will be a no-op.

michaelklishin:
[This Twitter thread](https://twitter.com/RabbitMQ/status/606858510828236803) makes me wonder if emitting events is a safer bet. Java client-style exception handler is another option.

rjrizzuto:
I will add my 2 cents that other libraries have dependencies, and that NuGet is designed to handle installing dependencies when a library is added.    For example, we are using NEsper in our application, and it has several dependencies, including one on Common.Logging.

I read the twitter thread, and I see their points as well, and have hit some issues with dll dependencies on different versions of a library, so I can see the merit of callbacks vs common.logging, especially since you have no dependencies currently.

michaelklishin:
We'll investigate LogLib and Logary.

michaelklishin:
We've decided to provide more event handlers so that applications can use the logging library that suits them. We may still consider LogLib or Logary in the future.

asbjornu:
[...]

Issue #91, opened 2015-05-29
Async methods not available on .NET 4 or 4.5


tommy-carlier:
The asynchronous methods (e.g. *SharedQueue.DequeueAsync*) are not available in .NET 4 or 4.5, even though the source code indicates that these methods should be available (#if NETFX_CORE || NET4).

I suspect that the NET4 conditional symbol is not set correctly.

In some places, #if NETFX_CORE is checked without including NET4, even though it should.

It might be a good idea to introduce a conditional symbol ASYNC_SUPPORTED and use that instead of "NETFX_CORE || NET4" in places where asynchronous functionality is used, and then define this symbol in the .NET 4/4.5 and WinRT projects.

michaelklishin:
WinRT is still a work in progress. Feel free to look into things that should be done differently and submit a PR :)

michaelklishin:
@tommy-carlier if we'd like to get this into 3.6.0, now is the time to look into a PR. There's only a week or so weeks left till 3.6.0 enters feature freeze for such changes.

warrenrumak:
For what it's worth, Microsoft no longer supports .NET 4.0 and 4.5 with security updates or fixes.  It might be okay to let this item go.

https://blogs.msdn.microsoft.com/dotnet/2015/12/09/support-ending-for-the-net-framework-4-4-5-and-4-5-1/

michaelklishin:
Thank you.

We will develop a new client for .NET Core from scratch. This client then will transition to maintenance mode only for a year or two until the new one (as well as .NET Core) matures.

michaelklishin:
This client is in maintenance mode now and uses .NET Core tooling in master, which we couldn't get to build the WinRT version :( Closing this as it's not entirely clear if we can support WinRT in the future. .NET Core eclipses it and will be what master in this client and a new client (see #148) will target.

Issue #80, opened 2015-04-29
Make continuation timeouts configurable


michaelklishin:
See #79. This is a lot less painful in C# as far as changing the API goes: we are not tied by checked exceptions.

michaelklishin:
CI fails because Erlang installer cannot be downloaded :/

Issue #56, opened 2015-04-06
Use named arguments in C# tutorials


michaelklishin:
See https://github.com/rabbitmq/rabbitmq-dotnet-client/issues/20. It's already possible in C# as we require .NET 4.0 now, so we should update the tutorials.

mfloryan:
Shouldn't this be an issue on https://github.com/rabbitmq/rabbitmq-tutorials ?

michaelklishin:
This issue grew out of something filed in this repo. We can move it.

michaelklishin:
Thank you. I'll close this when we update rabbitmq-website to match.

Pliner:
As I see now, I forgot to update cs files from dotnet folder... :(

Pliner:
Done!

Pliner:
@michaelklishin If you want I could update website too, but actually site has a bit different version of .cs in comparison to dotnet-visual-studio folder. Should I replace site versions to dotnet-visual-studio folder?

michaelklishin:
I'd rather adapt the tutorials for now. There will be more code changes coming, so we'll have a way to compare the two and pick a version we like more.

michaelklishin:
This is done.

PR #101, opened 2015-06-20
Drop `immediate` flag

### LINKED ISSUE #98: Drop immediate flag
It isn't supported as of RabbitMQ 3.0 and was removed in many other clients.

Pliner:
Hi @michaelklishin,

It's an attempt to drop immediate flag and fix #98. I hope everything is fine :)


michaelklishin:
Thanks Yury, I will take a look tonight.

michaelklishin:
Thank you!

If you are looking for more straightforward issues you can help with, #13, #56, #80 are good candidates. Any help we can get on #32 and #91 would be greatly appreciated as well :)

Pliner:
Thank you too :)
 
The only reason I choose simple issues is absense of time to dive into codebase :( Do you have any progress with #94?

michaelklishin:
Haven't even started on #94 and won't for a while. I will update that issue.

PR #104, opened 2015-06-25
When sending methods with concent, flush socket stream once

michaelklishin:
References rabbitmq/rabbitmq-server#156.


michaelklishin:
Will port to `stable` manually.

PR #107, opened 2015-06-30
Stop Heartbeat timers before disposing them

david-poirier:
To prevent any further callbacks and allow background threads to exit

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/impl/Connection.cs
+++ b/projects/client/RabbitMQ.Client/src/client/impl/Connection.cs
@@ -1094,10 +1094,12 @@ entry.ToString());
         {
             if(_heartbeatReadTimer != null)
             {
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you.

Will cherry-pick to `stable`.

michaelklishin:
Running tests with this change causes a lot of `ObjectDisposedException`s. However, since we are tearing down we can ignore them. I'll update the branch and merge.

david-poirier:
You've got me a bit worried there! If the exceptions are in MaybeStopHeartbeatTimers then the Dispose method won't get a change to call Abort, which seems pretty important...

michaelklishin:
@david-poirier I think it's OK now that I added `ObjectDisposedException` handling.

david-poirier:
Perfect, thanks v much.

PR #111, opened 2015-07-09
fix turkish I problem during code generation

idursun:
If project is compiled with Turkish Regional Settings, generated identifiers have Turkish I problem. It causes compilation errors as follows:

```
error CS0115: 'RabbitMQ.Client.Framing.BasicProperties.Correlationİd': no suitable method found to override
```

Notice İ in `Correlationİd`

This commit fixes it.


michaelklishin:
Thank you!

PR #114, opened 2015-07-11
Cleaned up project folder structure and renamed projects\ to src\

cH40z-Lord:
The RabbitMQ's project layout has been reworked to adopt a more common folder structure that other popular projects are using, such as [ASP.NET 5](https://github.com/aspnet/Mvc/tree/dev/src) or [Entity Framework 7](https://github.com/aspnet/EntityFramework/tree/dev/src).

This results in a much more flattened folder structure and removed unnecessary clutter from Visual Studio.

Unfortunately it looks like squashing didn't properly work as I wanted (to get one single commit instead of all the mini-step commits I did). If that's a problem for you then I'll be happy to try to squash the commits again into a single one.

Let me know what you think.

Oh and if this PR makes it necessary for me to sign a CA then I'll be happy to do it. :)


michaelklishin:
Merging this means we won't be able to merge stable into master easily unless we do this on stable as well. And even that has to wait until after 3.5.4.

It will complicate digging in code history.

I don't know how much benefit there is in this change.

michaelklishin:
Before submitting massive changes like this, please ask for opinion on rabbitmq-users first.

michaelklishin:
After some thinking I conclude we cannot make this change. We need to merge #95 and #114 first, and both will likely take some time to get into good shape. Then this needs to be done against both `stable` and `master` branches to make merging and cherry-picking from `stable` possible.

Other than that, while I don't see a lot of value in this change, making solution structure more standard is fine with me.

michaelklishin:
@cH40z-Lord we've migrated our sub-projects to a different build system and now is a good time to change VS projects structure. Would you be interested in reviving this pull request against master? We are happy to use a more common structure and now is the right moment to do this. Thank you.

cH40z-Lord:
Sure, this might take a bit but I'll start working on it right away.

michaelklishin:
@cH40z-Lord you are the best :+1:

cH40z-Lord:
@michaelklishin do you want me to file a PR against master or a different branch? I'd rather not have it directly merged into master since I can't setup an entire testing infrastructure on my local machine right now and therefore would prefer that you guys check it out and test it properly before it is approved.

I've also got a few questions:
- Is the entire `RabbitMQ.Client.ApigenBootstrap` even required since it's only used in `RabbitMQ.Client.Apigen`? I'd propose merging them.
- Can we upgrade the solution (or rather the projects) to (download [Community Edition](https://www.visualstudio.com/en-us/downloads/download-visual-studio-vs.aspx)) so that we can make use of C# 6 features which would help reduce code clutter by introducing new compiler features. Keep in mind that the produced binaries still can be consumed/used from the same .NET versions as before since there are no new IL instructions that require a new .NET Runtime.
- Are there any objections in separating both WinRT projects with a [Shared Project](http://www.dotnetjalps.com/2015/02/shared-project-visual-studio-2015.html) since the WinRT project symlinks the entire regular project anyway? A Shared Project is an IDE feature that automatically copies all files of the Shared Project into any other project that is referencing it on build.

michaelklishin:
Against master. We obviously will test your branch before merging and adapt our build scripts as needed ;)

cH40z-Lord:
Alright, what about my other questions?

michaelklishin:
@cH40z-Lord bootstrap in other repos is a Python script. We can merge them eventually but lets do one thing at a time. We already require VS Community Edition 2013, that can be bumped to 2015 soon. We require .NET 4.5 for master now.

WinRT project actually doesn't copy 100% of the regular one, we do so only for the tests. I need to read up on shared projects. We can do it as a separate step after 3.6.0 is shipped, lets simply change the structure to be more standard for now.

We are interested in moving more internals to newer C# versions and have been doing so since 2013 (when we still supported .NET 1.1!). But that's a separate bunch of issues and perhaps should first be discussed on the list to narrow down the scope.

PR #115, opened 2015-07-14
Support for http proxy proxy

celper:
### Scenario: On-Premises/Cloud Hybrid Application 
An on-premise server provides endpoints to cloud based web services via some RCP-style messaging pattern. Unfortunately, the current rabbitmq .net client does not offer support for dealing with http proxies.

This pull request adds http proxy support by
- wrapping all used functionality of .net's standard .tcp client in an interface (*ITcpClient*), so that it can be more easily replaced by some custom implementation 
- providing implementation for a tcp client  (*TunneledTcpClient*) that implements this interface and can connect through http proxies via "HTTP CONNECT" (http tunnel).

### Usage: 
Overwrite socket factory on connection factory object
```
mqConnectionfactory.SocketFactory = family => new TunneledTcpClient(family, "proxy.my_company", 8080, "username", "password");
```

### Remarks: 
- This should only be used with SSL. 
- It could be required to tunnel through port 443 (mqConnectionfactory.Port=443;), since many cooperate firewalls block the standard rabbitmq ports. The incoming traffic on the rabbitmq server must then be redirected from port 443 to port 5671.
- The implementation of *TunneledTcpClient* is quite rough and could use some polishing. However, I have tested a prototype in different environments and it seems to work just fine.
- One could consider to create separate properties in ConnectionFactory to allow for easier access to proxy settings (e.g. ConnectionFactory.ProxyHost="proxy.my_company"). At the moment, the use of ConnectionFactoryBase.SocketFactory was less invasive and thus preferred.

### DIFF
--- a/projects/client/RabbitMQ.Client/RabbitMQ.Client.csproj
+++ b/projects/client/RabbitMQ.Client/RabbitMQ.Client.csproj
@@ -146,6 +146,7 @@
     <Compile Include="src\client\api\IQueueingBasicConsumer.cs" />
     <Compile Include="src\client\api\IRecoverable.cs" />
     <Compile Include="src\client\api\IStreamProperties.cs" />
+    <Compile Include="src\client\api\ITcpClient.cs" />
     <Compile Include="src\client\api\NetworkConnection.cs" />
     <Compile Include="src\client\api\PlainMechanism.cs" />
     <Compile Include="src\client\api\PlainMechanismFactory.cs" />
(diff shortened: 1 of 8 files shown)


michaelklishin:
Thank you. We will take a look later this week.

michaelklishin:
This PR does not touch any WinRT client files. Can you please look into what may need updating there?

celper:
Sorry, but I have absolutely no experience with WinRT client code...
and no Win8.1 at work :-)

However, according to https://msdn.microsoft.com/library/windows/apps/br226882, *StreamSocket* seams to have some kind of proxy support built-in:

> Support for proxies
> In a Windows Store app, the StreamSocket class supports connecting to a remote endpoint when proxies are required to complete the connection. This support for proxies is automatic and transparent to the app. A StreamSocket can establish a connection through authenticating proxies as well as through other proxies where authentication is not needed. StreamSocket

michaelklishin:
OK, thanks.

michaelklishin:
With this change WinRT client no longer compiles.

celper:
Great.

PR #120, opened 2015-08-18
Dispose method should not throw an exception

### LINKED ISSUE #119: Dispose method should not throw an exception
In [EasyNetQ](https://github.com/EasyNetQ/EasyNetQ) we're making some [important refactor](https://github.com/EasyNetQ/EasyNetQ/pull/461), and [we notice](https://github.com/EasyNetQ/EasyNetQ/pull/461#discussion-diff-37196714) that we had to wrap RabbitMQ.Client `ModelBase.Dispose` in a try-catch block because looking at [source code](https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/master/projects/client/RabbitMQ.Client/src/client/impl/ModelBase.cs#L318) it could throws an exception.

This is the interested piece of code of [`ModelBase.cs`](https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/master/projects/client/RabbitMQ.Client/src/client/impl/ModelBase.cs):

```csharp
        void IDisposable.Dispose()
        {
            Close();
        }

        public void Close()
        {
            Close(Constants.ReplySuccess, "Goodbye");
        }

        public void Close(ushort replyCode, string replyText)
        {
            Close(replyCode, replyText, false);
        }

        public void Close(ushort replyCode, string replyText, bool abort)
        {
            Close(new ShutdownEventArgs(ShutdownInitiator.Application,
                replyCode, replyText),
                abort);
        }

        public void Close(ShutdownEventArgs reason, bool abort)
        {
            var k = new ShutdownContinuation();
            ModelShutdown += k.OnConnectionShutdown;

            try
            {
                ConsumerDispatcher.Quiesce();
                if (SetCloseReason(reason))
                {
                    _Private_ChannelClose(reason.ReplyCode, reason.ReplyText, 0, 0);
                }
                k.Wait();
                ConsumerDispatcher.Shutdown(this);
            }
            catch (AlreadyClosedException ace)
            {
                if (!abort)
                {
                    throw ace;
                }
            }
            catch (IOException ioe)
            {
                if (!abort)
                {
                    throw ioe;
                }
            }
        }
```

As you can see the `abort` parameter is false when calling `Dispose` and this could end-up in an exception, and it is not the expected behavior, as [Microsoft states that](https://msdn.microsoft.com/en-us/library/bb386039.aspx):

>A IDisposable.Dispose method should not throw an exception. Dispose is often called as part of the clean up logic in a finally clause. Therefore, explicitly throwing an exception from Dispose forces the user to add exception handling inside the finally clause.
The Dispose(false) code path should never throw exceptions, because this is almost always called from a finalizer.

micdenny:
What about just calling `Abort` instead of `Close`?

I've also change the `throw ex` into just `throw`, I don't know if the wanted behavior was to reset the stacktrace, because `throw ex` it does exactly that, it resets the stacktrace and that is not so kind when debugging, in other words you don't know which method generated the exception.

I've also added the catch of all generic Exception, and probably here I'm pushing too far :smile: probably that kind of try-catch should be moved in the Dispose itself, just tracing any exception without throwing up.

fixes #119

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/impl/ModelBase.cs
+++ b/projects/client/RabbitMQ.Client/src/client/impl/ModelBase.cs
@@ -304,18 +304,25 @@ namespace RabbitMQ.Client.Impl
                 k.Wait();
                 ConsumerDispatcher.Shutdown(this);
             }
-            catch (AlreadyClosedException ace)
+            catch (AlreadyClosedException)
             {
                 if (!abort)
                 {
-                    throw ace;
+                    throw;
                 }
             }
-            catch (IOException ioe)
+            catch (IOException)
             {
                 if (!abort)
                 {
-                    throw ioe;
+                    throw;
+                }
+            }
+            catch (Exception)
+            {
+                if (!abort)
+                {
+                    throw;
                 }
             }
         }
@@ -677,7 +684,7 @@ namespace RabbitMQ.Client.Impl
 
         void IDisposable.Dispose()
         {
-            Close();
+            Abort();
         }
 
         public abstract void ConnectionTuneOk(ushort channelMax,


michaelklishin:
@micdenny thank you. If you'd like this to go into 3.5.5, please submit the same change against the `stable` branch.

michaelklishin:
I agree with the idea behind this change. As for catching `Exception` or resetting stack trace, if we don't want `Dispose` to throw anything, that sounds OK. When we add pluggable logging capabilities, it would be even easier to handle such cases.

micdenny:
> As for catching Exception or resetting stack trace, if we don't want Dispose to throw anything, that sounds OK

Do you prefer to surround with a try-catch block the `Dispose` method, or leave as I implemented in this PR, that means, changing also the behavior of `Abort` and suppress any kind of exception that could ever be thrown by the `Close(abort = true)` method?

> When we add pluggable logging capabilities, it would be even easier to handle such cases.

If we don't have the pluggable logging capabilities yet, in the meanwhile should I use the `Trace.TraceError(ex.ToString())`?

> If you'd like this to go into 3.5.5, please submit the same change against the stable branch.

I will do when you will be happy with this PR.

michaelklishin:
@micdenny `Abort` is meant to ignore all exceptions. That's its raison d'être :) So I'm fine with your change.

Yes, `Trace.TraceError(ex.ToString())` sounds reasonable for now.

I believe we've

micdenny:
Great, I send the same PR to the `stable` branch.

michaelklishin:
Thank you! Closing this one then.

micdenny:
here we go #121

PR #124, opened 2015-09-28
Publish NuGet Packages

joefitzgerald:
- [x] Get Basic AppVeyor Build Working
- [x] Run Tests
- [x] Embed Encrypted .snk
- [x] Produce NuGet Artifacts
- [ ] ~~Store Encrypted API Key~~
- [ ] ~~Publish NuGet Artifacts~~

This builds on the great work of @ngbrown in https://github.com/rabbitmq/rabbitmq-dotnet-client/issues/32.

**Note:** Publishing Of NuGet artifacts to nuget.org will be addressed in a separate issue.


michaelklishin:
@joefitzgerald thank you for looking into this. There are certain challenges to making this work, at least with Travis CI:

 * Mono 4 no longer supports loading DLLs built for .NET versions prior to 4.5. This is our case (some of the build dependencies), so our client cannot be built right now. Running on Mono 4 is fine.
 * We have a WinRT client, so development and test environments need to use Windows 8.1 (or exclude WinRT client by explicitly picking the "regular" client project)

joefitzgerald:
@michaelklishin Yup, that's why I'm using AppVeyor and building this on Windows for reals.

michaelklishin:
OK, it looks like the AppVeyor build may already be addressing both points above. One last thing: the tests need to use `rabbitmqctl` to imitate various scenarios that should trigger connection recovery. They assume that RabbitMQ runs from the `../rabbitmq-server` directory and can use `../rabbitmq-server/scripts/rabbitmqctl`. There *may* be something that assumes Cygwin as well, although I can't immediately think of anything.

michaelklishin:
@joefitzgerald thank you. I'd be happy to help with this next week.

joefitzgerald:
@michaelklishin Got any idea what might be causing the apigen issue?

michaelklishin:
@joefitzgerald thank you!

joefitzgerald:
For now, packages are being published to https://ci.appveyor.com/nuget/rabbitmq-dotnet-client-b5dx7iginj5n. We will refine the semver used in this feed, and then once this has been finalized, we will consider publishing packages generated from the master branch as prerelease packages on nuget.org.

PR #128, opened 2015-10-21
Fixing race condition with models in AutorecoveringConnection

resnikb:
`CreateModel` and `UnregisterModel` methods lock on `this`, instead of using `m_models`, which is used by `RecoverModels` method. This creates a race condition, where `m_models` collection can be modified while models are being recovered, resulting in _"Collection was modified; enumeration operation may not execute"_ exception in `RecoverModels`.


michaelklishin:
Good catch, thank you!

PR #129, opened 2015-11-03
Introduce continuation timeouts

### LINKED ISSUE #80: Make continuation timeouts configurable

michaelklishin:
Fixes #80.


michaelklishin:
CI fails because Erlang installer cannot be downloaded :/

PR #131, opened 2015-11-07
Added Local.props.example copy step to Contributing.md

cH40z-Lord:
Added an extra, ncessary, step to `contributing.md` about copying the `Local.props.example` into `Local.props` to prevent Visual Studio from erroring while trying to load the projects


michaelklishin:
Thank you!

PR #136, opened 2015-11-14
Applied new project folder structure

cH40z-Lord:
Re #114 I've updated the project's folder structure to a more common structure that is used across other popular projects.

If there's anything I forgot please let me know and I'll fix it asap.


michaelklishin:
@cH40z-Lord thank you. According to GitHub, `Local.props.example` and a few other release build files are gone. Is this intentional?

Note that GitHub doesn't display all changes because the diff is large, so I can be missing something.

cH40z-Lord:
@michaelklishin that file was moved into `build\skeletons\`. Looks like I forgot to adjust docs to reflect this change :)

michaelklishin:
@cH40z-Lord this PR has conflicts with master. Can you please resolve them?

michaelklishin:
I've intentionally pushed one change ported from `stable`, will try resolving this myself.

michaelklishin:
@cH40z-Lord perhaps I'm missing something but there is no `build` directory that would contain `skeletons`. Is it generated somehow?

I've pushed a [new branch, rabbitmq-dotnet-136](https://github.com/rabbitmq/rabbitmq-dotnet-client/tree/rabbitmq-dotnet-136), with a couple of small changes from master merged and various properties files restored in root until we finish this.

Can you please re-create your branch based on above, and put `build/skeletons` in place for `Local.props.example`? I'd like `.in` files to stay in the root, since they are used by our release scripts and/or AppVoyer.

cH40z-Lord:
Oh, it looks like the .gitignore is also included the intentional `root/build/` folder. I'll fix that when I'll get home.

cH40z-Lord:
Sorry once again for the inconvenience. I've re-added the /build/ folder. Can you please verify if everything's still working for you? Other than that is there any general feedback so far?

One thing that still bothers me apart from the discussions we had in #114 is the Unit test projects: They desperately need a folder structure that matches the structure of the lib under test.

michaelklishin:
@cH40z-Lord I trust your judgement on the directory structure. My only requirement is that we don't affect any users (e.g. by altering namespaces) and don't make our build scripts too convoluted (again, fairly unlikely as they do relatively little and rely on msbuild for most of heavy lifting).

I still see conflicts with `master` on this branch. Would you like to take a look into it or should I? It's not a major inconvenience for me.

michaelklishin:
After an attempt to switch to the new structure I'm afraid it would take more time to update the build system and iron out all the issue than I'm willing to invest. Thank you for your time @cH40z-Lord.

I guess we will be using our ancient non-standard project structure for some more time.

michaelklishin:
FTR, my branch is `rabbitmq-dotnet-136` and the solution compiles successfully from the command line but unit tests (`rabbitmqctl.bat` paths) and release-related bits need updating.

PR #164, opened 2016-04-12
Support client specified connection name

hairyhum:
Related to rabbitmq/rabbitmq-server#104 and #99


hairyhum:
I'm not sure about redundancy of the new API and I haven't run tests.

Issue #13, opened 2012-11-13
DequeueNowait returns null first time.


tomstaijen:
Is it possible that DequeueNowait will return nothing the first time because no data has been prefetched yet, even when there are messages on the queue? 

If I use Dequeue with a wait time of 10ms i do get a message the first time. 

If this is possible, is it by design? Or is there a workaround other then using a wait period?

I'm using 2.8.7 client.

michaelklishin:
@cvanbergen thanks for confirming!

kjnilsson:
Hi @tomstaijen, @rbanderton, @cvanbergen,

I think it is time to finally bring this issue to a close. As @hyperthunk observed a few years ago this is a naturally occurring race. The Queue property exposed on QueueingBasicConsumer does not provide direct access to the remote RabbitMQ queue. It is a local buffer to which messages are asynchronously delivered which is why there is a delay, however slight. If you want direct access to the remote queue the simplest approach is to use the 'BasicGet' method on the `Model` instance. Yes it is slower but it is more intuitive.

Ultimately I don't think QueueingBasicConsumer is a helpful abstraction and have issued a PR to deprecate it in favour of `EventingBasicConsumer` and the `Subscription` class in the `RabbitMQ.Client.MessagePatterns` namespace. Once that PR is merged I think we can close this issue.  

Cheers
Karl

michaelklishin:
#173 is in.

PR #173, opened 2016-05-12
Deprecate QueueingBasicConsumer and reduce its use by other components

kjnilsson:
- re-implement behaviour from SharedQueue inside Subscription to ensure NextAsync works
    - signal closure to pending tasks

Fixes: #13

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/api/QueueingBasicConsumer.cs
+++ b/projects/client/RabbitMQ.Client/src/client/api/QueueingBasicConsumer.cs
@@ -90,6 +90,7 @@ namespace RabbitMQ.Client
     /// }
     /// </code></example>
     /// </remarks>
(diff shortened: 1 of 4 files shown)


projects/client/RabbitMQ.Client/src/client/api/QueueingBasicConsumer.cs
```diff
@@ -90,6 +90,7 @@ namespace RabbitMQ.Client
     /// }
     /// </code></example>
     /// </remarks>
+    [Obsolete("Deprecated. Use EventingBasicConsumer or Subscription instead")]
```
michaelklishin: Subscription is largely the same thing. I suspect it also doesn't support automatic recovery, so we shouldn't mention it.

michaelklishin: I will make the above change myself, nevermind.

PR #179, opened 2016-05-17
Added missing restart of timer on WinRT

WinstonSmith77:
Timer for writeHeartBeat is not restarted on WinRT.

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/impl/Connection.cs
+++ b/projects/client/RabbitMQ.Client/src/client/impl/Connection.cs
@@ -1096,6 +1096,10 @@ entry.ToString());
                     TerminateMainloop();
                     FinishClose();
                 }
+                else
+                {
+                    _heartbeatWriteTimer.Change(Heartbeat * 1000, Timeout.Infinite);
+                }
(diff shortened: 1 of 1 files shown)


michaelklishin:
@kjnilsson if this can go into 3.6.x, please rebase against or cherry-pick to `stable`.

kjnilsson:
@WinstonSmith77 

hi - thanks for this! 

Could you give me some more information about why this change was required and what is specific to WinRT about it? As it stands the write timer will fire periodically anyway and shouldn't need resetting. 

Cheers
Karl

WinstonSmith77:
@kjnilsson Thank you for your response to our problem. 

Sorry. We did this fix in a hurry and did not search the actual reason for this bug. I just added the restart in HeartbeatWriteTimerCallback because there is on in HeartbeatReadTimerCallback and it worked afterwards.

But today a took some to analyze it properly. The root are are Lines 1001 and 1002 in Connection.cs. It reads  .Milliseconds but obviously it must be .TotalMilliseconds. This explains the difference between classic .net and WinRT as well.

Thank your for your efforts again

PR #191, opened 2016-05-31
Use TimeSpan.TotalMilliseconds on Windows RT

WinstonSmith77:
Reads Milliseconds; Must read TotalMilliseconds

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/impl/Connection.cs
+++ b/projects/client/RabbitMQ.Client/src/client/impl/Connection.cs
@@ -998,8 +998,8 @@ entry.ToString());
                 _heartbeatWriteTimer = new Timer(HeartbeatWriteTimerCallback);
                 _heartbeatReadTimer = new Timer(HeartbeatReadTimerCallback);
 #if NETFX_CORE
-                _heartbeatWriteTimer.Change(200, m_heartbeatTimeSpan.Milliseconds);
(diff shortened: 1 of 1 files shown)


michaelklishin:
That's a bit terse. Can we state clearly what the bug is? Is it that instead of a total time span only its milliseconds part was returned? (then it's pretty unfortunate naming compared e.g. to JodaTime).

kjnilsson:
TimeSpan.Milliseconds returns only the millisecond component. I bet this isn't the only time this naming has caused a bug. I will reword and merge this commit.

https://msdn.microsoft.com/en-us/library/system.timespan.milliseconds(v=vs.110).aspx

kjnilsson:
ok this has been reworded and cherry-picked into the stable branch, Thank you @WinstonSmith77

michaelklishin:
Thank you, @WinstonSmith77.

WinstonSmith77:
Thank you for your quick response. Will there be an updated NuGet anytime soon?

kjnilsson:
The .NET client releases are in step with broker releases. This fix will be
included in the next release (3.6.3) but I don't have a date for when that
will be.

michaelklishin:
@WinstonSmith77 we expect 3.6.3 to be available within a month. FWIW it will be in nightly builds in case you have reasons to not build from source.

PR #201, opened 2016-06-09
Fix stale delivery tags in RecoveryAwareModel

### LINKED ISSUE #200: Versioned delivery tags, à la Bunny and Java client
Hi,
I am testing a recovery scenario by dropping the connection through management console. Below is my scenario.

1. Put 10 messages in a queue.
2. Once messages are consumed by the application, wait for 30 seconds and then call BasicAck on the delivery tag received.
3. While waiting for 30 seconds, I dropped the connection from management console. I noticed that all the Unacked messages are now in Ready state once the connection is back up. For this to happen, I set `AutomaticRecoveryEnabled` and `TopologyRecoveryEnabled` to true.
4. After the Reconnection, my application those messages again with new delivery tags.
5. Now, for the first set (before reconnection) of 10 messages wait time is up and do BasicAck. Now, I get Channel Exception and that is not getting recovered. At this point, my application is not able to consume or Ack because the channel is down.

After reading the RabbitMq docs, I noticed that the driver is supposed to map the old delivery tags to new delivery tags and ignore any stale delivery tags.

```
Manual Acknowledgements and Automatic Recovery
When manual acknowledgements are used, it is possible that network connection to RabbitMQ node fails
between message delivery and acknowledgement. After connection recovery, RabbitMQ will reset
delivery tags on all channels. This means that basic.ack, basic.nack, and basic.reject with old delivery
tags will cause a channel exception. To avoid this, RabbitMQ Java client keeps track of and updates
delivery tags to make them monotonically growing between recoveries. Channel.basicAck,
Channel.basicNack, and Channel.basicReject then translate adjusted delivery tags into those used by
RabbitMQ. Acknowledgements with stale delivery tags will not be sent. Applications that use manual
acknowledgements and automatic recovery must be capable of handling redeliveries.
```

Is this true for .NET driver too? Or Am i missing any other configuration step?

Entroper:
Fixes #200.

`RecoveryAwareModel` was attempting to prevent acking of stale delivery tags by checking if the un-adjusted tag was > 0.  However, the tag is a `ulong`, not a `long`, so it will always be >= 0.  This change compares the un-adjusted tag to the adjusted tag to ensure it hasn't wrapped around to a higher number.

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/impl/RecoveryAwareModel.cs
+++ b/projects/client/RabbitMQ.Client/src/client/impl/RecoveryAwareModel.cs
@@ -86,7 +86,7 @@ namespace RabbitMQ.Client.Impl
             bool multiple)
         {
             ulong realTag = deliveryTag - ActiveDeliveryTagOffset;
-            if (realTag > 0)
+            if (realTag > 0 && realTag <= deliveryTag)
             {
                 base.BasicAck(realTag, multiple);
             }
@@ -97,7 +97,7 @@ namespace RabbitMQ.Client.Impl
             bool requeue)
         {
             ulong realTag = deliveryTag - ActiveDeliveryTagOffset;
-            if (realTag > 0)
+            if (realTag > 0 && realTag <= deliveryTag)
             {
                 base.BasicNack(realTag, multiple, requeue);
             }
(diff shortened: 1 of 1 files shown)


kjnilsson:
works, good catch, thank you @Entroper for the contribution!

PR #219, opened 2016-07-11
trim .NET Core dependencies

lahma:
I've trimmed the current package dependencies from full 1.6 library to minimal required libs as it's considered a best practice: https://docs.microsoft.com/en-us/dotnet/articles/core/deploying/reducing-dependencies . System.Runtime.InteropServices could be thrown away if COM visibility would be sacrificed.

### DIFF
--- a/projects/client/RabbitMQ.Client/project.json
+++ b/projects/client/RabbitMQ.Client/project.json
@@ -12,8 +12,7 @@
       ]
     }
   },
-  "dependencies": {
-  },
+  "dependencies": {},
   "frameworks": {
     "net451": {},
     /*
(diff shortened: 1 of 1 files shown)


kjnilsson:
Great - thank you @lahma

waynebrantley:
Do we need  COM visibility ?

michaelklishin:
I don't think it's a particularly important thing to have but we are always surprised by what our users attempt to do ;) Is `System.Runtime.InteropServices` really a big deal to have on the dependency list?

waynebrantley:
Probably not, but COM visibility has no meaning on other platforms and the .net core is all about this being able to run on other platforms.  I would go with remove it and see what comes of it...can always put it back or look at other options if several people complain.

michaelklishin:
Given that this is in master, I'm not against it. Will leave it up to @kjnilsson to decide how beneficial it would be to drop this dependency :)

kjnilsson:
I think we should remove it. I agree with @waynebrantley 's point about portability.

lahma:
Yet another one to go probably could be System.Console. It might be beneficial to replace console logging with LibLog that allows you to embed it as an internal class. It provides quite powerful logging capabilities by delegating to NLog, Serilog and log4net.

michaelklishin:
Please leave logging changes to #94. I am not sure how much time we want to spend on this client vs. developing a new one but adding sensible logging makes sense. If only there was a consensus on what sensible logging looks like ;) But it's a matter for #94.

kjnilsson:
Remove InteropServices in commit: https://github.com/rabbitmq/rabbitmq-dotnet-client/commit/5e0750647cd7358cd46fdfa8ce1a19fa4c70c12d

Issue #244, opened 2016-08-25
TcpClientAdapter tries to open IPv4 connections with an IPv6 address


tomstaijen:
With the current TcpClientAdapter doing a First() on the available ip-adresses I found it is possible that it will try to make a ipv4 connection with an ipv6 IPAddress, which results in a "protocol not supported" exception.

This may be due to my setup, running rabbitmq on a vagrant. However, it seems better to filter the ip addresses on the appropriate addressfamily.

Suggested:

`var ep = adds.First(a => a.AddressFamily == sock.AddressFamily);`

I'd have submitted a pull request if I had time right now to figure out the whole umbrella thing.

kjnilsson:
Thanks it sounds like a good suggestion I will give it a try. 

Btw, you shouldn't need the umbrella to do development on the dotnet client. All you need is a running instance of RabbitMQ to run the tests against.

michaelklishin:
@tomstaijen FWIW we have a public CI service. Do you have a test case in mind for this that won't require a specialized environment?

kjnilsson:
Getting an integration test to reliably produce this scenario may be tricky however we could refactor the code to abstract the dns lookup into a unit testable component that could be tested for the scenario where an Ipv6 address is returned first in the collection but a IPv4 address is required. It does pass the current suite with this change in place.

kjnilsson:
I created a test that passes in a socket with a Unix "AddressFamily" type which at checks the case where no address is found with the same AddressFamily as the socket. Also added a missing tests for CreateConnection.

kjnilsson:
@tomstaijen I've just published a release candidate of 4.0.2 that contains these changes. 

https://www.nuget.org/packages/RabbitMQ.Client/4.0.2-rc1

Cheers

PR #247, opened 2016-08-25
Handle the case where a dns lookup returns an address for an

kjnilsson:
AddressFamily different to how the socket is configured

Fixes: #244


kjnilsson:
Resubmitting this formerly reverted change

Issue #251, opened 2016-08-26
ConsumerWorkService / BatchingWorkPool implementation is a performance bottleneck 


bording:
Recently, I've been doing some performance profiling of the [NServiceBus RabbitMQ transport](https://github.com/Particular/NServiceBus.RabbitMQ) and I've come across some interesting findings. 

When I started my performance investigation, the main machine I was using was an older Sandy Bridge i7-2600K. In the middle of this, I put together a new development machine that has a brand new Skylake i7-6700K. On my new machine, I noticed the throughput numbers I was seeing in my tests were much lower than on my older machine!

I put together a repro project [here](https://github.com/bording/RabbitPerfTest) and was able to run the scenario on a number of different machines:

```
Core 2 Duo E6550
--------------
1 consumer: 6908
32 consumer: 7667

i7-2600K
--------------
1 consumer: 7619
32 consumer: 11124

i7-4870HQ
--------------
1 consumer: 6103
32 consumer: 5237

i5-5200U
--------------
1 consumer: 7178
32 consumer: 1928

i7-6500U
--------------
1 consumer: 2243
32 consumer: 1253

i7-6700K
--------------
1 consumer: 6476
32 consumer: 2213
```

Those numbers are messages/sec. While the actual numbers aren't that important, the general trend here is problematic. On every CPU I tested it on that was newer than the 2nd-gen Core/Sandy Bridge, the more consumers there are, the worse the performance gets.

After spending some time analyzing some profiler results, my colleague @scooletz and I came to the conclusion that the current design of the [ConsumerWorkService](https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/master/projects/client/RabbitMQ.Client/src/client/impl/ConsumerWorkService.cs) / [BatchingWorkPool](https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/master/projects/client/RabbitMQ.Client/src/util/BatchingWorkPool.cs) is the culprit here. It appears that there is a lot of lock contention going on, and that seems to be causing a lot of trouble for the new cpus for some reason.

We have been able to come up with some alternative designs that eliminate this performance problem, though there are some trade-offs with them, so we wanted to show you both of them. 

[...]

PR #252, opened 2016-08-26
Improve batching work pool performance

bording:
This is the first PR mentioned in #251.

Currently this is based on my vs-support branch from #248, once that's merged, I'll rebase this against master instead.

https://github.com/rabbitmq/rabbitmq-dotnet-client/commit/f36ca331c217cfbc06aaf113ccdc242755bd4932 is what is new in this PR.

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/impl/ConsumerWorkService.cs
+++ b/projects/client/RabbitMQ.Client/src/client/impl/ConsumerWorkService.cs
@@ -62,25 +62,11 @@ namespace RabbitMQ.Client
 
         public void RegisterKey(IModel model)
         {
-            //special connection model 0 should never have consumers,
-            //so no need to register it
-            if (model.ChannelNumber == 0)
-            {
-                return;
-            }
-
             this.workPool.RegisterKey(model);
         }
 
         public void StopWork(IModel model)
         {
-            //special connection model 0 should never have consumers,
-            //so it won't be registered
-            if (model.ChannelNumber == 0)
-            {
-                return;
-            }
-
             this.workPool.UnregisterKey(model);
         }
 
--- a/projects/client/RabbitMQ.Client/src/client/impl/ConsumerWorkService.cs
+++ b/projects/client/RabbitMQ.Client/src/client/impl/ConsumerWorkService.cs
@@ -1,7 +1,6 @@
 ﻿using RabbitMQ.Util;
 using System;
 using System.Collections.Generic;
-using System.Threading;
 using System.Threading.Tasks;
 
 namespace RabbitMQ.Client
@@ -48,12 +47,7 @@ namespace RabbitMQ.Client
             }
             catch (Exception)
             {
-#if NETFX_CORE
-                // To end a task, return
                 return;
-#else   
-                //Thread.CurrentThread.Interrupt(); //TODO: what to do?
-#endif
             }
         }
 
@@ -68,11 +62,25 @@ namespace RabbitMQ.Client
 
         public void RegisterKey(IModel model)
         {
+            //special connection model 0 should never have consumers,
+            //so no need to register it
+            if (model.ChannelNumber == 0)
+            {
+                return;
+            }
+
             this.workPool.RegisterKey(model);
         }
 
         public void StopWork(IModel model)
         {
+            //special connection model 0 should never have consumers,
+            //so it won't be registered
+            if (model.ChannelNumber == 0)
+            {
+                return;
+            }
+
             this.workPool.UnregisterKey(model);
         }
 
--- a/projects/client/RabbitMQ.Client/src/util/BatchingWorkPool.cs
+++ b/projects/client/RabbitMQ.Client/src/util/BatchingWorkPool.cs
@@ -38,204 +38,117 @@
 //  Copyright (c) 2007-2016 Pivotal Software, Inc.  All rights reserved.
 //---------------------------------------------------------------------------
 
-using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.Threading;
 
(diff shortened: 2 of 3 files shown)


projects/client/RabbitMQ.Client/src/client/impl/ConsumerWorkService.cs
```diff
@@ -68,11 +62,21 @@ public void AddWork(IModel model, Action fn)
 
         public void RegisterKey(IModel model)
         {
+            if (model.ChannelNumber == 0)
```
michaelklishin: For those confused by this magic constant: channel 0 is used for "special purpose" protocol communication.

bording: Yes, I noticed that the current design causes even the `MainSession` model to be registered with the `ConsumerWorkService`, even though no consumers can be registered on it. Seemed like a reasonable optimization to skip those registrations.

michaelklishin: Makes sense. Please leave a comment explaining that, through.

bording:
@michaelklishin I pushed up a revised commit with a comment about model 0, let me know if it's ok.

bording:
@kjnilsson What race condition was https://github.com/rabbitmq/rabbitmq-dotnet-client/commit/166cf17f0c981cdf244392a7efd687b659a24b41 trying to fix? I was testing this PR initially against 3.6.5, which didn't have this, and it looks like the ChannelNumber 0 optimization I made here is now throwing an NRE because of it.

Same thing problem should be occuring on #253.

bording:
I've rebased this branch on top of the lastest vs-support branch changes, and I went ahead and removed the special model 0 handling that was throwing an exception because of the ordering change. Having that in this branch is much less critical.

kjnilsson:
The concourse build failed on this test:

```
Errors and Failures
1) Failed : RabbitMQ.Client.Unit.TestConsumerOperationDispatch.TestDeliveryOrderingWithSingleChannel
  Expected: 96
  But was:  100
at RabbitMQ.Client.Unit.TestConsumerOperationDispatch.TestDeliveryOrderingWithSingleChannel() in /tmp/build/b95f7b06/rabbitmq_dotnet_client/projects/client/Unit/src/unit/TestConsumerOperationDispatch.cs:line 142
```

michaelklishin:
@kjnilsson that's likely a timing issue.

kjnilsson:
@michaelklishin yes just wanted to note it as it isn't yet public. Running the tests again.

bording:
This branch is now rebased against master.

For the failing test, yes that is expected on this PR, because the changes to the BatchingWorkPool can't guarantee ordering. Looking at the test and thinking about it more, I can see how this likely isn't an acceptable trade-off, given how acking of multiple delivery tags works.

kjnilsson:
@bording can you explain where the ordering guarantees are given up with your approach?

michaelklishin:
If we really give up the ordering guarantee in this PR, it's a no go compared to #253.

bording:
@kjnilsson Take a look at `BatchingWorkPool.AddWorkItem` and how it uses the `ready` queue. It will add an entry in there for every call, so you can end up with multiple `NextWorkBlock` calls looking at the same model's work.

The existing implementation uses the `SetQueue` class to prevent the same model from being queued multiple times, but the class isn't thread-safe, and I couldn't find any sort of concurrent version of a collection that had the same behavior.

@scooletz and I tried to come up with another way to prevent the same model from being queued multiple times, but everything we came up with either required reintroducing locks (and defeating the point of this change) or could potentially run into the opposite problem of having work items and not scheduling a `NextWorkBlock` call to dispatch them.

michaelklishin:
@bording then it sounds like you and @scooletz consider #253 objectively superior to this PR?

bording:
@michaelklishin If ordering is a hard requirement that can't be relaxed, then I think #253 is the only option, unless there's some way we haven't thought of to bring it back to this PR without reintroducing a lock.

 I only realized today that it might not be possible to give up ordering because of the multiple delivery tag ack option. We don't use that in our code at all, and I had forgotten that it was even part of the spec. :cry: 

If ordering wasn't a requirement, than I actually prefer this approach, because it does maintain the ability to pass in a custom task scheduler for concurrency throttling, and since it's not using dedicated threads that block when they don't have work, you see a lot less "synchronization time" when profiling compared to #253.

Given that this approach does have some benefits, that was why we initially presented both options.

michaelklishin:
We cannot give up ordering for backwards compatibility, too. I think the numbers you posted from #253 are very comparable. Not sure how many users provide a custom task scheduler.

kjnilsson:
Ok I've had a play with the PR to see if I can get the `TestDeliveryOrderingWithSingleChannel` to pass somehow but even after adding back the SetQueue (or SeqQueue behaviour rather) _and_ adding a bunch of relatively course grained locks the test still fails. I suspect something else may be going on that isn't immediately obvious. I now agree we should proceed with https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/253 . It does add a breaking change to a public interface (hello v5!) and a fair bit of internal changes as well but the approach is simple and easy to reason about. I don't think that many use a custom TaskScheduler but with a one to one relationship between `Model`s and threads it may not be necessary. 
Also it looks like it may allocate less than a creating tasks recursively which would be nice. Once 253 is rebased I will review it properly and test it.

michaelklishin:
Speaking of versions, we should probably release `4.1.0` with it (the API change is not going to affect too many).

michaelklishin:
Closing this in favour of #253 based on a number of discussions in #251, #253 and here.

PR #253, opened 2016-08-29
No lock contention at consumer work service

Scooletz:
This is the second PR mentioned in #251, friendly competing with #252.

Currently this is based on @bording `vs-support` branch from #248, once that's merged, it can be rebased.

e919982 is what is new in this PR.


projects/client/Unit/RabbitMQ.Client.Unit.csproj
```diff
@@ -0,0 +1,178 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
bording: @Scooletz This csproj shouldn't be part of the PR, since the vs-support branch is using project.json/xproj instead.

bording:
@kjnilsson This PR will be having the same NRE problem as #252, caused by the change in https://github.com/rabbitmq/rabbitmq-dotnet-client/commit/166cf17f0c981cdf244392a7efd687b659a24b41

Being able to prevent model 0 from registering is even more important in this approach, because a dedicate thread will be started up otherwise!

bording:
One additional change that this approach would benefit from would be to change when models are registered with the ConsumerWorkService. Right now, they are registered as soon as the model is created, which means we'd be creating thread loops for models that never have BasicConsume called on them.

projects/client/Unit/RabbitMQ.Client.Unit.csproj
Scooletz: Will undo.

Scooletz:
@bording 7a4e9c7 addresses your https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/253#issuecomment-243290333 making the registration of workpools corresponding to 'IModel's lazy.

michaelklishin:
Bunny uses the same approach (lazy initialization of consumer work pool threads), it generally works well.

Scooletz:
`Bunny` ? :) @michaelklishin

michaelklishin:
@Scooletz the most popular Ruby client.

michaelklishin:
Note: this no longer merges cleanly into master.

michaelklishin:
According to a discussion in #252, that PR doesn't offer the same ordering guarantee so we should proceed with this one. Please rebase it against master.

Scooletz:
Ping @kjnilsson I've rebased changes on top of the master adding the condition similar to the connection for creating either a `Thread` or a long running task (thx @bording :+1: ).

Unfortunately it looks that the build is :red_circle: and it seems that I can't access `concourse` CI to see the logs. 

Anyway, looking forward to hearing more from you @kjnilsson  :)

michaelklishin:
It's a timing issue again (Concourse contains are resource constrained):

```
Errors and Failures
1) Failed : RabbitMQ.Client.Unit.TestConsumerOperationDispatch.TestDeliveryOrderingWithSingleChannel
  Expected: property Count equal to 100
  But was:  99
at RabbitMQ.Client.Unit.TestConsumerOperationDispatch.TestDeliveryOrderingWithSingleChannel() in /tmp/build/b95f7b06/rabbitmq_dotnet_client/projects/client/Unit/src/unit/TestConsumerOperationDispatch.cs:line 139
Run Settings
    WorkDirectory: /tmp/build/b95f7b06/rabbitmq_dotnet_client/projects/client/Unit
Test Run Summary
  Overall result: Failed
  Test Count: 221, Passed: 220, Failed: 1, Inconclusive: 0, Skipped: 0
    Failed Tests - Failures: 1, Errors: 0, Invalid: 0
  Start time: 2016-09-01 08:54:36Z
    End time: 2016-09-01 09:07:07Z
    Duration: 751.196 seconds
Results saved as /tmp/build/b95f7b06/rabbitmq_dotnet_client/projects/client/Unit/TestResult.xml
```

michaelklishin:
I triggered a new Concourse build.

projects/client/RabbitMQ.Client/src/client/impl/ConsumerWorkService.cs
```diff
@@ -1,84 +1,112 @@
...
+            }
+
+            public void Start()
+            {
+#if NETFX_CORE
+                Task.Factory.StartNew(Loop, TaskCreationOptions.LongRunning);
+#else
+                var thread = new Thread(Loop)
+                {
+                    Name = "WorkPool-" + name,
```
kjnilsson: Is it important to set the threadpool name here or could we just dump the `#if` directive and just use a long running task?

projects/client/RabbitMQ.Client/src/client/api/IConnectionFactory.cs
```diff
@@ -129,13 +129,6 @@ public interface IConnectionFactory
         IConnection CreateConnection(IList<string> hostnames, String clientProvidedName);
 
         /// <summary>
-        /// Advanced option.
-        ///
-        /// What task scheduler should consumer dispatcher use.
-        /// </summary>
-        TaskScheduler TaskScheduler { get; set; }
```
kjnilsson: If we leave this property in and mark it as `Obsolete` with an appropriate warning explaining it is no longer used we could release this as 4.1.0 and still remain SemVer-ish compatible.

kjnilsson:
The tests passed on my machine and it looks like concourse has now finally passed as well although AppVeyor has failed. It is interesting that some tests that I have never/rarely seen fail before now fail more regularly. I will need to have a closer look at these tests to see if they can be made more reliable.

projects/client/RabbitMQ.Client/src/client/impl/ConsumerWorkService.cs
Scooletz: I'd say that for profiling purposes it can quite helpful. For instancem if you use `Timeline` in dotTrace for instance you can exclude threads by their names narrowing the profiled area.

projects/client/RabbitMQ.Client/src/client/api/IConnectionFactory.cs
Scooletz: This won't be an API change, but will break the behavior. If, for some reason, somebody used a `TaskScheduler` that limits the number of executed tasks it will no longer behave in the same manner.

kjnilsson: Yes but it will still compile and function against current code, albeit with a warning. This will give users an opportunity to review and update their code for when the property is properly removed later.

Scooletz: I'm fine with restoring it if that will make this `4.1.0` according to the definition of semantic version you provided. In my opinion, we're in this grayish area of the semantic versioning. 
I'm pushing the commit right away.

Scooletz:
Ping @kjnilsson The `TaskScheduler` properties are back with `Obsolete` as in https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/253#discussion_r77160326

projects/client/RabbitMQ.Client/src/client/api/IConnectionFactory.cs
kjnilsson: Yep I agree we're in the grey area. SemVer often is, especially when deprecating features. We'll push out an RC before so people can make their voices heard if they feel we've got something wrong or if there is a surprisingly large number of users that write their own TaskSchedulers.

projects/client/RabbitMQ.Client/src/client/impl/ConsumerWorkService.cs
kjnilsson: Ok let's leave it as is.

bording:
@kjnilsson I want to follow up on the SemVer conversation. One of the things that makes SemVer harder to follow for this codebase is that everything is public, including all of the "internal" implementation classes.

For this PR, even though the `TaskScheduler` property has been added back, there are still breaking changes here in public classes:
- `BatchingWorkPool` has been removed
- `ConsumerWorkService` no longer has a constructor that takes a `TaskScheduler`
- `ConsumerWorkService.MAX_THUNK_EXECUTION_BATCH_SIZE` has been removed
- `ConsumerWorkService` no longer has the `ExecuteThunk` or `RegisterKey` methods

Given that these classes are implementation details, it's not likely that anyone is using them directly, but because they are public, you can't rule that out, and these are both binary and compile-time breaks.

I'm not arguing that all this stuff should be put back in when they aren't needed anymore. Instead I think it might be a good idea to go through and define what the actual public api surface should be, internalize the rest, and then take a single break up to 5.0 and be in much better shape for avoiding breaks in the future.

You have to be able to make these kind of changes to improve things, but having everything public means you'll have to be increasing major versions quite often.

Scooletz:
@bording is totally right. Having big public API will make you break it a lot. Another valid point, my changes are big and are breaking. This isn't only a scheduler but the whole change as everything is `public`.

michaelklishin:
While there certainly are things that can be made private or protected, not every public member or class or function is necessarily a part of the public API. We will see about how soon we want to reduce the scope of what's public. This change deserves a `4.1.0` and not a `5.0.0`. Let's focus on finishing it first.

Also note that we have a completely new .NET client for the cutting edge C# and .NET features planned for next year. While this doesn't mean that this client won't be maintained, there's only so much we can change in a 9 year old library, and our team is small and has to allocate our time wisely.

kjnilsson:
Many thanks to @Scooletz  and @bording :)

Scooletz:
> Also note that we have a completely new .NET client for the cutting edge C# and .NET features planned for next year. While this doesn't mean that this client won't be maintained, there's only so much we can change in a 9 year old library, and our team is small and has to allocate our time wisely.

Is there a repo for this @michaelklishin ?

michaelklishin:
@Scooletz there is no repo yet, we will likely begin around December or January 2017.

PR #261, opened 2016-10-03
Added a way to intentionally propagate exceptions through CallbackException

ndry:
Added a way to intentionally propagate exceptions through CallbackException. It gets lost otherwise; and there is no option to disable custom RabbitMQ.Client's exception handling.

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/impl/ModelBase.cs
+++ b/projects/client/RabbitMQ.Client/src/client/impl/ModelBase.cs
@@ -579,8 +579,11 @@ namespace RabbitMQ.Client.Impl
                     {
                         h(this, args);
                     }
-                    catch
+                    catch (Exception e)
                     {
+                        if (Object.ReferenceEquals(e, args.Exception)) {
+                            throw; // Exception was intentionally rethrown in callback -- propagate it further
+                        }
                         // Exception in
                         // Callback-exception-handler. That was the
                         // app's last chance. Swallow the exception.


michaelklishin:
If you look at a comment below the lines changed, ignoring exceptions in exception handlers was intentional.

ndry:
I saw that comment and if you look in my comment, you'll see that propagation is intentional too. Maybe I just do not understand the reason to break default .net exception flow, but leaving an option to keep it working would be nice.

michaelklishin:
Again, I believe

```
// Exception in
// Callback-exception-handler. That was the		                          
// app's last chance. Swallow the exception.
```

kind of explains what the idea is. We don't want to introduce exception handlers for exception handlers. I'm not convinced that this is an improvement. If someone isn't sure about the reliability of their own exception handlers perhaps catching and handling exceptions there and then isn't too much to ask for.

@kjnilsson WDYT? This is technically a breaking change and bumping the version to `4.2.0` because of this also seems a bit much to me.

kjnilsson:
I am not sure what is achieved by deliberately re-throwing the exception in the handler. @ndry could you explain how you would make use of this in your application code?

ndry:
I want my application to crash on an exception and do not look like things are still going great when something went wrong. And I cannot do that, because RabbitMQ.Client silently consumes all my exceptions. 

Using a hack I proposed I can rethrow it in exception callback to tell your invoker that I am intentionally propagating further throwing. Then I am (for example) going to catch that exception by AppDomain.UnhandledException handler, log it -- and crash noisily.

 I found no other way to propagate an exception by now. Can you suggest any? Maybe, I am missing something.

Moreover, I think, generally it would be better to explicitly throw an exception if there are no exception callback handlers subscribed (so that the default behavior is the one that does not affect the expected exception flow). But the last one it is really breaking change.

michaelklishin:
@ndry you can also do all of that in your exception handler. I understand the intent. I am not convinced that this change is a good idea, however.

This is the first time this was reported in ~ 4 years. Presumably not a lot of people run into exceptions thrown by exception handlers, or they handle them explicitly in their own code.

kjnilsson:
If you really want your application to exit on an unhandled exception you could just explicitly kill it in the exception handler using `Process.Kill` or similar.  

The OnCallbackException handler typically (always?) execute on a background thread which just will kill the app something that the vast majority of uses don't wan't to happen.

michaelklishin:
@kjnilsson good point. Consumer callbacks are executed in threads that are not explicitly started by the user, so using exceptions for flow control there is of questionable value and/or can be confusing.

Again, user-provided exception handlers can handle the exceptions they expect, and it is the most predictable (and backwards compatible) way to solve the problem.

ndry:
Okay. I understand you. Still imho, it is a strange practice to alter the irrelevant defaults in something that is intended to be integrated in other systems. Anyway, thank you for doing the great job on this project!

PR #275, opened 2016-10-31
Fix potential EventingBasicConsumer race condition

### LINKED ISSUE #242: EventingBasicConsumer race condition possibility
I was just looking through the code of the EventingBasicConsumer and noticed that when the even is Received is raised, no temp copy is used to prevent race condition.  According to [this ](https://msdn.microsoft.com/en-us/library/w369ty8x.aspx) article you should have a temporary copy of the delegate in order to avoid the possibility of a race condition if the last subscriber unsubscribes immediately after the null check and before the event is raised.

Example from the link above:

```
protected virtual void OnRaiseCustomEvent(CustomEventArgs e)
{
      // Make a temporary copy of the event to avoid possibility of
      // a race condition if the last subscriber unsubscribes
     // immediately after the null check and before the event is raised.
     EventHandler<CustomEventArgs> handler = RaiseCustomEvent;

    // Event will be null if there are no subscribers
    if (handler != null)
    {
        // Format the string to send inside the CustomEventArgs parameter
        e.Message += String.Format(" at {0}", DateTime.Now.ToString());

        // Use the () operator to raise the event.
        handler(this, e);
    }
}
```

Also [this](https://codeblog.jonskeet.uk/2015/01/30/clean-event-handlers-invocation-with-c-6/) article could be useful to see cleaner ways of dealing with the issue.

kjnilsson:
Fixes #242

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/events/EventingBasicConsumer.cs
+++ b/projects/client/RabbitMQ.Client/src/client/events/EventingBasicConsumer.cs
@@ -68,22 +68,14 @@ namespace RabbitMQ.Client.Events
         public override void HandleBasicCancelOk(string consumerTag)
         {
             base.HandleBasicCancelOk(consumerTag);
-
-            if (Unregistered != null)
-            {
-                Unregistered(this, new ConsumerEventArgs(consumerTag));
-            }
+            Raise(Unregistered, new ConsumerEventArgs(consumerTag));
         }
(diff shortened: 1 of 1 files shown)


michaelklishin:
@kjnilsson is this meant to go into `3.6.6`?

kjnilsson:
Yes! :)

PR #295, opened 2016-12-07
Ensures connection recovery does not keep going after the connection …

### LINKED ISSUE #294: Closing an AutoRecoveringConnection does not stop it from continuing recovery forever
Once an AutoRecoveringConnection connection enters auto recovery and trying to recover the connection will continue to do try forever even if the connection is manually closed. 

I am not super familiar with the code, but believe the issue lays in the RecoverConnectionDelegate method that enters a while loop and fails to check if the connection has been manually closed via the manuallyClosed variable.

The order for a repro is:

1. Create and Open auto recovering connection
2. Sever the network connection between the client and rabbit server. The connection will start attempting auto-recovery.
3. Call .Close() and/or .Dispose() on the connection
4. Observe that the connection continues to try and auto-recover ... forever. 

I would have expected that the connection would have given up trying to reconnect after it was closed, but i might be missing some subtle issue here that i am not aware of.

kjnilsson:
…has been manually closed.

Fixes #294 

So this involved more changes than I initially thought. `Close` and `Abort` now first checks if the connection `IsOpen` before attempting to close it. This is needed as otherwise `Close` would just throw if it was called during recovery which seems unnecessary to me unless this is semantics we want to keep?

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/api/IConnection.cs
+++ b/projects/client/RabbitMQ.Client/src/client/api/IConnection.cs
@@ -194,7 +194,7 @@ namespace RabbitMQ.Client
         /// <remarks>
         /// Note that all active channels, sessions, and models will be closed if this method is called.
         /// In comparison to normal <see cref="Close()"/> method, <see cref="Abort()"/> will not throw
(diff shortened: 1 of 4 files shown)


michaelklishin:
Started a discussion in #294 to not limit it to just this approach.

PR #297, opened 2016-12-09
noAck -> autoAck

### LINKED ISSUE #255: Clarify the noAck parameter of BasicConsume
When following the tutorial, I came across this line:

> Message acknowledgments are turned on by default. In previous examples we explicitly turned them off by setting the noAck ("no manual acks") parameter to true. It's time to remove this flag and send a proper acknowledgment from the worker, once we're done with a task.

This really bugs me. Indeed, the `noAck` option was set to `true` on the first example, but this option enables the acknowledgment instead of disabling it.

When looking for the documentation, I found [the java equivalent to this function](https://www.rabbitmq.com/releases/rabbitmq-java-client/v3.1.1/rabbitmq-java-client-javadoc-3.1.1/com/rabbitmq/client/Channel.html#basicConsume%28java.lang.String,%20boolean,%20java.lang.String,%20boolean,%20boolean,%20java.util.Map,%20com.rabbitmq.client.Consumer%29), where the parameter is named `autoAck`
- The parameter should be named `autoAck`, to keep aligned with the Java client
- The default value should be set to `true`
- The documentation should be updated.

kjnilsson:
Fixes #255

This renames the `noAck` parameter to `autoAck`.

Before we release this change we need to update `rabbitmq-website` and `rabbitmq-tutorials` with the new name.

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/api/IModel.cs
+++ b/projects/client/RabbitMQ.Client/src/client/api/IModel.cs
@@ -196,7 +196,7 @@ namespace RabbitMQ.Client
         [AmqpMethodDoNotImplement(null)]
         string BasicConsume(
             string queue,
-            bool noAck,
+            bool autoAck,
             string consumerTag,
             bool noLocal,
             bool exclusive,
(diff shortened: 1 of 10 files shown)


kjnilsson:
@michaelklishin I'll fix that. Must have forgotten to re-run autogen.

PR #302, opened 2016-12-23
Limited the capacity of ConsumerWorkService's WorkPools

or150:
A proposed fix for #301 

I've replaced the ConcurrentQueue inside WorkPool with a BlockingCollection in order to be able to limit the amount of work that a WorkPool can contain, and made it configurable.
Every call to AddWork when the pool is full will block.

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs
+++ b/projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs
@@ -142,6 +142,12 @@ namespace RabbitMQ.Client
         public const string DefaultVHost = "/";
 
         /// <summary>
+        /// Default capacity of the work pool buffer, with zero meaning unlimited (value: 100).
+        /// </summary>
+        /// <remarks> PLEASE KEEP THIS MATCHING THE DOC ABOVE.</remarks>
+        public const int DefaultWorkPoolCapacity = 100;
+
+        /// <summary>
         ///  Default SASL auth mechanisms to use.
         /// </summary>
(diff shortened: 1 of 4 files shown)


projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs
```diff
@@ -142,6 +142,12 @@ public class ConnectionFactory : ConnectionFactoryBase, IConnectionFactory
         public const string DefaultVHost = "/";
 
         /// <summary>
+        /// Default capacity of the work pool buffer, with zero meaning unlimited (value: 100).
+        /// </summary>
+        /// <remarks> PLEASE KEEP THIS MATCHING THE DOC ABOVE.</remarks>
+        public const int DefaultWorkPoolCapacity = 100;
```
michaelklishin: Please rename this to `DefaultConsumerWorkPoolCapacity` for clarity.

michaelklishin:
Thank you for submitting this PR. We will review it when our team is back from a holiday break (around Jan 3rd).

bording:
@or150 The changes proposed here will undo a lot of the work introduced in #253 to remove lock contention and blocking, and I'd hate to see performance benefits it brought be undone.

michaelklishin:
@bording good point. Given that manual acknowledgements and QoS are sufficient to prevent consumer overload, I agree with you.

or150:
After taking another look at it, it would probably cause more harm than good and introduce some new bugs.
I'll try using manual acks instead, it's probably a better way to handle this.

PR #316, opened 2017-03-17
Add SetUri to IConnectionFactory interface

### LINKED ISSUE #312: Consider moving ConnectionFactory.SetUri to an exetension method on IConnectionFactory
If [`SetUri`](https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/master/projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs#L516) was an extension method on `IConnectionFactory`, it would be easier to test code by having a dependency on `IConnectionFactory` instead of `ConnectionFactory`.

kjnilsson:
Also removed Uri properties that cause issues with CLS compliance.

Fixes #312.

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs
+++ b/projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs
@@ -271,25 +271,6 @@ namespace RabbitMQ.Client
             }
         }
 
-
-        /// <summary>
-        /// Set connection parameters using the amqp or amqps scheme.
-        /// </summary>
-        [Obsolete("Deprecated: please use SetUri instead.")]
-        public String Uri
-        {
-            set { SetUri(new Uri(value, UriKind.Absolute)); }
-        }
-
-        /// <summary>
-        /// Set connection parameters using the amqp or amqps scheme.
-        /// </summary>
-        [Obsolete("Deprecated: please use SetUri instead.")]
-        public Uri uri
-        {
-            set { SetUri(value); }
-        }
-
         /// <summary>
         /// Dictionary of client properties to be sent to the server.
         /// </summary>
--- a/projects/client/RabbitMQ.Client/src/client/api/IConnectionFactory.cs
+++ b/projects/client/RabbitMQ.Client/src/client/api/IConnectionFactory.cs
@@ -86,6 +86,13 @@ namespace RabbitMQ.Client
         /// </summary>
         string VirtualHost { get; set; }
 
+
+
+        /// <summary>
+        /// Sets the AMQP Uri to be used for connections.
+        /// </summary>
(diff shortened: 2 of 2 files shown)


paulomorgado:
Is there an ETA for this?

Although this commit is mentioned on the release notes for [v4.1.3](https://github.com/rabbitmq/rabbitmq-dotnet-client/compare/v4.1.3...master), it doesn't seem to be on the [respective package](https://www.nuget.org/packages/RabbitMQ.Client/4.1.3).

michaelklishin:
This commit was merged into master. I don't know if this was intentional or not but #312's milestone reflects that correctly.

@paulomorgado you are looking at the changes from 4.1.3 to master. Those changes are not in 4.1.3 by definition.

kjnilsson:
@paulomorgado this commit is included in 5.0.0-pre.4 - released here: https://www.nuget.org/packages/RabbitMQ.Client/5.0.0-pre4

Your comparison above is showing you what has gone into master that isn't in the 4.x branch not the other way around.

michaelklishin:
GitHub displays what tags a commit is "included into" but only in a single commit view, not pull request commit view :/ https://github.com/rabbitmq/rabbitmq-dotnet-client/commit/7c6c2598e6bba2e46e530bfaf5343dba00dec639

paulomorgado:
Any ETA for a non-Pre version of the NuGet package that includes this?

michaelklishin:
When it's done.

paulomorgado:
I fully support that decision!

Any estimate on when it will be done?

😄

michaelklishin:
We don't have an ETA for 5.0 because we don't have the scope for it. Currently there isn't much new stuff in master to warrant a 5.0 release: I don't think we want to go all Chrome with major version bumps. That's in part why we publish preview releases reasonably frequently.

paulomorgado:
How about 4.2 for this change? 😄

michaelklishin:
@kjnilsson WDYT?

paulomorgado:
@kjnilsson WDYT?

kjnilsson:
@paulomorgado I think we shouldn't break compatibility of this interface until v5, however small. What is your specific use-case for this? Regarding the v5 release, one of the issues that is holding us up is: https://github.com/rabbitmq/rabbitmq-dotnet-client/issues/308 which is breaking our release process. I just checked and although there are workarounds it hasn't been addressed properly as yet.

paulomorgado:
@kjnilsson, my use case is creating a connection using an URI for a queue that I'm receiving from an API and I would like to inject an `IConnectionFactory` as a dependency instead of a `ConnectionFactory`.

paulomorgado:
In the meantime, how do I implement `SetUri` over `IConnectionFactory`?

I know where the host, port, user and password go. But where does the SSL information go?

PR #320, opened 2017-04-25
Recover recorded topology entities using separate channels

svrx:
As discussed in bug report this is my proposed fix for #254.

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/impl/AutorecoveringConnection.cs
+++ b/projects/client/RabbitMQ.Client/src/client/impl/AutorecoveringConnection.cs
@@ -940,9 +940,25 @@ namespace RabbitMQ.Client.Framing.Impl
             // 2. Recover queues
             // 3. Recover bindings
             // 4. Recover consumers
-            RecoverExchanges();
-            RecoverQueues();
-            RecoverBindings();
+
+            using (var recoveryModel = new AutorecoveringModel(this,CreateNonRecoveringModel()))
+            {
+                FixChannelOrphanedTopologyEntities(recoveryModel);
+
+                RecoverExchanges();
+                RecoverQueues();
+                RecoverBindings();
+            }
+        }
+
+        protected void FixChannelOrphanedTopologyEntities(AutorecoveringModel recoveryModel)
+        {
+            RecordedQueues.Values.Cast<RecordedEntity>()
+                .Union(RecordedExchanges.Values)
+                .Union(m_recordedBindings.Keys)
+                .Where(p => p.Model.IsClosed)
+                .ToList()
+                .ForEach(p => p.Model = recoveryModel);
         }
 
         protected void RecoverExchanges()
--- a/projects/client/RabbitMQ.Client/src/client/impl/AutorecoveringConnection.cs
+++ b/projects/client/RabbitMQ.Client/src/client/impl/AutorecoveringConnection.cs
@@ -941,9 +941,10 @@ namespace RabbitMQ.Client.Framing.Impl
             // 3. Recover bindings
             // 4. Recover consumers
 
-            using (var recoveryModel = new AutorecoveringModel(this,CreateNonRecoveringModel()))
+            using (var recoveryModel = new AutorecoveringModel(this, CreateNonRecoveringModel()))
             {
-                FixChannelOrphanedTopologyEntities(recoveryModel);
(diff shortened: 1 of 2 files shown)


michaelklishin:
We've asked this in #254: in what scenario does it make sense to recover entities on closed channels? We need a step by step description, so be as detailed as possible, please.

Channel shutdown provides a shutdown reason on the shutdown event, so it's possible to determine whether a channel was closed explicitly vs. e.g. due to a channel-level exception.
Even with this information generally available to channels already it's not clear to me when it makes sense to recover closed channels.

@kjnilsson

svrx:
Conceptually, it makes sense for me to recover queues, exchanges and bindings, not channels, that currently active consumers depend on, because otherwise they will just fail silently.

The scenario I have stumbled upon is the following, but others may exist:
1. App starts 
1. RabbitMQ required topology is declared (it's own channel immediately closed after topology is declared)
1. App subscribes to non-durable queue using auto recoverable connection (previously declared)
1. RabbitMQ cluster experiences issues, looses majority and shut-down (disaster recovery testing, but may happen effectively)
1. RabbitMQ cluster comes back up 
1. App reconnects to broker, but silently fails to resume queue consumption, which leaves the App in stale state.

I know in this scenario there are different work-arounds, that would allow to avoid the issue and I'm not blocked by this but would like to improve it, so others do not hit the same problems.

Meanwhile I was trying to think about my proposed fix in multiple scenarios:
* Single connection per App, scenario is fine as the recorded entities reflect current active state (my case)
* Multiple connections in same App, it may have issues since, different connections may have conflicting states and the recorded data has no temporal awareness that could allow to reconcile those

Given this I'm thinking of a different strategy, that basically would allow to recover only the entities that should be alive, following these steps:
1. On recovery, identify active consumers (already recorded)
2. Identify the queues the consumers depend on
3. Identify the bindings the refer these to active consumer queues
4. Identify exchanges referred by consumer active bindings
5. Recover all identified queues, binding and exchanges (channel orphaned ones, in a recovery only channel as used in pull request)

Alternatively, it can be left as it is and just make it clear on the documentation that you either have to do declaration and consumption on the same channel or should avoid using non-durable entities with auto recovering connections.

Please let me know your thoughts?

michaelklishin:
And why was

> it's own channel immediately closed after topology is declared

?

michaelklishin:
Please understand that topology recovery intends to only recover stuff that is still used. It is impossible for a library to know what is actually used, so currently the distinction is made by recovering channels that were not explicitly closed.

I don't think it's unreasonable to assume that

 * An explicitly closed channel assumes it's no longer needed, and likely so is everything that was declared on it
 * If you need certain entities to survive potential node restarts, they *must* be declared durable

So I am seeing more and more evidence of what @kjnilsson has suggested in #254: you are trying to make this client recover entities that should be durable to begin with.

svrx:
That is because, we're doing some checks around topology before starting consumers. 
Just tested and even if the queues are declared on the same channel it doesn't help, because the recorded entity is updated to use last channel where queue was declared, which could have been done by a publisher, that is likely gone when the recovery happens.

As you suggested, using them as durable entities avoids the problem and in my case will use that for now. 
But it raises the question, if that is the recommended approach, is it even worth for the recovery process to try to record and re-declare these entities, since it will likely fail to redeclare them?

Just pushed a last change, that includes changes to only recover topology required by currently active consumers at the time of failure.

Do you think that would make any sense?

kjnilsson:
If you chose to use a non-durable entities you have to be prepared for them disappearing after connection loss. I think it is ok to re-declare those entities if the channels they were declared on are still around. At least I don't feel it does any harm. IIRC many of the opinionated wrappers (EasyNetQ, NSB etc) around the client tend to re-declare any entities required when setting up a consumer as it is the only way to guarantee the required topology is there. Feels like we are entering the realm of RabbitMQ "best practices" here.

Topology-recovery is never going to be perfect as it is not a formally described system. Unless we produce a spec for exactly how it should work I'd be reluctant to make any further behavioural changes.

svrx:
I understand your reluctance to change it. 
But it's just inconsistent, even if you declare on same channel as consumer and for any reason on the same connection you declare the queue again to publish some messages, the recorded queue will not be recoverable as it will be hijacked by the producer.

The fact is the outcome for re-declaration only really matters for non persisted entities, for which we are reckoning that recovery is broken anyway.

Thanks you for being so responsive and supportive around this issue.

PR #322, opened 2017-05-11
Switch to the new [old] project format (csproj)

kjnilsson:
To do this the net451 build has been removed as it currently
isn't able to build this using mono.

[#145272475]


bording:
FYI the 5.0 release of mono includes MSBuild, and it can be used to build net451 and netstandard, instead of `dotnet build`. 5.0 is still in beta, but I was trying it out yesterday as part of the VS for Mac release.

bording:
@michaelklishin By just building netstandard1.5, you've bumped the miniumu .NET version to 4.6.1, not 4.5.2: https://docs.microsoft.com/en-us/dotnet/articles/standard/library

michaelklishin:
we plan to continue with 4.x releases for some time, so I wonder if we can actually require 4.6.1. Until .NET Core and Mono are out of the "kind-of-works-if-you-are-careful" maturity stage, supporting all those .NETs is getting too taxing on our time.

kjnilsson:
I feel we can add multi targeting back once it "just works" with the dotnet core tooling. I tried it today with mono 5 and it doesn't as yet. I'd like to get an RC out soon

Also I think the .NET requirement is 4.6.2 until v2.0 of the tooling (not confusing at all is it? :) )

bording:
@kjnilsson I'll have to double check, but I thought the 2.0 tooling was referring to the new csproj format, with 1.0 being project.json.

What did you try with the mono 5 stuff? I was able to build a sample project targeting `net452` and netcoreapp1.0` by doing the following:
```
dotnet restore
msbuild
```
And that was able to build both assemblies.

kjnilsson:
Oh right , I have to try that, thanks.

bording:
Bumping the minimum required version would bring us some pain, because we currently support 4.5.2 as our minimum, but we also like to make sure we have releases that work with the latest RabbitMQ client.

adamralph:
> Also I think the .NET requirement is 4.6.2 until v2.0 of the tooling

I think @kjnilsson may be correct. Previously, the .NET Standard table was different and it only had .NET FX 4.6.2 as supporting .NET STD 1.5. The table now shows .NET FX 4.6.1 as supporting .NET STD 1.5. It may well be the 2.0 tooling which actually implements this change.

bording:
@adamralph I was just about to comment that @kjnilsson was correct, and "tooling 2.0" is referring to the new tooling coming as part of .NET Core 2.0.

kjnilsson:
`msbuild` works better - I'm baffled as to why they didn't just make `dotnet build` work as well. Great I will update the PR tomorrow. Thanks!

bording:
> msbuild works better - I'm baffled as to why they didn't just make dotnet build work as well. Great I will update the PR tomorrow. Thanks!

I think it's just different release schedules not aligning yet. Mono needed to ship msbuild instead of xbuild, and include the SDKs to build .NET Core stuff first, Then the `dotnet build` tooling can be updated to be aware of it. 

I need to check if the .NET Core 2.0 preview release handles it differently, or if that change is still to come.

bording:
BTW, it looks like mono 5.0 is officially released now: http://www.mono-project.com/docs/about-mono/releases/5.0.0/

kjnilsson:
Ok this seems to work with net451 and our Pr build. Once this is merged I can move it through the build stages and produce a pre-release and then an RC.

PR #325, opened 2017-06-01
Add CreateConnection overload to IConnectionFactory

### LINKED ISSUE #324: Consider expanding IChannelFactory with method from ChannelFactory
The `ChannelFactory` has a [very handy method](https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/master/projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs#L422) for creating a connection where each hosts is declared with a port (and `SslOptions`):

```csharp
factory.CreateConnection(new List<AmqpTcpEndpoint>
{
	new AmqpTcpEndpoint
	{
		HostName = "server1",
		Port = 4672
	},
	new AmqpTcpEndpoint
	{
		HostName = "server2",
		Port = 4673
	},
	new AmqpTcpEndpoint
	{
		HostName = "server3",
		Port = 4674
	}
});
```
This method is not part of the `IConnectionFactory` interface. My dilemma is that I want to use the interface so that I can mock the calls in my unit tests but I also have the ability to specify port for each host. Is there any chance you could add this method to the interface?

Cheers!

pardahlman:
The method was implemented in ConnectionFactory, so the only thing done was to pull it up to the interface.

This fixes #324

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/api/IConnectionFactory.cs
+++ b/projects/client/RabbitMQ.Client/src/client/api/IConnectionFactory.cs
@@ -41,6 +41,7 @@
 using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using RabbitMQ.Client.Exceptions;
 
 namespace RabbitMQ.Client
 {
@@ -134,6 +135,21 @@ namespace RabbitMQ.Client
         /// </param>
         /// <returns></returns>
         IConnection CreateConnection(IList<string> hostnames, String clientProvidedName);
+
+        /// <summary>
+        /// Create a connection using a list of endpoints. By default each endpoint will be tried
+        /// in a random order until a successful connection is found or the list is exhausted.
+        /// The selection behaviour can be overriden by configuring the EndpointResolverFactory.
+        /// </summary>
+        /// <param name="endpoints">
+        /// List of endpoints to use for the initial
+        /// connection and recovery.
+        /// </param>
+        /// <returns>Open connection</returns>
+        /// <exception cref="BrokerUnreachableException">
+        /// When no hostname was reachable.
+        /// </exception>
+        IConnection CreateConnection(IList<AmqpTcpEndpoint> endpoints);
 
         /// <summary>
         /// Advanced option.


projects/client/RabbitMQ.Client/src/client/api/IConnectionFactory.cs
```diff
@@ -136,6 +137,21 @@ public interface IConnectionFactory
...
+        /// The selection behaviour can be overriden by configuring the EndpointResolverFactory.
+        /// </summary>
+        /// <param name="endpoints">
+        /// List of endpoints to use for the initial
+        /// connection and recovery.
+        /// </param>
+        /// <returns>Open connection</returns>
+        /// <exception cref="BrokerUnreachableException">
+        /// When no hostname was reachable.
+        /// </exception>
```
adamralph: Should exceptions be documented on interfaces, or only on implementations? I've never been sure, and I've heard arguments both ways.

I guess you could also argue that some of the content in the `summary` tag is implementation specific.

michaelklishin: In this case I'm leaning towards a "yes". `BrokerUnreachableException` is something mock implementations can throw, too.

kjnilsson: @adamralph what would Liskov say? :)  I suspect _yes_ as it is very much something you'd have to account for in making any implementation "replacable" without changes to the consuming code. That said specifying exactly what exceptions should be allowable for any one implementation isn't always easy to do up front.

I agree that the summary comment is implementation specific in this case it depends on the implementation of  `IEndpointResolver`.

kjnilsson:
I've created a pre-release that includes change: https://www.nuget.org/packages/RabbitMQ.Client/5.0.1-pre3

projects/client/RabbitMQ.Client/src/client/api/IConnectionFactory.cs
adamralph: @kjnilsson good point. I guess, from that point of view, the exception docs can be considered part of the interface definition, requiring that every implementation behaves that way. The concrete way to do that would be something like checked exceptions in Java, but that is its own can of worms. :-)

> I agree that the summary comment is implementation specific in this case it depends on the implementation of `IEndpointResolver`.

Do you think it's worth changing the comment?

kjnilsson: Yes it should probably be changed.

adamralph: I raised https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/332

PR #338, opened 2017-07-12
Dispose of sockets that failed to connect

### LINKED ISSUE #337: ConnectionFactory doesn't release socket on exception in CreateConnection
RabbitMq client doesn't release socket in Linux in case when RabbitMq is unreachable. 

My service opens Http endpoint and tries to connect to RabbitMq. **ConnectionFactory.CreateConnection** throws **BrokerUnreachableException** and my program retries to connect to RabbitMq after some time. Since RabbitMq does not exists on the machine (it is a valid case for my service) my application can exceed Linux limitation on sockets and it leads to impossibility to establish http connection with my service.

Linux version: Ubuntu 16.04.1 LTS
dotnet core version: 1.0.4
rabbitmq-dotnet-client version: 4.1.3

Example (without http endpoint): [Gist](https://gist.github.com/lomomike/f21de9a3f47810331a9f83deb1cfbdae)

Run result:
```
PID = 5476
Press any key to exit
[3] Cannot connect to RabbitMq
[4] Cannot connect to RabbitMq
[3] Cannot connect to RabbitMq
[4] Cannot connect to RabbitMq
[3] Cannot connect to RabbitMq
[4] Cannot connect to RabbitMq.
...
```
In second terminal:
```
lomomike@lomomike-pc:~$ sudo lsof -p 5476  | grep sock | wc -l
207
lomomike@lomomike-pc:~$ sudo lsof -p 5476 | grep sock | wc -l
239
lomomike@lomomike-pc:~$ sudo lsof -p 5476 | grep sock | wc -l
271
lomomike@lomomike-pc:~$ sudo lsof -p 5476 | grep sock | wc -l
464
lomomike@lomomike-pc:~$ sudo lsof -p 5476 | grep sock | wc -l
476
```

michaelklishin:
This makes sure that we dispose of sockets that failed to connect plus does some refactoring
and spring cleanup of the code in the area. This is for `5.0` only, so changes that will require
recompilation are intentional.

Note that WinRT version of `SocketFrameHandler` already does this.

Fixes #337.


projects/client/RabbitMQ.Client/src/client/impl/TcpClientAdapter.cs
```diff
@@ -41,8 +41,16 @@ public TcpClientAdapter(Socket socket)
 
         public virtual void Close()
         {
-            if(sock != null)
+            // TODO: Socket.Close() would be a better option here. MK.
```
kjnilsson: How come? Socket.Close simply calls Socket.Dispose anyway: https://referencesource.microsoft.com/#System/net/System/Net/Sockets/Socket.cs,1140

michaelklishin: I didn't check the source. The docs didn't make me believe they are identical. Will remove this comment then.

Issue #256, opened 2016-09-11
CreateConnection() seemingly fails to connect when run in an async context with Nito.AsyncEx


JohannesRudolph:
I'm using the ``Nito.AsyncEx`` library (https://github.com/StephenCleary/AsyncEx) to provide an Async Context for my workers (.netcore console app). 

See this thread for more info on why this is necessary and a good idea: http://stackoverflow.com/questions/9208921/cant-specify-the-async-modifier-on-the-main-method-of-a-console-app/9212343#9212343

In any case, this throws a ``BrokerUnreachableException``
```
        static void Main( string[] args )
        { 
            AsyncContext.Run( () =>var mq = new RabbitMqConfiguration( isTest: false ).ConnectionFactory.CreateConnection(); );
        }
```

while this does not: 
```
        static void Main( string[] args )
        {  
             var mq = new RabbitMqConfiguration( isTest: false ).ConnectionFactory.CreateConnection();
        }
```

Probably a bug in RabbitMqClient. I've tried versions 4.0, 4.1 at latest rc1 at the time of writing.

michaelklishin:
Please post questions to rabbitmq-users or Stack Overflow. RabbitMQ uses GitHub issues for specific actionable items engineers can work on, not questions. Thank you.

michaelklishin:
There is no evidence that this is a RabbitMQ client bug because it only can be reproduced when a particular external library is used. Please post a runnable code example to the mailing list and maybe someone will be able to figure out what's going on.

JohannesRudolph:
Your call. This bug report contains all facts needed to reproduce. Replace ``RabbitMqConfiguration(...).ConnectionFactory`` with 
```
new ConnectionFactory()
            {
                HostName = "192.168.99.100",
                Port = AmqpTcpEndpoint.UseDefaultPort,
                Protocol = Protocols.DefaultProtocol,
                RequestedConnectionTimeout = 2000, // fail the test early
                VirtualHost = "/",
                UserName = "user",
                Password ="password"
}
``` 
or whatever is appropriate in your dev environment. [Nito.AsyncEx](https://www.nuget.org/packages/Nito.AsyncEx) is a very popular package to write proper async context enabled console apps.

Maybe rabbitmq forgets to call ``.ConfigureAwait(false)`` on some internal async code? This is a common problem for libraries that can't assume too much about the async context they run in. See also https://msdn.microsoft.com/en-us/magazine/jj991977.aspx 

[...]

PR #349, opened 2017-09-11
add configure await in tcp client adapter

jacobpovar:
It seems that lack of one `.ConfigureAwait(false)` call causes deadlock in MassTransit project when running under corefx. Can we change it?
See https://github.com/MassTransit/MassTransit/issues/946.


michaelklishin:
Related to #240 and potentially #256.

PR #354, opened 2017-09-19
Reduce lock contention for frame writes (with extra tests)

michaelklishin:
This is #350 by @YulerB and @vendre21 with some additional tests from our team. See #350 for the original discussion.

The idea is the following:

 * Frames and frame sets (for frames that carry content and need, ahem, content framing) are written in a single operation to a memory-backed stream
 * `BufferedStream` is no longer used (it was not efficiently used for writes anyway)
 * The above seeks to guarantee that socket stream writes cannot be interleaved in a way that will confuse RabbitMQ protocol parser
 * As such, locking is no longer necessary

This PR was created to simplify the final rounds of QA:

 * Extra tests if our team comes up with any
 * A fresh discussion without a dozen of comments about line endings :)


michaelklishin:
@bording @danielmarbach how can our team make it easier for you folks to test NServiceBus against this branch? So far it's holding well on our own integration tests.

bording:
@michaelklishin I've been meaning to try out these changes, but haven't gotten to them yet.

The only way I can think to make trying it out easier would be for your to publish some sort of prerelease package, perhaps to a custom MyGet feed instead of nuget.org.

Did you see @danielmarbach's request on the other PR about adding tests with several different payload sizes? Seems like a good idea to validate this continues to be better as message size increases.

michaelklishin:
@bording OK, we had an AppVeyor feed for development releases but I'm afraid that thing hasn't been updated in a while.

I am going to add more tests with different payload sizes, very good idea @danielmarbach. Any particular values you have in mind?

michaelklishin:
I did some benchmarking with 20 runs of the concurrent connection access test suite on two machines (from 2017 and 2013).

In one case, median went down from 1.35s (`master`) to 1.04s (this PR). In the other, from 1.41s to 1.06s. While that's not a lot of data points, on a CPU (32 threads, together with RabbitMQ completely max out 8 hyperthreads on both machines) + loopback I/O heavy workload, removing the lock makes a nice difference.

bording:
@michaelklishin That sounds really good! I spent some time today getting some of my testing setup ready to try out the changes here, so I should have something to report back tomorrow.

projects/client/RabbitMQ.Client/src/client/impl/Connection.cs
```diff
@@ -1099,7 +1099,7 @@ public void HeartbeatWriteTimerCallback(object state)
                     if (!m_closed)
                     {
                         WriteFrame(m_heartbeatFrame);
-                        m_frameHandler.Flush();
+                        //m_frameHandler.Flush();
```
kjnilsson: remove commented code

projects/client/RabbitMQ.Client/src/client/impl/ModelBase.cs
```diff
@@ -465,11 +464,11 @@ public void ModelSend(MethodBase method, ContentHeaderBase header, byte[] body)
         {
             if (method.HasContent)
             {
-                lock (m_flowSendLock)
-                {
+                //lock (m_flowSendLock)
```
kjnilsson: remove commented code

projects/client/RabbitMQ.Client/src/client/impl/ModelBase.cs
```diff
@@ -908,11 +912,11 @@ public void HandleChannelFlow(bool active)
             }
             else
             {
-                lock (m_flowSendLock)
-                {
+                //lock (m_flowSendLock)
```
kjnilsson: remove commented code

projects/client/RabbitMQ.Client/src/client/impl/Connection.cs
```diff
@@ -767,7 +767,7 @@ public void MainLoopIteration()
             }
         }
 
-        public void NotifyHeartbeatListener()
+        public  void NotifyHeartbeatListener()
```
kjnilsson: remove double space

projects/client/RabbitMQ.Client/src/client/impl/ModelBase.cs
```diff
@@ -88,7 +87,7 @@ public abstract class ModelBase : IFullModel, IRecoverable
 
         private EventHandler<EventArgs> m_recovery;
 
-        public IConsumerDispatcher ConsumerDispatcher { get; private set; }
+        public  IConsumerDispatcher ConsumerDispatcher { get; private set; }
```
kjnilsson: remove extra space

projects/client/RabbitMQ.Client/src/client/impl/CommandAssembler.cs
```diff
@@ -98,20 +98,19 @@ public Command HandleFrame(Frame f)
                 }
                 case AssemblyState.ExpectingContentBody:
                 {
-                    if (f.Type != Constants.FrameBody)
+                    if (!f.IsBody())
                     {
                         throw new UnexpectedFrameException(f);
                     }
-                    byte[] fragment = f.Payload;
```
kjnilsson: Is there real value in removing this variable? I would have thought (hoped) the compiler would optimise this away anyway.

YulerB:
Since the frame sizes are calculable, might be wothwhile setting the capacity on the memory streams to avoid resizing and memory fragmentation.  

For example, 

1. HEADER           var ms = new MemoryStream(8);
2. FRAME->         var ms = new MemoryStream(frame.Size);
3. FRAMESET->   var ms = new MemoryStream(frames.Sum(_=>_.Size));

danielmarbach:
@YulerB if you worry about allocs then better no use Linq ;)

YulerB:
@danielmarbach  Linq gets a bad rap, but I know what you mean.

Many of the Framesets with 4k, which would require 2 re-allocs of the byte buffer in the MemoryStream.

Many of the message like Ack are much smaller than the default 1k, so we have over-allocation of memory .

This might avoid some of the memory thrashing.

To avoid linq, here is an extension method.

        public int Sum<T>(this IEnumerable<T> list, Func<T, int> expression)
        {
            int sum = 0;
            foreach (var item in list)
            {
                sum += expression(item);
            }
            return sum;
        }

kjnilsson:
@bording @michaelklishin I've set our ci to push each PR build to myget - would this work for testing this PR? https://www.myget.org/feed/Packages/rabbitmq

bording:
@kjnilsson Yes, having a MyGet feed that the results of each PR will be helpful for the future.

FYI, assuming your MyGet account name is `rabbitmq`, then the feed URL is `https://www.myget.org/F/rabbitmq/api/v3/index.json`. It doesn't look like any packages have been pushed to it yet, though!

bording:
I have done some testing of these changes with a locally built copy of the client. I am seeing a small benefit on some tests that specifically ensure that we're publishing messages with multiple channels from the same connection.

I'm also not seeing any performance regressions, so overall it seems like this PR is 👍 from me, with the caveat that I haven't spent a lot of time actually looking at the content of the changes.

michaelklishin:
@bording my personal concerns were about concurrency hazard safety but I assume if there were any failures in your test suites you'd have mentioned it. I'll do some last round naming and test structure changes before merging.

bording:
@michaelklishin I'll try and run a few more rounds of tests to be sure, but I've not seen any problems so far.

michaelklishin:
In the end, so far we didn't manage to run into any concurrency hazards with this implementation and indeed there's a nice efficiency gain with concurrent writes on a shared connection. Thank you, @YulerB and @vendre21! Your contribution will be availalbe in the `5.1.0` version of the client :)

michaelklishin:
We published a [5.1.0-pre1](https://www.nuget.org/packages/RabbitMQ.Client/5.1.0-pre1) to nuget.org for others to try this change out.

YulerB:
Thanks for you help on this @michaelklishin, @bording, @danielmarbach, @kjnilsson, @lukebakken, @vendre21.

PR #360, opened 2017-10-05
Async Support + Big-Endian support

YulerB:
Updates to support an Async api asside to the sync api and also to support operating systems with big endian byte order.


projects/client/RabbitMQ.Client/src/client/content/BytesMessageBuilder.cs
```diff
@@ -1,168 +1,172 @@
...
-//  basis, WITHOUT WARRANTY OF ANY KIND, either express or implied. See
-//  the License for the specific language governing rights and
-//  limitations under the License.
-//
-//  The Original Code is RabbitMQ.
-//
-//  The Initial Developer of the Original Code is Pivotal Software, Inc.
-//  Copyright (c) 2007-2016 Pivotal Software, Inc.  All rights reserved.
-//---------------------------------------------------------------------------
+//// This source code is dual-licensed under the Apache License, version
```
sigiesec: The doubled comment prefix should be reverted.

sigiesec: The whole file is commented out. Should it rather be deleted?

projects/client/RabbitMQ.Client/src/client/content/BytesMessageReader.cs
```diff
@@ -43,12 +43,12 @@ namespace RabbitMQ.Client.Content
     /// <summary>
     /// Analyzes AMQP Basic-class messages binary-compatible with QPid's "BytesMessage" wire encoding.
     /// </summary>
-    public class BytesMessageReader : BasicMessageReader, IBytesMessageReader
+    public class BytesMessageReader : BasicMessageReader//, IBytesMessageReader
```
sigiesec: Why is the interface commented out? Is it no longer necessary to implement the interface?

In that case it should be deleted rather than commented out.

projects/client/RabbitMQ.Client/src/client/impl/AutorecoveringModel.cs
```diff
@@ -556,10 +592,31 @@ public void HandleConnectionUnblocked()
                 arguments);
         }
 
+        public Task _Private_BasicConsumeAsync(string queue,
+    string consumerTag,
```
sigiesec: Indentation is broken

projects/client/RabbitMQ.Client/src/client/impl/AutorecoveringModel.cs
```diff
@@ -778,19 +1034,57 @@ public void BasicCancel(string consumerTag)
...
         {
             return m_delegate.BasicGet(queue, autoAck);
         }
 
+        public Task<BasicGetResult> BasicGetAsync(string queue,
+            bool autoAck)
+        {
+            return m_delegate.BasicGetAsync(queue, autoAck);
+        }
+
```
sigiesec: Extra blank lines

projects/client/RabbitMQ.Client/src/util/NetworkBinaryReader.cs
```diff
@@ -53,147 +54,207 @@ namespace RabbitMQ.Util
     /// Relies on BinaryReader always being little-endian.
```
sigiesec: This remark is no longer true,  and should be removed...

projects/client/RabbitMQ.Client/src/util/NetworkBinaryReader.cs
```diff
@@ -53,147 +54,207 @@ namespace RabbitMQ.Util
...
     /// </para>
     /// </remarks>
-    public class NetworkBinaryReader : BinaryReader
+    public class NetworkBinaryReader 
     {
-        // Not particularly efficient. To be more efficient, we could
```
sigiesec: I think it were good to keep an explanation similar to this comment, that explains why BinaryReader is not used here as a basis.

projects/client/RabbitMQ.Client/src/client/content/BytesMessageBuilder.cs
YulerB: Don't have visual studio 2017, had to upload the changes, had to comment out files/items not used.

projects/client/RabbitMQ.Client/RabbitMQ.Client.csproj
```diff
@@ -28,7 +28,7 @@
 
   <ItemGroup>
     <Compile Remove="build\**\*" />
-    <Compile Include="..\..\..\gensrc\autogenerated-api-0-9-1.cs" Exclude="build\**\*;bin\**;obj\**;**\*.xproj;packages\**" />
```
michaelklishin: Why was this line changed?

michaelklishin:
If this client can be adapted to require only VS 2015, we are open to doing it. Some of our team members use Visual Studio Code with great success.

bording:
> If this client can be adapted to require only VS 2015, we are open to doing it. Some of our team members use Visual Studio Code with great success.

@michaelklishin That's not possible. The new project system isn't supported by VS 2015 and is never going to be either.

michaelklishin:
@bording thanks for confirming. Well, then perhaps Visual Studio Code can be a more convenient solution that copying code from a modified private clone.

projects/client/RabbitMQ.Client/RabbitMQ.Client.csproj
YulerB: To allow me to compile the project.
Included the change to allow a diff on the output from the templating.
The change makes a class partial to allow inclusion of async overloads in another file.
I get why your doing this, but is it really necessary?

PR #375, opened 2017-11-30
RabbitMQ.Client.csproj: add support for targeting 4.6.2 framework

kforbesni:
RabbitMQ added a new dependency on an obsolete Microsoft assembly
(Microsoft.Diagnostics.Tracing.EventSource 1.1.28). It is described as a "stop
gap" solution by Microsoft
(https://www.nuget.org/packages/Microsoft.Diagnostics.Tracing.EventSource) This
functionality is provided in the framework as of .NET 4.6, so providing a
version of the RabbitMQ dotnet client targeted to the 4.6.2 framework
(Microsoft's currently recommended framework) allows deployment of the RabbitMQ
dotnet client without having to redistribute the stop gap Microsoft assembly.

Signed-off-by: Keenan Forbes <keenan.forbes@ni.com>

## Proposed Changes

see description

## Types of Changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist

- [x ] I have read the `CONTRIBUTING.md` document

## Further Comments

considered, etc.

We were able to build the net462 version of the RabbitMQ.Client.dll successfully with this change (adding appropriate "-f net462" and "-c Release" to the dotnet command; however, without the signing key we're unable to fully verify or generate the final output (and don't know how to generate a version of the actual NuGet package)

### DIFF
--- a/projects/client/RabbitMQ.Client/RabbitMQ.Client.csproj
+++ b/projects/client/RabbitMQ.Client/RabbitMQ.Client.csproj
@@ -6,7 +6,7 @@
   <PropertyGroup>
     <Description>The RabbitMQ .NET client is the official client library for C# (and, implicitly, other .NET languages)</Description>
     <VersionPrefix>5.0.0</VersionPrefix>
-    <TargetFrameworks>net451;netstandard1.5</TargetFrameworks>
+    <TargetFrameworks>net451;net462;netstandard1.5</TargetFrameworks>
     <NoWarn>$(NoWarn);CS1591</NoWarn>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
@@ -35,6 +35,10 @@
     <PackageReference Include="Microsoft.Diagnostics.Tracing.EventSource.Redist" Version="1.1.*" />
     <Reference Include="System" />
     <Reference Include="Microsoft.CSharp" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net462' ">
+    <Reference Include="System" />
   </ItemGroup>
 
   <PropertyGroup Condition=" '$(TargetFramework)' == 'netstandard1.5' ">


pivotal-issuemaster:
@kforbesni Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-dotnet-client&pullRequestId=375)!

[Click here](https://cla.pivotal.io/sync/rabbitmq?repositoryId=rabbitmq/rabbitmq-dotnet-client&pullRequestId=375) to manually synchronize the status of this Pull Request.

kjnilsson:
netstandard 1.5 is already compatible with .NET 4.6.2 - why do we need another target?

kforbesni:
netstandard1.5 does not appear to support RemoteCertificateValidationCallback, used in RabbitMQ.Client.SslOption. I've been trying to research solutions for this but so far have been hitting a brick wall. The functionality doesn't appear to have been added until netstandard2.0. I'm consulting with our Microsoft contacts to see what other options might be available.
I created a stock 4.6.2 ConsoleApp, installed the 5.0.1 RabbitMQ.Client NuGet package, then manually modified the HintPath to point to netstandard1.5 version of the dll like so:
    <Reference Include="RabbitMQ.Client, Version=5.0.0.0, Culture=neutral, PublicKeyToken=89e7d7c5feba84ce, processorArchitecture=MSIL">
      <HintPath>..\packages\RabbitMQ.Client.5.0.1\lib\netstandard1.5\RabbitMQ.Client.dll</HintPath>
      <Private>True</Private>
    </Reference>

Then added the following code snippet (yes I know it’s not functionally correct, just trying to minimize code):
        {
            var factory = new ConnectionFactory();
            factory.Ssl.CertificateValidationCallback += ValidateServerCertificate;
        }

        private static bool ValidateServerCertificate(object sender, X509Certificate certificate, X509Chain chain, SslPolicyErrors sslPolicyErrors)
        {
            return true;
        }

which results in the following error:
The type 'RemoteCertificateValidationCallback' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Net.Security, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.

One of our in-house Microsoft experts indicated that there are many such deficiencies in netstandard1.x and that it wasn't really usable until netstandard2.0

michaelklishin:
@kforbesni thank you for the details. @kjnilsson @danielmarbach @adamralph @bording do you think supporting .NET 4.5.1 is important at this point or can we bump the version to 6.0 and begin requiring 4.6?

michaelklishin:
Looks like I've misunderstood the changes here. So this does not *require* 4.6, only allows for targeting.

kforbesni:
yes, this in no way requires 4.6, only adds another target framework (I don't like the bloat in the package, but also don't like redistributing the obsolete Microsoft stop-gap assembly)

michaelklishin:
@kforbesni I agree with you, let's see what other maintainers and major contributors say :)

bording:
Taking a closer look at the Microsoft.Diagnostics.Tracing.EventSource package and what this PR is trying to do reveals some interesting things to consider.

With the current targets, `net451;netstandard1.5`, all versions of the .NET Framework will choose the `net451` assembly, since that is a closer framework match than a .NET Standard assembly. That means any version of the framework will pick up the EventSource dependency.

The `netstandard1.5` assembly is pretty much only going to be chose when running .NET Core (and the other platforms that support it like UWP and Xamarin), so there's no way to avoid the package when running on the .NET Framework.

The "simplest" thing to do here would be to move to `netstandard2.0` as a single target, and then all of these APIs should just be available everywhere. That means the minium .NET Framework version supported by the client would be `net461`.

However, for the next NServiceBus release, we're targeting both `net452` and `netstandard2.0`, so having the RabbitMQ client just supporting `netstandard2.0` would be a problem for us.

So, to be able to continue supporting `net452` but also let `net46` and greater not have the EventSource package dependency, the only way is to add a third target.

bording:
@kjnilsson Is the EventSource package even required? There is a version of that class in `net451`. Is the client using features that aren't in the `net451` version but are in the package/`net46`+ version?

kjnilsson:
@bording it uses the `EventSourceSettings` that is only available since 4.6. there may well be a way around that with some judicious application of compiler directives.

kforbesni:
I wanted to follow up on this and see what the plan is for future RabbitMQ client releases. For our current shipping product we can stick with the 3.5.6 client; however, we'd really like to have a plan moving forward.

michaelklishin:
@kforbesni if you question is "will this PR be accepted", I don't see a consensus. Otherwise the plan is "release 5.1.0 in the next month or so" and then more releases, eventually requiring .NET 4.6.2 (but that can take a year or more). .NET client doesn't have to be upgraded in lock step with the server and 3.5.x is not getting any more releases. Most .NET users managed to move beyond 3.x clients.

kjnilsson:
I think the way forward for this is to look at whether it is possible to factor out the requirement on the dependency for the `net451` build.

rwkarg:
Another logging option that is also standard is Microsoft.Extensions.Logging.

It has a netstandard1.1 and 2.0 package.
The 1.1 package can be used back to .NET Framework 4.5

That would, though, be a bigger change than just adding an additional target.

michaelklishin:
@rwkarg thanks. That sounds like something we can take a closer look at. @kjnilsson @bording any reasons for us not to consider `Microsoft.Extensions.Logging`?

bording:
The `Microsoft.Extensions.Logging` package is the logging abstraction built for ASP.NET Core, though it can be used outside of that. It supports multiple logging providers, including one that works with EventSource.

While there is a version of the package that works with `net451`, it's not doing anything that couldn't be done with the EventSource usage that's already in the codebase. As @kjnilsson suggested, it should be possible to make the `net451` code not rely on features that come from the Microsoft.Diagnostics.Tracing.EventSource package.

From a quick look at the existing code, the main thing would be not using the self-describing event format, and instead use the manifest option that `net451` supports.

kjnilsson:
If we can refactor the current use of EventSource it would be much preferred to take a dependency on anything else. I will take a look at the feasibility of that.

kjnilsson:
I have created this PR for your review:

https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/387

rwkarg:
M.E.L is an abstraction for logging that allows many sinks to be added on. This allows simply adding a console logger to get stdout logging (the standard for 12-factor apps) or adding EventSource, Syslog, NLog or whatever even a custom sink that an application framework may have in place.

Does EventSource allow attaching a sink for other frameworks? Specifically, I have need for both Syslog and Console in different products.

kjnilsson:
@rwkarg yes just implement an event listener. See this for an example:

https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/master/projects/client/RabbitMQ.Client/src/logging/RabbitMqConsoleEventListener.cs

rwkarg:
My concern is that M.E.L already has a wide range of sinks that are already written and maintained. What do we get out of using EventSource other than requiring that users write their own EventListeners?

To @bording's comment on M.E.L.:
"...it's not doing anything that couldn't be done with the EventSource usage that's already in the codebase."

M.E.L does do something that couldn't be done with EventSource and that is use an existing ecosystem of already created and maintained sinks for logging frameworks that many projects are already using.

I can understand if there is not a desire to change logging on a new major version that was just released. If that's the reason for not wanting to change then I apologize for not providing input sooner :D
Otherwise, M.E.L. does look like the better and more adopted option for logging in shared libraries without subjecting specific sink implementations on consumers.

bording:
@rwkarg My comment that you quoted was about the EventSource provider for the `Microsoft.Extensions.Logging` package, not the entire logging abstraction itself!

Yes, bringing in the Microsoft.Extensions.Logging abstraction would open up more than just being able to use EventSource, but that's not really what this conversation is about. It's about removing the current depedency on the `Microsoft.Diagnostics.Tracing.EventSource.Redist` package. My point was that introducing `Microsoft.Extensions.Logging` doesn't really solve that.

Take a look at #357 for more discussion about adding another logging abstraction.

kjnilsson:
I just merged a PR that removes the offending dependency from the net451 target. I will close this now. Let me know if there is anything here that needs further discussion.

https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/387

PR #377, opened 2017-12-06
Collect all exceptions when selecting an endpoint

### LINKED ISSUE #376: EndpointResolverExtensions-SelectOne trashes exception stack trace
```
        public static T SelectOne<T>(this IEndpointResolver resolver, Func<AmqpTcpEndpoint, T> selector)
        {
            var t = default(T);
            Exception exception = null;
            foreach(var ep in resolver.All())
            {
                try
                {
                    // ...
                }
                catch (Exception e)
                {
                    exception = e;
                }
            }

            if(Object.Equals(t, default(T)) && exception != null)
            {
                throw exception;
            }

            return t;
    }
```

Here, inside the loop the code captures exceptions from attempts to connect to the endpoints. It remembers the last one, and at the end, if it didn't get a successful result, it rethrows this last exception.

However, in doing so, it trashes the stack trace of the exception, making debugging the underlying problem hard or impossible. (In my particular case, I'm getting a TypeInitializerException complaining that it can't initialize Interop.Http - but RabbitMQ has, according to my decompiler, no direct or indirect dependency on System.Net.Http ...)

It should do one of two things:

Simple: It should rethrow the exception using the ExceptionDispatchInfo method:

```
ExceptionDispatchInfo.Capture(exception).Throw();
```

More complicated but better: it should capture all exceptions and wrap them in an AggregateException:

```
        public static T SelectOne<T>(this IEndpointResolver resolver, Func<AmqpTcpEndpoint, T> selector)
        {
            var t = default(T);
            var exceptions = new List<Exception>();
            foreach(var ep in resolver.All())
            {
                try
                {
                    // ...
                }
                catch (Exception e)
                {
                    exceptions.Add(e);
                }
            }

            if(Object.Equals(t, default(T)) && exceptions.Count > 0)
            {
                throw new AggregateException(exceptions);
            }

            return t;
    }
```

CornedBee:
## Proposed Changes

This fixes #376 in the second way proposed there. It not only preserves the stack trace of the caught exception, but collects all exceptions caught while trying out endpoints, and throws them all in an AggregateException if no suitable endpoint can be found.

Note that this is a user-visible behavior change that might break some edge cases: the InnerException of the BrokerUnreachableException is now an AggregateException instead of the last exception thrown from some endpoint (such as a Client.Exceptions.ConnectFailureException).

## Types of Changes

- [x] Bugfix (non-breaking change which fixes issue #376)
- [x] Documentation (correction or otherwise)

## Checklist

I have run the tests with a running RabbitMQ, but not within the full umbrella test suite, so everything that needs rabbitmqctl failed locally.

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/api/IEndpointResolverExtensions.cs
+++ b/projects/client/RabbitMQ.Client/src/client/api/IEndpointResolverExtensions.cs
@@ -49,7 +49,7 @@ namespace RabbitMQ.Client
         public static T SelectOne<T>(this IEndpointResolver resolver, Func<AmqpTcpEndpoint, T> selector)
         {
             var t = default(T);
-            Exception exception = null;
+            var exceptions = new List<Exception>();
             foreach(var ep in resolver.All())
             {
                 try
@@ -62,13 +62,13 @@ namespace RabbitMQ.Client
                 }
                 catch (Exception e)
                 {
-                    exception = e;
+                    exceptions.Add(e);
                 }
             }
 
-            if(Object.Equals(t, default(T)) && exception != null)
+            if(Object.Equals(t, default(T)) && exceptions.Count > 0)
             {
-                throw exception;
+                throw new AggregateException(exceptions);
             }
 
             return t;
--- a/projects/client/RabbitMQ.Client/src/client/exceptions/BrokerUnreachableException.cs
+++ b/projects/client/RabbitMQ.Client/src/client/exceptions/BrokerUnreachableException.cs
@@ -47,8 +47,8 @@ namespace RabbitMQ.Client.Exceptions
     ///ConnectionFactory.CreateConnection attempt.</summary>
     public class BrokerUnreachableException : IOException
     {
-        ///<summary>Construct a BrokerUnreachableException. The inner exception is associated
-        ///with only one connection attempt.</summary>
+        ///<summary>Construct a BrokerUnreachableException. The inner exception is
+        ///an AggregateException holding the errors from multiple connection attempts.</summary>
         public BrokerUnreachableException(Exception Inner)
             : base("None of the specified endpoints were reachable", Inner)
         {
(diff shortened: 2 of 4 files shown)


pivotal-issuemaster:
@CornedBee Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-dotnet-client&pullRequestId=377)!

[Click here](https://cla.pivotal.io/sync/rabbitmq?repositoryId=rabbitmq/rabbitmq-dotnet-client&pullRequestId=377) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@CornedBee Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-dotnet-client&pullRequestId=377)!

kjnilsson:
This looks ok to me. I'm not sure if changing the type of an inner exception qualifies as a breaking change as nowhere in the API are we actually documenting which inner exceptions form part of the API. So by that extension and `AggregateException` is fine. It's a SemVer grey area.

bording:
@kjnilsson As long as there is an inheritance hierarchy between the types, going from less specific to more specific is okay, but going from more specific to less specific would be a break.

Exception -> AggregateException `OK`
AggregateException -> Exception  `BREAKING CHANGE`

bording:
I've only looked at the actual changes showing up in the PR diff and not elsewhere in the code, but I assume that the exception throw by `SelectOne` is caught elsewhere and then wrapped in `BrokerUnreachableException` and rethrown before the enduser sees it?

kjnilsson:
@bording exactly at the type level it is non-breaking, however I still think Liskov would have a case here. It would be possible that some users may have logic hanging off the type of the inner exception and this will break their code. That is the problem with throwing exceptions - there is no proper way to communicate which may possibly be thrown except documentation.

michaelklishin:
Merging per discussion with @kjnilsson. We will decide on what version bump (if any) this should warrant at a later point.

kjnilsson:
there is a pre release available that contain this change: https://www.nuget.org/packages/RabbitMQ.Client/5.1.0-pre2

PR #387, opened 2018-01-26
Use self describing event source format for .NET core only

kjnilsson:
Only simple error logging is supported for net451

Remove Microsoft.Diagnostics.Tracing.EventSource.Redist dependency for net451

### DIFF
--- a/projects/client/RabbitMQ.Client/RabbitMQ.Client.csproj
+++ b/projects/client/RabbitMQ.Client/RabbitMQ.Client.csproj
@@ -32,7 +32,6 @@
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'net451' ">
-    <PackageReference Include="Microsoft.Diagnostics.Tracing.EventSource.Redist" Version="1.1.*" />
     <Reference Include="System" />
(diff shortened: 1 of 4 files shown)


kjnilsson:
@kforbesni - here is a PR to remove the event source dependency for net451 - I have only been able  to test it for .NET core - are you able to give this a test on net451?

@bording - WDYT?

bording:
@kjnilsson LGTM 

I tried out the `RabbitMqConsoleEventListener` with a `net451` project, and everything appears to be working fine. Being able to remove a dependency is always a good thing!

The downside of this change is that it's definitely a breaking change. If this goes in, the version should definitely be bumping to 6.0.0.

kjnilsson:
Thanks for checking!

Breaking because of the change to the Error method? Would anyone really
depend on this?

bording:
@kjnilsson It's not just the Error method. The base types of all of those classes have changed from the ones in the package to the ones in the Framework, so they are completely different types now from a binary compatibility standpoint.

kjnilsson:
This has been merged in (with a rebased commit which is why it didn't auto close).

PR #389, opened 2018-02-04
Add a static property to AmqpTcpEndpoint to specify the default SSL protocol versions and use it on ConnectionFactory.SetUri

paulomorgado:
## Proposed Changes

When creating a connection from an `Uri` there's no way to specify the accepted TLS versions.

Having a property  on [`AmqpTcpEndpoint`](blob/master/projects/client/RabbitMQ.Client/src/client/api/AmqpTcpEndpoint.cs) that is used by [`ConnectionFactory.SetUri`](blob/master/projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs) would allow users of the library to control the default SSL protocol versions used.

## Types of Changes

## Checklist

## Further Comments

considered, etc.

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/api/AmqpTcpEndpoint.cs
+++ b/projects/client/RabbitMQ.Client/src/client/api/AmqpTcpEndpoint.cs
@@ -40,7 +40,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Security.Authentication;
 using System.Text.RegularExpressions;
 
 namespace RabbitMQ.Client
@@ -62,7 +61,6 @@ namespace RabbitMQ.Client
         /// Default Amqp ssl port.
         /// </summary>
         public const int DefaultAmqpSslPort = 5671;
-
 
         /// <summary>
         /// Indicates that the default port for the protocol should be used.
@@ -141,11 +139,6 @@ namespace RabbitMQ.Client
         }
 
         /// <summary>
-        /// The default Amqp SSL protocols.
-        /// </summary>
-        public static SslProtocols DefaultAmqpSslProtocols { get; set; } = SslProtocols.Tls | SslProtocols.Tls11 | SslProtocols.Tls12;
-
-        /// <summary>
         /// Retrieve or set the hostname of this <see cref="AmqpTcpEndpoint"/>.
         /// </summary>
         public string HostName { get; set; }
@@ -181,7 +174,7 @@ namespace RabbitMQ.Client
         /// <summary>
         /// Used to force the address family of the endpoint
         /// </summary>
-        public System.Net.Sockets.AddressFamily AddressFamily { get; set; }
+        public System.Net.Sockets.AddressFamily AddressFamily {get; set;}
 
         /// <summary>
         /// Retrieve the SSL options for this AmqpTcpEndpoint. If not set, null is returned.
@@ -235,7 +228,7 @@ namespace RabbitMQ.Client
         /// </remarks>
         public static AmqpTcpEndpoint[] ParseMultiple(string addresses)
         {
-            string[] partsArr = addresses.Split(new[] { ',' });
+            string[] partsArr = addresses.Split(new[] {','});
             var results = new List<AmqpTcpEndpoint>();
             foreach (string partRaw in partsArr)
             {
--- a/projects/client/RabbitMQ.Client/src/client/api/AmqpTcpEndpoint.cs
+++ b/projects/client/RabbitMQ.Client/src/client/api/AmqpTcpEndpoint.cs
@@ -40,6 +40,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Security.Authentication;
 using System.Text.RegularExpressions;
 
 namespace RabbitMQ.Client
@@ -61,6 +62,7 @@ namespace RabbitMQ.Client
         /// Default Amqp ssl port.
         /// </summary>
         public const int DefaultAmqpSslPort = 5671;
+
 
         /// <summary>
         /// Indicates that the default port for the protocol should be used.
@@ -139,6 +141,11 @@ namespace RabbitMQ.Client
         }
 
         /// <summary>
+        /// The default Amqp SSL protocols.
+        /// </summary>
+        public static SslProtocols DefaultAmqpSslProtocols { get; set; } = SslProtocols.Tls | SslProtocols.Tls11 | SslProtocols.Tls12;
+
+        /// <summary>
         /// Retrieve or set the hostname of this <see cref="AmqpTcpEndpoint"/>.
         /// </summary>
         public string HostName { get; set; }
@@ -174,7 +181,7 @@ namespace RabbitMQ.Client
         /// <summary>
         /// Used to force the address family of the endpoint
         /// </summary>
-        public System.Net.Sockets.AddressFamily AddressFamily {get; set;}
+        public System.Net.Sockets.AddressFamily AddressFamily { get; set; }
 
         /// <summary>
         /// Retrieve the SSL options for this AmqpTcpEndpoint. If not set, null is returned.
@@ -228,7 +235,7 @@ namespace RabbitMQ.Client
         /// </remarks>
         public static AmqpTcpEndpoint[] ParseMultiple(string addresses)
         {
-            string[] partsArr = addresses.Split(new[] {','});
+            string[] partsArr = addresses.Split(new[] { ',' });
             var results = new List<AmqpTcpEndpoint>();
             foreach (string partRaw in partsArr)
             {
--- a/projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs
+++ b/projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs
@@ -45,6 +45,7 @@ using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using System.Linq;
+using System.Security.Authentication;
 
 #if !NETFX_CORE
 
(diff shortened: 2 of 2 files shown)


projects/client/RabbitMQ.Client/src/client/api/AmqpTcpEndpoint.cs
```diff
@@ -138,6 +140,11 @@ public AmqpTcpEndpoint CloneWithHostname(string hostname)
             return new AmqpTcpEndpoint(hostname, _port, Ssl);
         }
 
+        /// <summary>
+        /// The default Amqp SSL protocols.
+        /// </summary>
+        public static SslProtocols DefaultAmqpSslProtocols { get; set; } = SslProtocols.Tls | SslProtocols.Tls11 | SslProtocols.Tls12;
```
michaelklishin: Does it make sense for this field to be an `AmqpTcpEndpoint` field and not, say, `ConnectionFactory`?

projects/client/RabbitMQ.Client/src/client/api/AmqpTcpEndpoint.cs
```diff
@@ -138,6 +140,11 @@ public AmqpTcpEndpoint CloneWithHostname(string hostname)
             return new AmqpTcpEndpoint(hostname, _port, Ssl);
         }
 
+        /// <summary>
+        /// The default Amqp SSL protocols.
```
michaelklishin: How about "TLS protocol versions that will be used by default"?

projects/client/RabbitMQ.Client/src/client/api/AmqpTcpEndpoint.cs
kjnilsson: I agree it would make sense to keep global config in the connection factory.

paulomorgado: That was my first thought. But than I saw a `DefaultAmqpSslPort` in `AmqpTcpEndpoint`...

For me, it makes a lot more sense to honor the SSL options already part of `ConnectionFactory` when using an URI than having `AmqpTcpEndpoint` run free with a mind if its own.

Whst do you think?

projects/client/RabbitMQ.Client/src/client/api/AmqpTcpEndpoint.cs
paulomorgado: LGTM

projects/client/RabbitMQ.Client/src/client/api/AmqpTcpEndpoint.cs
michaelklishin: I'd move both this field and `DefaultAmqpSslPort ` to connection factory and bump the version to `5.2.0`. @kjnilsson WDYT?

bording: Moving `DefaultAmqpSslPort` to a different class is a breaking change, so a new major is needed if you do that.

michaelklishin: It's a rarely used directly field. It would be a true facepalm to release a 6.0 because of that.

michaelklishin: Well, this new field can go into `ConnectionFactory` for 5.x and we can move the port one in master only.

kjnilsson: We can leave DefaultAmqpSslPort, mark it as obsolete and move it to ConnectionFactory.

We are gradually accumulating breaking and proposed breaking changes to warrant considering a major release at some point in the not so distant future.

paulomorgado: That would keep it binarily compatible while directing development to the new way.

michaelklishin: Sounds like we have a consensus then? Note that the port property move ideally should be in a separate PR.

paulomorgado: Property or constant filed?

And the `DefaultAmqpSslProtocols` would still be a static property, right? And with the good defaults, right?

paulomorgado: So, what's the verdict?

michaelklishin: @paulomorgado in this PR, let's move the field to `ConnectionFactory`. In a separate one for master, move the port one with a backwards compatible reference. Both @kjnilsson and I are leaning towards using properties for no particular reason, merely as a matter of taste.

Thank you!

paulomorgado: It's not a matter of taste. It's semantic.

I've split SSL protocols settings into static and instance properties and named them to convey the meaning that they will only be used with URIs.

As for the default port, that one can be overridden in the URI and, isn't 5671 the default as 80 is for HTTTP and 443 for HTTPS? If it is, it's pretty fine where it is. `System.Uri` has default ports per syntax.

projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs
```diff
@@ -141,6 +142,18 @@ public class ConnectionFactory : ConnectionFactoryBase, IAsyncConnectionFactory
         /// <remarks> PLEASE KEEP THIS MATCHING THE DOC ABOVE.</remarks>
         public const string DefaultVHost = "/";
 
+        /// <summary>
+        /// The default AMQP URI SSL protocols.
+        /// </summary>
+        public static SslProtocols DefaultAmqpUriSslProtocols { get; set; } =
```
bording: Why does this have a setter? If that's the default, shouldn't it be get-only?

bording:
Given that one of the new properties is `DefaultAmqpUriSslProtocols`, that should be used regardless of how I set up the properties of the connection factory. There doesn't appear to be anything that sets this value on the `Ssl` property outside of calling `SetUri`.

projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs
paulomorgado: In case the user want's to change the default globally.

michaelklishin: For this particular property it sounds OK to me to allow for overriding.

kjnilsson:
Should `DefaultAmqpUriSslProtocols ` be consulted here: https://github.com/paulomorgado/rabbitmq-dotnet-client/blob/ca989563b55abaf4835dfe9c4c575bc19b74f5f6/projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs#L253

paulomorgado:
> Should `DefaultAmqpUriSslProtocols` be consulted here: https://github.com/paulomorgado/rabbitmq-dotnet-client/blob/ca989563b55abaf4835dfe9c4c575bc19b74f5f6/projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs#L253

Makes sense. @kjnilsson.

projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs
paulomorgado: `AmqpUriSslProtocols` would actually be the one overriding the default.

paulomorgado:
I've been away for a few days. Any update on this?

projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs
michaelklishin: Then I agree with @bording: the setter can be dropped. We don't need two ways to override the default.

michaelklishin:
@kjnilsson I'd like to drop one setter and modify a comment. Happy to do that on my own. Let me know   if you have any objections.

projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs
paulomorgado: The intention was to have a wider default that would apply to all connection factories if not changed and a setting for each connection factory.

michaelklishin: Fair enough, the setter is something we can accept.

paulomorgado:
The build is ending in error with 

    Couldn't find 'project.json' in '.\projects\client\Unit'

But the PR only touches one .cs file. Why is this failing?

paulomorgado:
Maybe we shouldn't provide a default out of the box.

[Transport Layer Security (TLS) best practices with the .NET Framework](https://docs.microsoft.com/en-us/dotnet/framework/network-programming/tls)

michaelklishin:
Not providing a default is usually a very bad idea for libraries. It's not like if you don't provide a default, your users will make an informed decisions or will read the docs. No, some will hack it until it stops complaining and move on.

michaelklishin:
The guide says

> To ensure .NET Framework applications remain secure, the TLS version should not be hardcoded

we don't hardcode it and won't be hardcording it with this PR. And also

> We recommend that you don't use Default; setting SslProtocols.Default
> forces the use of SSL 3.0 /TLS 1.0 and prevent TLS 1.2

I don't think we use `SslProtocols.Default` and RabbitMQ won't accept SSLv3 connections anyway.

So I'm not sure what the argument (and suggested alternative) for "not having a library default" is. Please elaborate?

paulomorgado:
We are creating a default for URIs. And in the PR I've set the default for `Tls | Tls11 | Tls12`. The default should be `None`, I think.

michaelklishin:
Why should it be `None`? What would happen if the user never overrides it?

paulomorgado:
It's my understanding of the recommendation. Am I wrong?

What should the default be, than?

michaelklishin:
Merging per verbal approval from @kjnilsson. The defaults seem OK for the time being, we may revisit the decision to include TLSv1.0 into the default list within a year.

paulomorgado:
> Merging per verbal approval from @kjnilsson. The defaults seem OK for the time being, we may revisit the decision to include TLSv1.0 into the default list within a year.

What do you mean, @michaelklishin? The default is `SslProtocols.Tls | SslProtocols.Tls11 | SslProtocols.Tls12`.

michaelklishin:
@paulomorgado TLSv1.0 is no longer considered to be suitable for some environments but dropping it would be a breaking change.

paulomorgado:
When will this be available on NuGet?

michaelklishin:
@paulomorgado we make no promises about GA releases. A new 5.1.x milestone can be released shortly.

PR #399, opened 2018-04-05
Unspecified socket family matches any single IPAddress

Pzixel:
## Proposed Changes

Today I have spend around 10 hours trying to investigate why client is so upset and it spam me with exceptions. After some investigation I found that there is a problem that if `IPAddressType` is unspecified then `TcpClientAdapter` just fails to do anything.

This change allows to resolve single IPAddress as matching one if Socket type is `Unspecified` , Otherwise, the old behaviour sustains. 

## Types of Changes

## Checklist

## Further Comments

I moved code block that makes decision about matching address into another class to make testing easier.

---

My real issue was that EasyNetQ is always creating sockets with `Unspesified` address family and it cannot be configured outside. I'm going to propose some fixes to them to make their code correctly determine address family, but I think that unspecified with single alternative is a valid change here too.

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/impl/TcpClientAdapterHelper.cs
+++ b/projects/client/RabbitMQ.Client/src/client/impl/TcpClientAdapterHelper.cs
@@ -1,0 +1,20 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Net.Sockets;
+
+namespace RabbitMQ.Client
+{
+    public static class TcpClientAdapterHelper
+    {
+        public static IPAddress GetMatchingHost(IReadOnlyCollection<IPAddress> addresses, AddressFamily addressFamily)
+        {
+            var ep = addresses.FirstOrDefault(a => a.AddressFamily == addressFamily);
+            if (ep == null && addresses.Count == 1 && addressFamily == AddressFamily.Unspecified)
+            {
+                return addresses.Single();
+            }
+            return ep;
+        }
+    }
+}
--- a/projects/client/RabbitMQ.Client/src/client/impl/TcpClientAdapter.cs
+++ b/projects/client/RabbitMQ.Client/src/client/impl/TcpClientAdapter.cs
@@ -26,8 +26,8 @@ namespace RabbitMQ.Client
         {
             AssertSocket();
             var adds = await Dns.GetHostAddressesAsync(host).ConfigureAwait(false);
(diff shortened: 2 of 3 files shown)


michaelklishin:
Thank you, the change makes sense at first glance, we will take a closer look soon.

Pzixel:
Some clarification: the problem is that `CreateConnection(String clientProvidedName)` loads sockets via `LocalEndpoints()` that returns `new List(this.EndPoint)`, where `this.EndPoint` is 

```cs
        public AmqpTcpEndpoint Endpoint
        {
            get { return new AmqpTcpEndpoint(HostName, Port, Ssl); }
            set
            {
                Port = value.Port;
                HostName = value.HostName;
                Ssl = value.Ssl;
            }
        }
```

as result this endpoint is always `Unspecified` (because `AddressFamily` is never set where `Unspecified is default for this enum`) thus adapter fails any operation.

michaelklishin:
Thank you, we will get to review it soon.

kjnilsson:
I think we can generalise this to always return addresses.FirstOrDefault() when ep == null. Any reason not to?

projects/client/RabbitMQ.Client/src/client/impl/TcpClientAdapterHelper.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Net.Sockets;
+
+namespace RabbitMQ.Client
+{
+    public static class TcpClientAdapterHelper
+    {
+        public static IPAddress GetMatchingHost(IReadOnlyCollection<IPAddress> addresses, AddressFamily addressFamily)
+        {
+            var ep = addresses.FirstOrDefault(a => a.AddressFamily == addressFamily);
+            if (ep == null && addresses.Count == 1 && addressFamily == AddressFamily.Unspecified)
```
kjnilsson: Let's make this `if (ep == null) return addresses.FirstOrDefault()` - I'm not convinced we need all the other checks - might as well try to return something if we got it.

Pzixel: @kjnilsson I agree that check for `address.Count` may be too restrictive, but I think we only should perform it if socket is `Unspecified` because otherwise we may return `ipv4` address when socket explicitely says it's `ipv6` or something

kjnilsson: ok makes sense

PR #408, opened 2018-04-10
Removed allot of casting from the code.

YulerB:
## Proposed Changes

Removing casting using generics, and correct typing.

## Types of Changes

- [] Cosmetics (white space, appearance)

## Checklist

- [] I have added necessary documentation (if appropriate)
- [] Any dependent changes have been merged and published in related repositories

### DIFF
--- a/projects/client/Apigen/src/apigen/Apigen.cs
+++ b/projects/client/Apigen/src/apigen/Apigen.cs
@@ -602,7 +602,7 @@ namespace RabbitMQ.Client.Apigen
 
             EmitLine("");
             EmitLine("    public " + MangleClass(c.Name) + "Properties() {}");
-            EmitLine("    public override int ProtocolClassId { get { return " + c.Index + "; } }");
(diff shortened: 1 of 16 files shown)


paulomorgado:
`ushort` is not [CLS compliant](https://docs.microsoft.com/en-us/dotnet/standard/language-independence-and-language-independent-components#Types). The code should move away from `ushort`  and not towards it.

bording:
> ushort is not CLS compliant. The code should move away from ushort and not towards it.

@paulomorgado Given that the underlying AMQP protocol relies on unsigned values, I don't think it's possible or desirable to do that.

paulomorgado:
@bording, like many Windows APIs. And, yet, you won't find a .NET public API that is not CLS compliant.

kjnilsson:
I merged this with some changes. I don't have a problem with using the most appropriate type of the fields in question. We can have a separate discussion as to what it is we want to achieve or not with CLS compliancy.

Issue #437, opened 2018-08-26
Add ClientProvidedName to connection factories


paulomorgado:
This would be a property om `IConnectionFactory` that is used instead of null when a client name is not provided when invoking `CreateConnection`.

paulomorgado:
According to [semver.og](semver.org) this could be on 5.2.0:

> 2. MINOR version when you add functionality in a backwards-compatible manner, and

paulomorgado:
OK! I already made the same changes on my wrapper. When this change is released, I'll just delete a few lines of code.

PR #438, opened 2018-08-26
Add ClientProvidedName to connection factories

paulomorgado:
## Proposed Changes

This would be a property om `IConnectionFactory` that is used instead of null when a client name is not provided when invoking `CreateConnection`.

Solves #437 

## Types of Changes

## Checklist

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs
+++ b/projects/client/RabbitMQ.Client/src/client/api/ConnectionFactory.cs
@@ -337,6 +337,11 @@ namespace RabbitMQ.Client
         }
 
         /// <summary>
+        /// Default client provided name to be used for connections.
+        /// </summary>
+        public string ClientProvidedName { get; set; }
+
+        /// <summary>
         /// Given a list of mechanism names supported by the server, select a preferred mechanism,
         ///  or null if we have none in common.
         /// </summary>
@@ -364,7 +369,7 @@ namespace RabbitMQ.Client
         /// </exception>
         public virtual IConnection CreateConnection()
         {
-            return CreateConnection(this.EndpointResolverFactory(LocalEndpoints()), null);
+            return CreateConnection(this.EndpointResolverFactory(LocalEndpoints()), ClientProvidedName);
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


bording:
Adding a member to an interface is a breaking change.

paulomorgado:
@bording, I don't know what you intend as "breaking change".

According to [semantic versioning](https://semver.org/), this is requires a minor version increment because it adds functionality in a backwards-compatible manner.

According to the [definition of breaking changes for the coding guidelines for the .NET Core Frameweork](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-definitions.md), I don't think it's breaking anything.

Unless I'm missing something.

michaelklishin:
@paulomorgado implementations of an interface that has changed will be affected.

paulomorgado:
@michaelklishin,

Oh! if someone else implemented a connection factory.

Nevertheless, this is a case for when this will be added, not the merit of the change. Right?

michaelklishin:
@paulomorgado sure, we can definitely consider it for a 6.0. Thank you.

kjnilsson:
Pre release published: https://www.nuget.org/packages/RabbitMQ.Client/6.0.0-pre1

paulomorgado:
According to [semver.og](semver.org) this could be on 5.2.0:

> 2. MINOR version when you add functionality in a backwards-compatible manner, and

michaelklishin:
It changes an interface, however, so existing implementations will require modifications (even though I don't expect there to be many of those). The exact interpretation of semver.org is project-specific in practice. We seem to have a discussion about that every few months 🙄, e.g. I posted something in https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/442#issuecomment-422716188

michaelklishin:
They seem to have added an [FAQ](https://semver.org/#what-if-i-inadvertently-alter-the-public-api-in-a-way-that-is-not-compliant-with-the-version-number-change-ie-the-code-incorrectly-introduces-a-major-breaking-change-in-a-patch-release) which boils down to "when in doubt, use your best judgement". For this specific project w.r.t. interface changes that historically has meant "we can only do that in a major releasee" because of how certain projects that build on top are developed.

paulomorgado:
I can't wait for default interface members in C#. 😄

PR #443, opened 2018-09-06
Reduce outbound frame allocations

Pliner:
## Proposed Changes

This PR continues work on allocations(see #442). 
Before PR:
1. `OutboundFrame` and its inheritants allocate  `MemoryStream`, `NetworkBinaryStream` in constructor and call `MemoryStream.ToArray` in `OutboundFrame.WriteTo`.
2. `SocketFrameHandler` allocates `MemoryStream` and calls `MemoryStream.ToArray` in`WriteFrameSet`, `WriteFrame` and `SendHeader`.
After PR:
1. `OutboundFrame` and its inheritants try not to allocate where possible and write directly to  `NetworkBinaryStream`. In other case instead of calling `MemoryStream.ToArray` they try to use internal buffer of  `MemoryStream` (but it is not possible in netstandart1.5, but possible in netstandart2.0).
2. `SocketFrameHandler` still allocates `MemoryStream`, but instead of calling `MemoryStream.ToArray` it tries to use internal buffer of  `MemoryStream` (but it is not possible in netstandart1.5, but possible in netstandart2.0)

## Types of Changes

## Checklist

## Further Comments

`Producer.cs`
```
    {
        {
            {
                {
                    {
                    }
                }
            }   

            Console.ReadKey();
        }
    }
```

Before PR:
image
[Before.zip](https://github.com/rabbitmq/rabbitmq-dotnet-client/files/2356982/Before.zip) - dotMemory Workspace

After PR:
image
[After.zip](https://github.com/rabbitmq/rabbitmq-dotnet-client/files/2356980/After.zip) - dotMemory Workspace

As I see now, the next step can be pooling of the `MemoryStream`(https://github.com/Microsoft/Microsoft.IO.RecyclableMemoryStream). What do you think about it @michaelklishin @kjnilsson?

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/impl/ExtensionMethods.cs
+++ b/projects/client/RabbitMQ.Client/src/client/impl/ExtensionMethods.cs
@@ -65,7 +65,7 @@ namespace RabbitMQ.Client.Impl
             return list.ElementAt<T>(hashCode % n);
         }
 
-        public static ArraySegment<byte> GetBufferSegment(this MemoryStream ms)
+        internal static ArraySegment<byte> GetBufferSegment(this MemoryStream ms)
         {
 #if CORECLR15
             var payload = ms.ToArray();
--- a/projects/client/RabbitMQ.Client/src/client/impl/ExtensionMethods.cs
+++ b/projects/client/RabbitMQ.Client/src/client/impl/ExtensionMethods.cs
@@ -40,6 +40,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 
 namespace RabbitMQ.Client.Impl
@@ -62,6 +63,17 @@ namespace RabbitMQ.Client.Impl
 
             var hashCode = Math.Abs(Guid.NewGuid().GetHashCode());
             return list.ElementAt<T>(hashCode % n);
+        }
+
+        public static ArraySegment<byte> GetBufferSegment(this MemoryStream ms)
+        {
+#if CORECLR15
+            var payload = ms.ToArray();
+            return new ArraySegment<byte>(payload, 0, payload.Length);
+#else
+            var buffer = ms.GetBuffer();
+            return new ArraySegment<byte>(buffer, 0, (int)ms.Position);
+#endif
         }
     }
 }
--- a/projects/client/RabbitMQ.Client/src/client/impl/SocketFrameHandler.cs
+++ b/projects/client/RabbitMQ.Client/src/client/impl/SocketFrameHandler.cs
@@ -222,7 +222,7 @@ namespace RabbitMQ.Client.Impl
                 nbw.Write((byte)Endpoint.Protocol.MajorVersion);
                 nbw.Write((byte)Endpoint.Protocol.MinorVersion);
             }
-            Write(ms.ToArray());
+            Write(ms.GetBufferSegment());
         }
 
         public void WriteFrame(OutboundFrame frame)
@@ -231,7 +231,7 @@ namespace RabbitMQ.Client.Impl
             var nbw = new NetworkBinaryWriter(ms);
             frame.WriteTo(nbw);
             m_socket.Client.Poll(m_writeableStateTimeout, SelectMode.SelectWrite);
-            Write(ms.ToArray());
+            Write(ms.GetBufferSegment());
         }
 
         public void WriteFrameSet(IList<OutboundFrame> frames)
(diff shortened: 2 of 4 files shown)


projects/client/RabbitMQ.Client/src/client/impl/SocketFrameHandler.cs
```diff
@@ -240,21 +240,21 @@ public void WriteFrameSet(IList<OutboundFrame> frames)
             var nbw = new NetworkBinaryWriter(ms);
```
Pliner: Could you tell why do we not write frames to `m_writer` directly instead of buffering them first? @michaelklishin @kjnilsson

projects/client/RabbitMQ.Client/src/client/impl/SocketFrameHandler.cs
```diff
@@ -231,7 +231,7 @@ public void WriteFrame(OutboundFrame frame)
...
             frame.WriteTo(nbw);
             m_socket.Client.Poll(m_writeableStateTimeout, SelectMode.SelectWrite);
-            Write(ms.ToArray());
+            Write(ms.GetBufferSegment());
```
Pliner: Could you tell why do we not write frames to `m_writer` directly instead of buffering them first?

Pliner: @michaelklishin @kjnilsson

projects/client/RabbitMQ.Client/src/client/impl/ExtensionMethods.cs
```diff
@@ -63,5 +64,16 @@ public static T RandomItem<T>(this IList<T> list)
...
             return list.ElementAt<T>(hashCode % n);
         }
+
+        internal static ArraySegment<byte> GetBufferSegment(this MemoryStream ms)
+        {
+#if CORECLR15
+            var payload = ms.ToArray();
+            return new ArraySegment<byte>(payload, 0, payload.Length);
+#else
+            var buffer = ms.GetBuffer();
```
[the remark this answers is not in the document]
Pliner: https://docs.microsoft.com/en-us/dotnet/api/system.io.memorystream.getbuffer?redirectedfrom=MSDN&view=netframework-4.7.2#System_IO_MemoryStream_GetBuffer

It seems yes.

projects/client/RabbitMQ.Client/src/client/impl/SocketFrameHandler.cs
Pliner: Thanks 👍 

So, the only way to reduce allocations is to pool `MemoryStream`.

kjnilsson:
This PR looks good to me. Nice improvement again. :)

About MemoryStream pooling I am less sure. Don't really want to take on a dependency as only just managed to get the client dependency less again.

Pliner:
>About MemoryStream pooling I am less sure. Don't really want to take on a dependency as only just managed to get the client dependency-free again.

I can try to add simple pooling of `MemoryStream` to have no dependency. Is it worth to do?

kjnilsson:
It's very hard to say if it is worth doing without first doing it :)

Anyone else got a view on MemoryStream pooling? @michaelklishin @bording @danielmarbach

danielmarbach:
I think @Scooletz would be the right person to answer that question

Scooletz:
IMO: If you want to stick to `MemoryStream` and do not take any dependencies, especially on `System.Memory`, a simple `ConcurrentQueue<MemoryStream>` or `ConcurrentStack<MemoryStream>` could do (queue will not allocate and is a bit more concurrent-friendly but will use FIFO which might trash cache a little bit, stack will allocate but ensures reuse of the recent one). Kestrel's pool uses the queue approach to minimize the friction.

Are streams more less equal in length? If not, pooling might over-allocate (using a stream with length 10000 to write 1byte). Maybe having a fixed-size pool would be a good compromise.

michaelklishin:
Pooling sounds trivial but from my experience is hard to get right. With such impressive gains like in this PR and #442, do we really want to incur a high complexity cost?

Scooletz:
Just to clarify: I'm not advocating for pooling `MemoryStream`, but rather trying to share some findings from my exp. Having a bounded pool (an array, or a `ConcurrentQueue` with a counter), is a not hard thing to implement that can address some part of the concerns without introducing a need to reclaim the memory.

michaelklishin:
@Pliner thanks as usual for your solid contributions to this client!

PR #483, opened 2018-09-22
Make IAsyncConnectionFactory public

### LINKED ISSUE #482: Consider making IAsyncConnectionFactory public
While the API for creating asynchronous connections is not the best approach, it is what we have and it's impossible to rely on interfaces only to create asynchronous connections.

This would be immediately solved if `IAsyncConnectionFactory` was public.

paulomorgado:
## Proposed Changes

While the API for creating asynchronous connections is not the best approach, it is what we have and it's impossible to rely on interfaces only to create asynchronous connections.

This would be immediately solved if `IAsyncConnectionFactory` was public.

Fixes #482 

## Types of Changes

## Checklist

## Further Comments

I don't think this would be a breaking change for the next minor version.

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/api/IAsyncConnectionFactory.cs
+++ b/projects/client/RabbitMQ.Client/src/client/api/IAsyncConnectionFactory.cs
@@ -1,7 +1,55 @@
+// This source code is dual-licensed under the Apache License, version
+// 2.0, and the Mozilla Public License, version 1.1.
+//
+// The APL v2.0:
+//
+//---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


kjnilsson:
This looks fine with me - any other comments on this?

One think I would say is that we could improve the async api for 6.0 quite happily.

paulomorgado:
We should, definitely, improve async on 6.0.

Will this be available in 5.2?

michaelklishin:
It's not clear if there is going to be a 5.2.0 and when. Too many changes in master are breaking and all focus seems to be on 6.0.

PR #490, opened 2018-10-11
make implementations internal and sealed

### LINKED ISSUE #466: Make implementation internal
As an exercise, I made almost every implementation detail internal: https://github.com/paulomorgado/rabbitmq-dotnet-client/tree/internal

There's surely room for improvement. But I think it's a good start.

paulomorgado:
## Proposed Changes

The code base is filled with unnecessary public types or members and unnecessary virtual members.

This PR locks down everything possible to lock down.

Fixes #466 

## Types of Changes

## Checklist


paulomorgado:
I think I got it solved.

I'm still having some trouble setting up a test environment. I'm running the server from a container but the tests are missing some rabbitmqctl.bat file. Is there a way to get it without having to install a server locally.

michaelklishin:
@paulomorgado RabbitMQ CLI tools ship with the broker.

paulomorgado:
@michaelklishin, can you tell me why the tests are failing now?

michaelklishin:
The only error I could spot was

```
CSC : error CS1617: Invalid option '7.3' for /langversion. Use '/langversion:?' to list supported values. [/tmp/build/b95f7b06/rabbitmq_dotnet_client/projects/client/Unit/Unit.csproj]
    0 Warning(s)rifying Checksum
    1 Error(s)Download complete
```

paulomorgado:
You need to upgrade the build tools on that build server.

michaelklishin:
We will. Note that we are happy to merge your contributions if they pass on our developer machines and update Concourse environment(s) later.

bording:
@kjnilsson If you're actually running against the latest .NET Core SDK, then I would expect it to understand the 7.3 langversion. It seems like you've got something wrong in your CI still.

kjnilsson:
```root@29fc37abe41d:/# dotnet --version
2.1.4```

kjnilsson:
It builds ok if we remove LangVersion - please update the PR without a specific lang version (Why only add it for Unit anyway?). Cheers.

bording:
> It doesn't build ok at all with no langversion but 7.2 seems ok

If you don't specify a language version, it defaults to the latest major version of the C# language, which right now means 7.0. So if you're using any language features introduced in 7.1, 7.2, or 7.3, you'll get errors.

BTW, the `LangVersion` property also accepts "latest" as a value, which would let the project pick up the newest language features without needing to update the project value.

paulomorgado:
@bording, `latest` in my machine is `7.3`. That's why it's a good practice to make the version explicit.

Meanwhile, I was able to make it work with `7.0` (`default`).

PR #507, opened 2018-12-26
Added routingKey fallback to string.Empty when null is passed as its …

starkmsu:
…value for BasicPublish

## Proposed Changes

Since routing key is ignored by documentation for fanout exchange, passing Null as its value should be ok when calling BasicPublish.
But such case results in exception.
That's why this change is proposed.

## Types of Changes

- [x] Bugfix (non-breaking change which fixes issue #NNNN)

## Checklist

## Further Comments

considered, etc.

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/impl/AutorecoveringModel.cs
+++ b/projects/client/RabbitMQ.Client/src/client/impl/AutorecoveringModel.cs
@@ -575,8 +575,12 @@ namespace RabbitMQ.Client.Impl
             IBasicProperties basicProperties,
             byte[] body)
         {
-            m_delegate._Private_BasicPublish(exchange, routingKey, mandatory,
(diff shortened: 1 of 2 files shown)


michaelklishin:
I’m more inclined throw an illegal argument exception if the argument is a `null` than fall back to an empty string. @kjnilsson @bording thoughts?

kjnilsson:
I agree with @michaelklishin - it is better to throw the exception when the routing key is null as for other scenarios a null value may be indicative of a bug somewhere in the user code. It is easy enough to create your own domain-specific extension methods that encode your invariants.

michaelklishin:
@starkmsu please submit a PR that rejects `null` values instead. Note that we can only include it into the 6.0 release since it’s a breaking change in a very widely used public API method.

PR #579, opened 2019-03-19
Byte conversion support

sergeyshaykhullin:
## Proposed Changes

I tried to add byte conversion support for arguments. but this not a final version. Solutions are:
1. Add new byte case and write as byte
2. Add new byte case and write as int/short
3. Do nothing because has int-workaround

Motivation:
RabbitMQ has `x-max-priority` for queues, `x-priority` for consumers and priority in `IBasicProperties` for producers, but for them is ok. Commonly i used them to balance messages.
Priorities is in 1...255 range and perfectly suites to byte, but when i tried wrap arguments with byte extensions i got a `WireFormattingException`(`Value cannot appear as table value`) and nothing more. Boxing int for priority works fine.

Usage: https://gist.github.com/sergeyshaykhullin/bca8964c0ebbd5a8737a559bcb0287e3

If English isn't your first language, don't worry about it and try to
communicate the problem you are trying to solve to the best of your abilities.
As long as we can understand the intent, it's all good.

## Types of Changes
- [x] (Possible) Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

## Further Comments
There are possible breaking changes, f.e. new version writes 'B', but old version doesnt know about it and throws `SyntaxError`

Docs https://www.rabbitmq.com/priority.html has
For 'x-max-priority' - `This argument should be a positive integer between 1 and 255`
For `x-priority` - `The message priority field is defined as an unsigned byte, so in practice priorities should be between 0 and 255.`
IBasicProperties has `byte` priority
So maybe it can help someone in future

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/impl/WireFormatting.cs
+++ b/projects/client/RabbitMQ.Client/src/client/impl/WireFormatting.cs
@@ -57,11 +57,17 @@ namespace RabbitMQ.Client.Impl
                 throw new SyntaxError("Unrepresentable AMQP decimal table field: " +
                                       "scale=" + scale);
             }
-            return new decimal((int)(unsignedMantissa & 0x7FFFFFFF),
-                0,
-                0,
-                ((unsignedMantissa & 0x80000000) == 0) ? false : true,
-                scale);
+
+            return new decimal(
+                // The low 32 bits of a 96-bit integer
+                lo: (int)(unsignedMantissa & 0x7FFFFFFF),
+                // The middle 32 bits of a 96-bit integer.
+                mid: 0,
+                // The high 32 bits of a 96-bit integer.
+                hi: 0,
+                isNegative: (unsignedMantissa & 0x80000000) != 0,
+                // A power of 10 ranging from 0 to 28.
+                scale: scale);
         }
 
         public static void DecimalToAmqp(decimal value, out byte scale, out int mantissa)
@@ -91,7 +97,7 @@ namespace RabbitMQ.Client.Impl
         {
             IList array = new List<object>();
             long arrayLength = reader.ReadUInt32();
-            Stream backingStream = reader.BaseStream;
+            var backingStream = reader.BaseStream;
             long startPosition = backingStream.Position;
             while ((backingStream.Position - startPosition) < arrayLength)
             {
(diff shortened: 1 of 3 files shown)


projects/client/RabbitMQ.Client/src/client/impl/WireFormatting.cs
```diff
@@ -129,10 +135,12 @@ public static object ReadFieldValue(NetworkBinaryReader reader)
                 case 'F':
                     value = ReadTable(reader);
                     break;
-
                 case 'A':
                     value = ReadArray(reader);
                     break;
+                case 'B':
```
sergeyshaykhullin: Reading

projects/client/RabbitMQ.Client/src/client/impl/WireFormatting.cs
```diff
@@ -264,93 +272,82 @@ public static void WriteArray(NetworkBinaryWriter writer, IList val)
...
+                    break;
+                case IDictionary val:
+                    WriteOctet(writer, (byte)'F');
+                    WriteTable(writer, val);
+                    break;
+                case IList val:
+                    WriteOctet(writer, (byte)'A');
+                    WriteArray(writer, val);
+                    break;
+                case byte val:
```
sergeyshaykhullin: Writing

projects/client/RabbitMQ.Client/src/client/impl/WireFormatting.cs
```diff
@@ -264,93 +272,82 @@ public static void WriteArray(NetworkBinaryWriter writer, IList val)
...
+                    WriteOctet(writer, (byte)'t');
+                    WriteOctet(writer, (byte)(val ? 1 : 0));
+                    break;
+                case BinaryTableValue val:
+                    WriteOctet(writer, (byte)'x');
+                    WriteLongstr(writer, val.Bytes);
+                    break;
+                default:
+                    throw new WireFormattingException(
+                        $"Value of type '{value.GetType().Name}' cannot appear as table value",
```
sergeyshaykhullin: Some exception details

michaelklishin:
Thank you for your ongoing pull requests!

It's not clear what's in scope for this PR. Adding some tests would help.

Please make sure you have read the [AMQP 0-9-1 Errata document](https://www.rabbitmq.com/amqp-0-9-1-errata.html).

sergeyshaykhullin:
Yes, i got 'B' for byte from this document

michaelklishin:
Thank you. I will cherry-pick to `5.x.`.

PR #635, opened 2019-03-21
Add unsigned 32 bit integer support

sergeyshaykhullin:
## Proposed Changes
I tried to use for queues `x-max-length` and `x-max-length-bytes` arguments unsigned integers, as mentioned here https://www.rabbitmq.com/maxlength.html
1. Maximum number of messages can be set by supplying the x-max-length queue declaration argument with a *non-negative integer* value.
2. Maximum length in bytes can be set by supplying the x-max-length-bytes queue declaration argument with a *non-negative integer* value.

And got the same `WireFormattingException(Value cannot appear as table value)`.

From https://www.rabbitmq.com/amqp-0-9-1-errata.html 'Unsigned 32-bit' is 'i', i wanted to add ulong support, but if i'm not mistaken rabbitmq does not support it?

## Types of Changes
- [x] (Possible) Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

## Further Comments

There is still possible breaking change, when new-version producer send u32, but old-version consumer don't know about it, but guess all client code now uses signed integers for args

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/impl/WireFormatting.cs
+++ b/projects/client/RabbitMQ.Client/src/client/impl/WireFormatting.cs
@@ -126,6 +126,9 @@ namespace RabbitMQ.Client.Impl
                 case 'I':
                     value = reader.ReadInt32();
                     break;
+                case 'i':
+                    value = reader.ReadUInt32();
+                    break;
                 case 'D':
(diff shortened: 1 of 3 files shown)


michaelklishin:
Thank you!

michaelklishin:
Backported to `5.x`.

PR #636, opened 2019-04-03
Improved Heartbeat Write Timer Handling

ash-ricado:
- Resolves an issue with the Heartbeat Write Timer repeatedly firing and blocking on the sychronization lock. The timer will now fire once and be restarted (if not disposed) at the end of method processing.

## Proposed Changes

Modified synchronization and parameters of the Write Heartbeat Timer to ensure it can only fire once. It will then be started again at the end of the callback.

This resolves an issue that can occur when writing a heartbeat frame to the socket fails (e.g. broken connection) and the Write Heartbeat Timer continued to fire and block on the lockable `_heartbeatWriteLock` object. Over time this would eventually exhaust the ThreadPool if the socket write didn't return.

## Types of Changes

- [x] Bugfix (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/impl/Connection.cs
+++ b/projects/client/RabbitMQ.Client/src/client/impl/Connection.cs
@@ -1015,15 +1015,32 @@ entry.ToString());
                 m_hasDisposedHeartBeatReadTimer = false;
                 m_hasDisposedHeartBeatWriteTimer = false;
 #if NETFX_CORE
-                _heartbeatWriteTimer = new Timer(HeartbeatWriteTimerCallback);
-                _heartbeatReadTimer = new Timer(HeartbeatReadTimerCallback);
-                _heartbeatWriteTimer.Change(200, (int)m_heartbeatTimeSpan.TotalMilliseconds);
-                _heartbeatReadTimer.Change(200, (int)m_heartbeatTimeSpan.TotalMilliseconds);
+                lock(_heartBeatWriteLock)
+                {
+                    _heartbeatWriteTimer = new Timer(HeartbeatWriteTimerCallback);
+                    _heartbeatWriteTimer.Change(200, Timeout.Infinite);
+                }
+
+                lock (_heartBeatReadLock)
+                {
+                    _heartbeatReadTimer = new Timer(HeartbeatReadTimerCallback);
+
+                    _heartbeatReadTimer.Change(300, Timeout.Infinite);
+                }
 #else
-                _heartbeatWriteTimer = new Timer(HeartbeatWriteTimerCallback, null, 200, (int)m_heartbeatTimeSpan.TotalMilliseconds);
-                _heartbeatReadTimer = new Timer(HeartbeatReadTimerCallback, null, 200, (int)m_heartbeatTimeSpan.TotalMilliseconds);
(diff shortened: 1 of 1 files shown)


ash-ricado:
It appears builds are failing due to configuration issues?

michaelklishin:
@ash-ricado you are correct, CI needs propping every so often. We will QA this, don't worry.

michaelklishin:
@ash-ricado any way we can reasonably reliably reproduce this?

projects/client/RabbitMQ.Client/src/client/impl/Connection.cs
```diff
@@ -1015,15 +1015,32 @@ public void MaybeStartHeartbeatTimers()
...
+                {
+                    _heartbeatWriteTimer = new Timer(HeartbeatWriteTimerCallback);
+                    _heartbeatWriteTimer.Change(200, Timeout.Infinite);
+                }
+
+                lock (_heartBeatReadLock)
+                {
+                    _heartbeatReadTimer = new Timer(HeartbeatReadTimerCallback);
+
+                    _heartbeatReadTimer.Change(300, Timeout.Infinite);
```
kjnilsson: Why the change from 200 to 300 here?

michaelklishin:
This looks reasonable. @ash-ricado please undo the interval change (or justify it) and provide some detail on the problem this PR addresses.

projects/client/RabbitMQ.Client/src/client/impl/Connection.cs
ash-ricado: I simply changed this to prevent both the Read and Write Timers firing at the same time. I was considering using Random.Next(100,300) to ensure that with multiple channels, the timers will fire at different times.

Not required to fix the issue this PR deals with. More than happy to remove if desired.

ash-ricado:
@michaelklishin Apologies for the delay in my response.

**Problem solved by this PR**
Resolves an issue that can occur when writing a heartbeat frame to the socket fails (e.g. broken connection) and the Write Heartbeat Timer continues to fire and block on the lockable _heartbeatWriteLock object. Over time this would eventually exhaust the ThreadPool if the socket write didn't return.

Identified by our use of remote nodes that talk to a RabbitMQ Cluster over connections that can sometimes become unstable. Monitoring allowed us to see the ThreadPool increase over a period of an hour until the maximum number of threads was consumed.

Should be able to reliably reproduce by publishing data (12K or larger) to a channel every 5 seconds, then drop the network interface that provides access to the RabbitMQ Server. The Heartbeat Write Timer should continue to fire and wait on the lock.

_NOTE: This does rely on the socket hanging after an unclean network disconnection. If the socket is able to be cleanly shutdown, the attempt to write frames will return immediately._

michaelklishin:
OK, so injecting a latency spike with Toxiproxy or similar would do. Thank you.

projects/client/RabbitMQ.Client/src/client/impl/Connection.cs
michaelklishin: We strongly discourage heartbeat timeouts < 5 seconds so any timer interval < 1s should be reasonable, and anything < 500 ms is optimal IMO.

PR #639, opened 2019-04-16
Don't consider a heartbeat frame write failure to be fatal

### LINKED ISSUE #638: Heartbeat frame write failures should not close connection
There is only so much information in #637 but while inspecting the stack trace, I've noticed that in this client, a [failed heartbeat write closes the connection](https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/5.x/projects/client/RabbitMQ.Client/src/client/impl/Connection.cs#L1131). The code also doesn't account for other activity on the connection.

As a result, on a very busy connection or a connection that managed to trigger a resource alarm or transient flow control for a long enough period, it is feasible to see a heartbeat write failure.

Other clients don't close connections this aggressively. For example, Java client [does not](https://github.com/rabbitmq/rabbitmq-java-client/blob/master/src/main/java/com/rabbitmq/client/impl/HeartbeatSender.java) and, in fact, wouldn't even send heartbeat frames if there was recent activity on a connection.

michaelklishin:
## Proposed Changes

This changes failed heartbeat frame write behavior to not close the connection.
Other clients do not do it [1], the spec doesn't instruct us to do so.
Let the heartbeat reader side detect peer unavailability (as it
currently does, using two missed heartbeats/intervals).

Heartbeat frame write can fail on on a highly congested connection
or a connection that's blocked by a resource alarm. The alarm might
clear soon enough for connection to be considered functional
by the peer.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #638)

## Checklist

## Further Comments

Closes #638.

1. https://github.com/rabbitmq/rabbitmq-java-client/blob/5.6.x-stable/src/main/java/com/rabbitmq/client/impl/HeartbeatSender.java

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/impl/Connection.cs
+++ b/projects/client/RabbitMQ.Client/src/client/impl/Connection.cs
@@ -1146,7 +1146,7 @@ entry.ToString());
             }
             catch (Exception)
             {
-                // ignore, let the read callback detected
+                // ignore, let the read callback detect
                 // peer unavailability. See rabbitmq/rabbitmq-dotnet-client#638 for details.
             }
 
--- a/projects/client/RabbitMQ.Client/src/client/impl/Connection.cs
+++ b/projects/client/RabbitMQ.Client/src/client/impl/Connection.cs
@@ -1132,31 +1132,11 @@ entry.ToString());
                 }
             }
 
-            bool shouldTerminate = false;
-
             try
             {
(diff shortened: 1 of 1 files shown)


michaelklishin:
I suggest we releasee this in `5.1.0` because this is technically a breaking change (even though I cannot think of a case where the previous behavior would be desired).

PR #653, opened 2019-09-05
Updates to AutorecoveringConnection.

michac:
## Proposed Changes

We have a customer using Rabbit that's experiencing network issues that are causing auto-recovery to fail. We think it may be related to the network coming up long enough to reconnect but then failing again during the auto-recovery process. Unfortunately when I started looking into the implementation of auto recovery it was difficult to reconstruct the sequence of events through the delegate callbacks and multiple locking objects.

The primary purpose of my change here is to simplify the auto-recovery class by removing all the locks I could and giving auto recovery a single dedicated background thread run on a blocking collection to synchronize all recovery operations.

## Types of Changes

## Checklist

## Further Comments

Created a clean version off of master as requested by @michaelklishin . He also mentioned breaking API changes. Is that specific to the addition of `ConnectionRecoveryTriggeringCondition` to the factory? I added that early on because the equivalent was in the java client API but I could easily take it out, it's not critical to everything else. Other than that the my intention was not to make API changes.


pivotal-issuemaster:
@michac Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-dotnet-client&pullRequestId=653)!

[Click here](https://cla.pivotal.io/sync/rabbitmq?repositoryId=rabbitmq/rabbitmq-dotnet-client&pullRequestId=653) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@michac Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-dotnet-client&pullRequestId=653)!

michaelklishin:
@michac thank you. We will discuss whether this approach is a simpler one with the team @lukebakken @kjnilsson.

@bording do you have an opinion? ;)

bording:
@michaelklishin I took a quick look, but it appears there are a lot of unrelated/cleanup changes in there in addition to the actual changes, so I'll need to spend some more time to figure it out.

The idea behind the change sounds fine to me, though. I'm all for removing locks where they aren't needed.

projects/client/RabbitMQ.Client/src/client/impl/AutorecoveringConnection.cs
```diff
@@ -63,339 +60,233 @@ internal sealed class AutorecoveringConnection : IConnection
...
+
+        private readonly BlockingCollection<RecoveryCommand> m_recoveryLoopCommandQueue = new BlockingCollection<RecoveryCommand>();
+        private readonly CancellationTokenSource m_recoveryCancellationToken = new CancellationTokenSource();
+        private readonly TaskCompletionSource<int> m_recoveryLoopComplete = new TaskCompletionSource<int>();
 
-        public IDictionary<string, object> ServerProperties
+        /// <summary>
+        /// This is the main loop for the auto-recovery thread.
+        /// </summary>
+        private void MainRecoveryLoop()
```
michac: This is the core change. Instead of kicking off a Task on the shutdown event this version of AutoRecoveringConnection queues a message in the m_recoveryLoopCommandQueue to switch to auto-recovery mode. While it auto-recovery mode it queues a PerformAutomaticRecovery message at the specified NetworkRecoveryInterval. When the recovery succeeds it goes back into the connected state and the thread sits idle. Closing the connection uses the cancellation token to cancel this recovery loop, and waits on the Complete event to be set so we don't Close on a half-finished recovery (which may not be a problem but seems messy).

projects/client/RabbitMQ.Client/src/client/impl/AutorecoveringConnection.cs
```diff
@@ -63,339 +60,233 @@ internal sealed class AutorecoveringConnection : IConnection
...
-                }
-            }
-            set
-            {
-                lock(manuallyClosedLock)
-                {
-                    manuallyClosed = value; }
-                }
-        }
+        public event EventHandler<EventArgs> RecoverySucceeded;
```
michac: I replaced delegates protected by lock statements with events, since these are also thread safe.

projects/client/RabbitMQ.Client/src/client/impl/AutorecoveringConnection.cs
```diff
@@ -63,339 +60,233 @@ internal sealed class AutorecoveringConnection : IConnection
...
-        }
+        public event EventHandler<QueueNameChangedAfterRecoveryEventArgs> QueueNameChangeAfterRecovery;
 
         public string ClientProvidedName { get; private set; }
 
-        public ushort ChannelMax
-        {
-            get { return m_delegate.ChannelMax; }
-        }
+        public ushort ChannelMax => m_delegate.ChannelMax;
```
michac: This is just me letting Resharper do some code reformatting. In retrospect it does add noise to the change and I could have skipped it.

bording: At the very least, you should make these sort of changes in a separate commit.

bording: At the very least, it's a good idea to make these type of changes in a separate commit.

michaelklishin:
Superseded by #656.

PR #654, opened 2019-09-06
Deprecate messaging patterns

kjnilsson:
So they can be removed in 6.0

fixes: #514


projects/client/RabbitMQ.Client/src/client/messagepatterns/ISubscription.cs
```diff
@@ -51,6 +51,7 @@ namespace RabbitMQ.Client.MessagePatterns
     /// for <see cref="Subscription" /> easier.
     ///</para>
     ///</remarks>
+    [Obsolete("Messaging patters are no longer recommended for use and will be removed.")]
```
bording: Typo "patters"

michaelklishin: Corrected, thank you.

PR #657, opened 2019-09-10
Make it possible to use a Docker-provided RabbitMQ node for tests

michac:
## Proposed Changes

I was unable to get the umbrella project running locally to run the unit tests against. However, I was able to make some changes to the Fixtures.cs file to support using a docker instance of rabbit. My test sequence is:

* `docker run -d --hostname rabbitmq01 --name rabbitmq01 -p 15672:15672 -p 5672:5672 rabbitmq:3-management`
* set my `RABBITMQ_RABBITMQCTL_PATH` env variable to `DOCKER:rabbitmq01`
* run unit tests

I figured since I had the changes working I'd submit there here to see if you were interested.

## Types of Changes

## Checklist

## Further Comments

This could be enhanced to actually create an appropriately versioned docker container and tear it down when done so the unit tests just work if you have docker available, but this change doesn't do anything that fancy.

### DIFF
--- a/RUNNING_TESTS.md
+++ b/RUNNING_TESTS.md
@@ -30,16 +30,22 @@ This will complete the code AMQP 0-9-1 protocol code generation and build all pr
 Tests can be run from Visual Studio using [NUnit Test Adapter](https://visualstudiogallery.msdn.microsoft.com/6ab922d0-21c0-4f06-ab5f-4ecd1fe7175d).
 Note that it may take some time for the adapter to discover tests in the assemblies.
 
-The test suite assumes there's a RabbitMQ node [built from source](https://www.rabbitmq.com/build-server.html) running locally
-with all defaults. Team RabbitMQ uses [rabbitmq-public-umbrella](https://github.com/rabbitmq/rabbitmq-public-umbrella) to
-do that:
+The test suite assumes there's a RabbitMQ node running locally with all defaults, and the tests will need to be able to run commands against the [rabbitmqctl](https://www.rabbitmq.com/rabbitmqctl.8.html) tool for that node. Two options to accomplish this are:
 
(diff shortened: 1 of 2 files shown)


michaelklishin:
This looks reasonable to me. Let's see what @kjnilsson @lukebakken say.

kjnilsson:
yep this is handy, thanks!

michaelklishin:
We can add a minimalistic `Dockerfile` that uses https://github.com/docker-library/rabbitmq/ with all defaults. It should be sufficient for this test suite.

michac:
To be clear on the requested changes, I could
a. Document the two manual setup steps I listed above, which might be sufficient if using umbrella continues to be the standard approach and docker is just a backup option.
b. Integrate the run command into the unit tests such that it deploys a container to run the tests then  cleans it up afterwards. This would require a little research on my part for best practices, I've never integrated docker with unit tests like that. This makes more sense to me if you think Docker may become the standard approach in the long term.

michaelklishin:
@michac thank you!

Please use a separate script/target for that. The `./run-tests` we have is used by CI which provisions a node separately. Our team runs RabbitMQ locally and not in a container during client development, too.

michac:
@michaelklishin I wasn't very clear in my comment. What I was asking was whether I should do a OR b. 
I would do option a if I'm the only one who will use it or option b if you think it's worth a little polish and others may use it too.
If I did option b I figured I would still operate based off of the environment variable and just have the test do setup and teardown if docker was specified.

TrymBeast:
Hey guys!
How about having a docker file that builds & runs the tests and a docker compose that builds the docker file and also declares the rabbitmq container as dependency?

michaelklishin:
@TrymBeast that's a reasonable idea, even though I expect many contributors would not mind running the suite in the host (and on different platforms).

michaelklishin:
@michac I would be happy with option A. If this topic seems interesting to you, feel free to investigate. If it takes too long, option A would still be there ;)

michac:
I'm not sure if I'll find time to do anything fancier with Docker, so I just went ahead and updated the documentation in this pull request.

PR #679, opened 2019-12-10
Bump minimum .NET version supported to 4.5.2

kjnilsson:
4.5.1 is long out of support.

Also remove WinRT support.


michaelklishin:
@bording FYI

bording:
This is 100% ok with me!

@kjnilsson NuGet.Build.Tasks.Pack should be able to be removed entirely (along with the `<NuGetBuildTasksPackTargets>junk-value-to-avoid-conflicts</NuGetBuildTasksPackTargets>` stuff at the top of the file)

IIRC that was all needed because of you using mono somewhere on the backend build system, but the latest .NET Core SDK can build the `net452` target as well, so you could probably drop mono.

I also think that newer mono versions have added the packing stuff, so that would be another reason you could remove that.

kjnilsson:
@bording thanks - I'll remove it.

PR #682, opened 2020-01-07
DO NOT MERGE Reintroduce locking around socket writes

michaelklishin:
This reintroduces locking around socket writes when transmitting multiple frames at once.

While changes in #350, #354 are supposed to be safe, there's some
evidence (#681) that sometimes they are not which leads to incorrect
byte stream interleaving on the wire.


michaelklishin:
Superseded by #683.

PR #761, opened 2020-03-17
Wrapped concurrentqueue + semaphore pair together for readability, an…

michac:
…d unblocked recovery main loop during retry delay.

## Proposed Changes

In the process of changing the AutorecoveringConnection to run via the TPL versus having its own dedicated thread I noticed that the Task.Delay operations for recovery retries were being awaited. My expectation with the RecoveryLoop was that it would be able to process commands that came in during the delay, which it cannot do if it's awaiting the delay inside of a command handler.

I also think that if we're awaiting a Task.Delay without passing it a cancellation token then we won't be able to easily cancel out of an auto-recovering connection that's in a Delay, it will instead timeout in StopRecoveryLoop if RequestConnectionTimeout is less than NetworkRecoveryInterval and log a warning. I solved this by not awaiting the Task.Delay operation, it could also be solved by passing a CancellationToken.

Also I wrapped the semaphore and concurrentqueue together in a class to clarify that they were acting together (i.e. Enqueue should be followed by semaphore.Release()). This improved the readability in my opinion, but I'd be happy to take this part out if it's not popular.

## Types of Changes

- [ x] Bug fix (non-breaking change which fixes issue #NNNN)
- [ x] New feature (non-breaking change which adds functionality)

## Checklist

- [ x] I have read the `CONTRIBUTING.md` document
- [ x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)

## Further Comments

This is an identical copy of https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/752 , I just squashed the 6 commits to one to make it less noisy. Feel free to dismiss this or the other one to get the count back to one. I also updated to a personal fork instead of one in my organization, to see if that made it possible for others to push changes. @lukebakken


michaelklishin:
Is this still relevant given the recent changes such as #772 and #775?

michaelklishin:
OK, I see what this really affects, so it seems to be relevant.

bording:
@michaelklishin This was something I was still planning on taking a closer look at. I'll still try to do that and if I see anything important.

michaelklishin:
@bording oh, OK, sorry. We can always amend this PR, of course.

Issue #764, opened 2020-03-23
6.0: TLS connection fails when SslProtocols.None is unavailable or effectively disabled


BasieP:
I tried upgrading to the rabbitmq.client version 6 (latest RC).

After i upgraded the client all my code was the same, but i got the BrokerUnreachable exception.

After some investigation i figured out that the ConnectionFactory no longer recognises uri's that start with amqps:// as amqps.

I feed my ConnectionFactory just a complete Uri.
In my case:  amqps://user:pass@myserver/myVhost

When i check the properties of the ConnectionFactory object i just created i see that he fills in most fields correctly based on the connectionstring i supplied, except for the protocol.

the 'Endpoint' property now is like this:
"amqp://myserver:5671"

so short version:
I can't use amqps with a connectionstring in the new client. And I could in the old one.

michaelklishin:
I have added assertions on whether TLS is enabled on a `ConnectionFactory` in d8c3020beb4dd35dc109c9afeebc989fc9eb7fce and cannot reproduce. Please put together a small executable example.

michaelklishin:
The suite above does not require a RabbitMQ node to be running locally and can be executed using

```sh
dotnet test projects/client/Unit --filter "FullyQualifiedName~RabbitMQ.Client.Unit.TestAmqpUri"
```

michaelklishin:
@BasieP an effectively identical case is covered in `RabbitMQ.Client.Unit.TestAmqpUri`. Please provide evidence (e.g. a dump of object properties and server logs or a traffic capture) of your findings. It would also be great if you could run the above test in your environment.

michaelklishin:
Also it would be incredibly helpful if this report provided a reasonably complete amount of information. Without it our team can only guess what is going on (at least given the above test suite modifications) and guessing is a very expensive way to spend our time.

Things such as

 * *Complete* exception stack trace
 * Server logs that would [contain connection event messages](https://www.rabbitmq.com/logging.html#connection-lifecycle-events)
 * An executable example e.g. in a public repository on GitHub

BasieP:
Hi there,

I don't quite understand...
It's very easy to reproduce. It's just 3 lines i posted above:

I don't have access to server logs, but the server works. (i can connect with older clients)
The exact version i use of the rabbitmq client is: 6.0.0-pre9
The entire stacktrace:
```
RabbitMQ.Client.Exceptions.BrokerUnreachableException
  HResult=0x80131620
[...]

PR #766, opened 2020-03-23
Fall back to TLSv1.2 when System.Security.Authentication.SslProtocols.None is disabled/unavailable

michaelklishin:
## Proposed Changes

See #764 for the background. In some environments, we cannot assume
that SslProtocols.None will behave as expected:

 * It can be disabled via app context and other means
 * Its effective behavior is different between .NET versions according
   to .NET community blog posts and GitHub issues

So we borrow from [1] and special case this exception to retry with an
explicitly defined version (TLSv1.2, same as modern .NET default).

The discussion in [2] should also be mentioned as they have led us
to a much better understanding of many aspects of this intricate setting.

## Types of Changes
- [x] Bug fix (non-breaking change which fixes issue #764)

## Checklist

## Further Comments

Should address #764.

1. https://github.com/mysql-net/MySqlConnector/commit/715c3b542a1862b308f8afbb535ba1fee34afb93
2. https://github.com/robinrodricks/FluentFTP/issues/452#issuecomment-563965519

### DIFF
--- a/projects/client/RabbitMQ.Client/src/client/api/SslOption.cs
+++ b/projects/client/RabbitMQ.Client/src/client/api/SslOption.cs
@@ -149,5 +149,16 @@ namespace RabbitMQ.Client
         /// Retrieve or set the Ssl protocol version.
         /// </summary>
         public SslProtocols Version { get; set; }
+
+        /// <summary>
+        /// Reconfigures the instance to enable/use TLSv1.2.
+        /// Only used in environments where System.Security.Authentication.SslProtocols.None
+        /// is unavailable or effectively disabled, as reported by System.Net.ServicePointManager.
+        /// </summary>
+        public SslProtocols UseFallbackTlsVersions()
+        {
+            this.Version = SslProtocols.Tls12;
+            return Version;
+        }
     }
 }
--- a/projects/client/RabbitMQ.Client/src/client/api/SslOption.cs
+++ b/projects/client/RabbitMQ.Client/src/client/api/SslOption.cs
@@ -75,7 +75,7 @@ namespace RabbitMQ.Client
         }
 
         /// <summary>
-        /// Retrieve or set the set of ssl policy errors that are deemed acceptable.
+        /// Retrieve or set the set of TLS policy errors that are deemed acceptable.
         /// </summary>
         public SslPolicyErrors AcceptablePolicyErrors { get; set; }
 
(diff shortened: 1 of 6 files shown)


projects/client/RabbitMQ.Client/src/client/api/SslOption.cs
```diff
@@ -149,5 +149,16 @@ public X509CertificateCollection Certs
         /// Retrieve or set the Ssl protocol version.
         /// </summary>
         public SslProtocols Version { get; set; }
+
+        /// <summary>
+        /// Reconfigures the instance to enable/use TLSv1.2.
+        /// Only used in environments where System.Security.Authentication.SslProtocols.None
+        /// is unavailable or effectively disabled, as reported by System.Net.ServicePointManager.
+        /// </summary>
+        public SslProtocols UseFallbackTlsVersions()
```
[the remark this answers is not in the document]
michaelklishin: Fair enough.

projects/client/Unit/src/unit/APIApproval.Approve.approved.txt
```diff
@@ -15,6 +15,7 @@ namespace RabbitMQ.Client
         public int Port { get; set; }
         public RabbitMQ.Client.IProtocol Protocol { get; }
         public RabbitMQ.Client.SslOption Ssl { get; set; }
+        public RabbitMQ.Client.SslOption Tls { get; set; }
```
bording: What's going on here? I don't see anything in the PR that would cause this to show up.

bording:
FYI there will be some conflicts between this and #763 because I've made `SslHelper` internal and moved the file to to the impl folder.

projects/client/Unit/src/unit/APIApproval.Approve.approved.txt
michaelklishin: I have added this alias in `master` as part of the initial investigation of #764 where I added some more URI parsing tests. `APIApproval` expectations were not updated there, I will do so now. So the code change does not show up here because it's already in master but the `APIApproval` change is not.

bording: Can you remove it instead? Not sure why we need two different values here.

bording:
> The dream of eradicating "SSL" everywhere we can over a decade
after the name has been deprecated is still alive, though.

@michaelklishin Since we're already making breaking changes for 6.0, we could consider just renaming this class and dropping "SSL".

projects/client/Unit/src/unit/APIApproval.Approve.approved.txt
michaelklishin: Done. My dream of eradicating "SSL" everywhere we can in favor of "TLS" is still alive, though 😆

michaelklishin:
Given that 6.0 introduces substantial resource footprint improvements, I'd like the upgrade to it to be as straightforward as possible, so renaming can wait. "SSL" in naming is probably not something many care about anyway.

michaelklishin:
@bording I suspect this PR will go in first as it is arguably a bug fix. I'm afraid some adjustments ot your PR would be necessary. I can do them if you want.

bording:
> @bording I suspect this PR will go in first as it is arguably a bug fix. I'm afraid some adjustments ot your PR would be necessary. I can do them if you want.

I can make the adjustments.

PR #774, opened 2020-03-24
Drop unneeded shutdown

danielmarbach:
## Proposed Changes

Looks like the code is dead and no longer used


michaelklishin:
Thank you!

michaelklishin:
Backported to `5.x`.

PR #787, opened 2020-03-28
Remove more unneeded directories

bording:
If we're getting rid of [unneeded directories](https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/786), can I suggest a few more tweaks?


michaelklishin:
Thank you!

PR #818, opened 2020-04-17
Fall back to rabbitmqctl (or rabbitmqctl.bat) if RABBITMQ_RABBITMQCTL_PATH or umbrella path are not available

michaelklishin:
Per discussion with @lukebakken


lukebakken (approved):
Works great, tested on Debian Buster

PR #824, opened 2020-05-04
Optimizing memory usage even further.

stebet:
## Proposed Changes

After [writing a blog on the allocation fixes](https://stebet.net/real-world-example-of-reducing-allocations-using-span-t-and-memory-t/) and publishing last weekend, I got a few pointers from @davidfowl on saving even more allocations by using ArrayPool instead of MemoryPool. I also cleaned up some related code while I was at it.

## Types of Changes

## Checklist

### DIFF
--- a/projects/RabbitMQ.Client/client/impl/Frame.cs
+++ b/projects/RabbitMQ.Client/client/impl/Frame.cs
@@ -251,44 +251,35 @@ namespace RabbitMQ.Client.Impl
                 ProcessProtocolHeader(reader);
             }
 
-            byte[] headerBytes = null;
+            Span<byte> headerBytes = stackalloc byte[6];
+            reader.Read(headerBytes);
+            int channel = NetworkOrderDeserializer.ReadUInt16(headerBytes);
+            int payloadSize = NetworkOrderDeserializer.ReadInt32(headerBytes.Slice(2)); // FIXME - throw exn on unreasonable value
+            byte[] payloadBytes = ArrayPool<byte>.Shared.Rent(payloadSize);
+            Memory<byte> payload = new Memory<byte>(payloadBytes, 0, payloadSize);
+            int bytesRead = 0;
             try
             {
-                headerBytes = ArrayPool<byte>.Shared.Rent(6);
-                Memory<byte> headerSlice = new Memory<byte>(headerBytes, 0, 6);
-                reader.Read(headerSlice);
-                int channel = NetworkOrderDeserializer.ReadUInt16(headerSlice);
-                int payloadSize = NetworkOrderDeserializer.ReadInt32(headerSlice.Slice(2)); // FIXME - throw exn on unreasonable value
-                byte[] payloadBytes = ArrayPool<byte>.Shared.Rent(payloadSize);
-                Memory<byte> payload = new Memory<byte>(payloadBytes, 0, payloadSize);
-                int bytesRead = 0;
-                try
+                while (bytesRead < payloadSize)
                 {
-                    while (bytesRead < payloadSize)
-                    {
-                        bytesRead += reader.Read(payload.Slice(bytesRead, payloadSize - bytesRead));
-                    }
+                    bytesRead += reader.Read(payload.Slice(bytesRead, payloadSize - bytesRead));
                 }
-                catch (Exception)
-                {
-                    // Early EOF.
-                    ArrayPool<byte>.Shared.Return(payloadBytes);
-                    throw new MalformedFrameException($"Short frame - expected to read {payloadSize} bytes, only got {bytesRead} bytes");
-                }
-
-                int frameEndMarker = reader.ReadByte();
-                if (frameEndMarker != Constants.FrameEnd)
-                {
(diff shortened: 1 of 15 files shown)


stebet:
Example of allocation reductions (using my benchmark app, sending and reciving 50.000 events with 512 byte payloads).

Before:
512AllocsAfter

After:
512Optimized

stebet:
I didn't notice https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/816 which makes the same change to the contract. I'll make changes to accomodate. I do think adding a method to the contracts should not require a major change as additions are normally not considered breaking, unless they might affect certain overloads, so we could add the ReadOnlyMemory<byte> overload to the batch publishing interface and still keep the byte[] version as well. What do you think @bording?

davidfowl:
I see a bunch of low hanging fruit:

- What’s allocating the Task<bool>? That should be cached as there’s only 2 values.
- Why is that Func being allocated?

projects/Unit/APIApproval.Approve.verified.txt
```diff
@@ -259,6 +259,7 @@ namespace RabbitMQ.Client
     public interface IBasicPublishBatch
     {
         void Add(string exchange, string routingKey, bool mandatory, RabbitMQ.Client.IBasicProperties properties, byte[] body);
+        void Add(string exchange, string routingKey, bool mandatory, RabbitMQ.Client.IBasicProperties properties, System.ReadOnlyMemory<byte> body);
```
bording, line 262: Unfortunately since this is a change to an interface, this means this is a breaking change. If you're wanting to get this in to a release before 7.0, you can't do this.

michaelklishin: I think this should go into 7.0 which we can release fairly soon. @lukebakken WDYT?

lukebakken: As far as I know there are quite a few things that have been discussed for 7.0, one of which is a fully-async client. I think @stebet has done some work on that.

If "fairly soon" means within 6 months... maybe???

bording:
> I do think adding a method to the contracts should not require a major change as additions are normally not considered breaking, unless they might affect certain overloads, 

That is only true when you're talking about a class. Any changes to an interface are breaking because any existing implementations of that interface break. That's why default interface methods were added to C# 8.

stebet:
I'll take a closer look and follow up later tonight. Thanks for this @davidfowl and @bording. It makes sense to skip the interface change as most of this is very applicable to 6.X.

projects/Unit/APIApproval.Approve.verified.txt
stebet: Yeah, I've done some research and experiments. It'd probably require quite a lot of changes to the public interface, all of which are good i.m.o, like departing from Event based notifications (can be replaced with a pluggable interface), but I'll write up a proposal soon for that. Reverting any interface changes for now :)

lukebakken:
@stebet I'm going to test and review this today. Since there are no interface changes, I assume this is OK for `6.1.0`

projects/RabbitMQ.Client/client/impl/Frame.cs
```diff
@@ -236,42 +251,51 @@ internal static InboundFrame ReadFrom(Stream reader)
                 ProcessProtocolHeader(reader);
             }
 
-            using (IMemoryOwner<byte> headerMemory = MemoryPool<byte>.Shared.Rent(6))
+            byte[] headerBytes = null;
+            try
             {
-                Memory<byte> headerSlice = headerMemory.Memory.Slice(0, 6);
+                headerBytes = ArrayPool<byte>.Shared.Rent(6);
```
davidfowl, line 257: stackalloc?

lukebakken: @stebet I'm assuming that instead of using `new byte[6]` as a field in the class you're doing this because it uses the memory buffer pool (reducing allocation churn) and supports `.Slice(2)` on this line:

```
int payloadSize = NetworkOrderDeserializer.ReadInt32(headerSlice.Slice(2));
```

Is that a correct understanding?

stebet: Yes. stackalloc might work there. Good catch. Also, @davidfowl , I'd there a simple way to reuse/cache a TaskCompletionSource? I'm thinking of the ConcurrentQueue work-item notifier in the AsyncEventingSubscriber. SemaphoreSlim seemed to perform a lot worse for signaling than just completing a TCS and creating a new one.

stebet: Added support for stackalloc.

davidfowl: ConcurrentQueue + SemaphoreSlim = Channels. It’ll do what you need and be more efficient.

stebet: Adding channels in https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/835

PR #844, opened 2020-05-21
Frame optimizations

stebet:
## Proposed Changes

- Buffers outbound frames using a Channel, and create a background task that consumes from that channel and writes to the socket. This makes for more efficient buffering of outbound frames as well as removes that write-lock from the SocketFrameHandler since there is only ever one writer accessing the socket.
- As a result, we also get rid of a temporary list allocation when writing frame sets, and we can simplify the command->frame splitting code as well as simplify the socket writers.

## Types of Changes

## Checklist

### DIFF
--- a/projects/RabbitMQ.Client/client/impl/Command.cs
+++ b/projects/RabbitMQ.Client/client/impl/Command.cs
@@ -99,63 +99,19 @@ namespace RabbitMQ.Client.Impl
 
         internal void Transmit(int channelNumber, Connection connection)
         {
-            if (Method.HasContent)
-            {
-                TransmitAsFrameSet(channelNumber, connection);
-            }
-            else
-            {
-                TransmitAsSingleFrame(channelNumber, connection);
-            }
-        }
-
-        internal void TransmitAsSingleFrame(int channelNumber, Connection connection)
-        {
             connection.WriteFrame(new MethodOutboundFrame(channelNumber, Method));
-        }
-
-        internal void TransmitAsFrameSet(int channelNumber, Connection connection)
-        {
-            var frames = new List<OutboundFrame> { new MethodOutboundFrame(channelNumber, Method) };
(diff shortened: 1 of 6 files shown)


stebet:
Memory usage before:
Before

Memory usage after:
After

CPU usage before:
CPUBefore

CPU usage after:
CPUAfter

stebet:
One failed test... taking a look.

stebet:
Ahhh, the test is broken :) It's failing to copy the buffers of the received messages, and is therefore breaking our new rule in 6.0 to make sure to copy the contents of the message body in the event handlers.

michaelklishin:
@stebet please rebase this on top of `master`. Thank you.

stebet:
Arrghh.... think I royally messed up this rebase...

stebet:
Ok, think I managed to fix it :)

michaelklishin:
@stebet should I backport this?

stebet:
Should be safe. @bording ?

bording:
I think the rebase was still not right, because it looks like this is PR changed the site submodule, which I wouldn't expect this PR to touch.

> Should be safe. @bording ?

Yeah I think this would be okay to bring over to 6.1.0. There's no behavior or public API changes.

michaelklishin:
Backported to `6.x`.

stebet:
The submodule keeps giving me a headache, and sneaking in there. Is there a way to ignore it automatically?

bording:
@michaelklishin I think you'll need to fix the submodule on master and 6.x

michaelklishin:
I wouldn't do it even if there was. Don't worry about it, docs can always be regenerated and we are not
introducing any meaningful enough API changes most of the time. When we do for 7.0, doc updates
will have to be a part of the release, including rabbitmq-tutorials and website updates.
________________________________

PR #845, opened 2020-05-24
simplify WriteShortstr

bollhals:
## Proposed Changes

Simplifying write of string values. GetBytes returns the number of bytes written, so there is no need to precalculate the length.

## Types of Changes

## Checklist

## Further Comments

Question: There is a bug in the code (before this and also after this) that if you write a string resulting in more than 255 bytes, the length variable (byte) will be cut off and the resulting deserialized value will be different from the source. Shall I open an issue for this?

### DIFF
--- a/projects/RabbitMQ.Client/client/impl/WireFormatting.cs
+++ b/projects/RabbitMQ.Client/client/impl/WireFormatting.cs
@@ -293,11 +293,11 @@ namespace RabbitMQ.Client.Impl
             {
                 case string val:
                     memory.Span[0] = (byte)'S';
-                    if (MemoryMarshal.TryGetArray(memory.Slice(5, Encoding.UTF8.GetByteCount(val)), out ArraySegment<byte> segment))
+                    if (MemoryMarshal.TryGetArray(memory, out ArraySegment<byte> segment))
                     {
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you!

bollhals:
> Question: There is a bug in the code (before this and also after this) that if you write a string resulting in more than 255 bytes, the length variable (byte) will be cut off and the resulting deserialized value will be different from the source.
Shall I open an issue for this or is this "as designed"?

stebet:
An AMQP shortstr value can only contain 256 bytes: https://www.rabbitmq.com/amqp-0-9-1-reference.html#domain.shortstr

bollhals:
Yes, that I understand, but the current code does the following for payload all 'A':

| real length | length | payload |
|------------ |------------:|-----------:|
| 255 | 255| 65, 65, ... (255 times) |
| 256 | 0 | 65, 65, ... (256 times) |

So when you read it back, it reads wrong values (due to the wrong ones sent). 

There is nowhere a prevention built in to not being able to send more than 255 chars.

michaelklishin:
We should fail early and visibly so yes @bollhals, this is worth changing, even though this is how the protocol is designed and not really a bug.

michaelklishin:
Backported to `6.x`.

PR #854, opened 2020-05-26
simplify missed heartbeat recognition

bollhals:
## Proposed Changes

The `NotifyHeartbeatListener()` was showing up on my analysis. Even though it is fairly small and quick, but it is called for every received frame.

I think this simplified implementation should be fast and as safe as before. 
Some comments to the implementation:
- AFAIK _heartbeatCounter is only updated by a single thread (MainLoop), so no additional thread safety needed.
- The reading do not need to be 100% exact as 1) as long as they're different it's good enough and 2) they're called so little, they won't suffer reading the wrong cached value and even if, nothing changes until 8 missed heartbeat anyway.

## Types of Changes

## Checklist

### DIFF
--- a/projects/RabbitMQ.Client/client/impl/Connection.cs
+++ b/projects/RabbitMQ.Client/client/impl/Connection.cs
@@ -83,12 +83,10 @@ namespace RabbitMQ.Client.Framing.Impl
         //
         // Heartbeats
         //
-
         private TimeSpan _heartbeat = TimeSpan.Zero;
         private TimeSpan _heartbeatTimeSpan = TimeSpan.FromSeconds(0);
         private int _missedHeartbeats;
-        private int _heartbeatCounter;
-        private int _lastHeartbeatCounter;
+        private bool _heartbeatDetected;
 
         private Timer _heartbeatWriteTimer;
         private Timer _heartbeatReadTimer;
@@ -616,14 +614,9 @@ namespace RabbitMQ.Client.Framing.Impl
             }
         }
 
-        public void NotifyHeartbeatListener()
+        private void NotifyHeartbeatListener()
         {
-            // https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/checked-and-unchecked
-            // No worries if this overflows. What matters is that the value changes.
-            unchecked
-            {
-                _heartbeatCounter++;
-            }
+            _heartbeatDetected = true;
         }
 
         public void NotifyReceivedCloseOk()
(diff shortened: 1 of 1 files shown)


michaelklishin:
Yes, we don't need perfect precision here simply for the fact that the absolute recommended minimum for heartbeat timeout is 5 seconds, and in most cases something closer to 15-60 seconds would be reasonable.

michaelklishin:
Actually we would overflow `int` in less than day since every delivery will use at least three frames:

```
2147483647/(10000*3*3600*24)
```

An unsigned 64-bit integer would last us quite a bit longer (over 1900 years or so if my math is correct).

lukebakken:
@michaelklishin I tested [this commit](https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/854/commits/1af867acfb2a1a85542b16c1c0a8e7ccf674483e) using the following code. As @bollhals states we don't care what the value of `_heartbeatCounter` is, just that it changes.

```
lbakken@shostakovich ~/issues/rabbitmq/rabbitmq-dotnet-client/test-overflow
$ cat Program.cs 
using System;

namespace test_overflow
{
    class Program
    {
        {
            int val;

            unchecked
            {
                val = int.MaxValue + 2;
            }

            Console.WriteLine("0 int.MaxValue += 2 {0}", val);

            unchecked
            {
                val = val + 2;
            }

            Console.WriteLine("1 val = val + 2 {0}", val);                                                                                                                                                                                                                      
        }                                                                                                                                                                                                                                                                       
    }                                                                                                                                                                                                                                                                           
}
                                                                                                                                                                                                                                                                               
lbakken@shostakovich ~/issues/rabbitmq/rabbitmq-dotnet-client/test-overflow                                                                                                                                                                                                     
$ dotnet run                                                                                                                                                                                                                                                                    
0 int.MaxValue += 2 -2147483647                                                                                                                                                                                                                                                 
1 val = val + 2 -2147483645
```

michaelklishin:
OK, this looks like a simplification overall then. Let's wait for some profiler numbers to assess the difference.

bollhals:
Thanks for the comments, I thought about the overflow but somehow I thought that it will never happen. Glad you did the math.

I've checked again the implementation and I think we can even make it simpler. Since we don't care about the value, just that it changed, it should also be implementable with just 1 normal bool variable.

bollhals:
> > The NotifyHeartbeatListener() was showing up on my analysis
> 
> Can we get more details about what "showing up on my analysis" means?

Sure, this is from the test application of stebet:
image

bollhals:
image
With the bool implementation (I'll push it soon)

bollhals:
> @bollhals I am probably mis-reading the screenshots in [this comment](https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/854#issuecomment-634841287), but it seems like the number of method calls to `NotifyHeartbeatListener` should be equal to the number of calls to `ReadFrame`. You can see that in your first screenshot, but the second one shows far fewer calls to `NotifyHeartbeatListener` which seems to suggest that the code run in the second example isn't what is shown in this PR. Please correct me if I am mistaken!

I wasn't sure either, and I had to do something else first.
But to me it looks like the method iteself gets inlined at some point later from the JIT or something else, because when I specify NoInlining to the method, the result is this:
image
image

lukebakken:
@bollhals awesome, I really appreciate the explanation.

michaelklishin:
Backported to `6.x`.

PR #855, opened 2020-05-31
remove task yield

bollhals:
## Proposed Changes

Removed the task.yield in the Work baseclass as well as the try catch, as there is already a try catch at the callee (Workpool of AsyncConsumerWorkService). This eliminates another chunk of allocations.

## Types of Changes

## Checklist

### DIFF
--- a/projects/RabbitMQ.Client/client/impl/Work.cs
+++ b/projects/RabbitMQ.Client/client/impl/Work.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using System.Threading.Tasks;
+﻿using System.Threading.Tasks;
 
 namespace RabbitMQ.Client.Impl
 {
@@ -12,17 +11,9 @@ namespace RabbitMQ.Client.Impl
             _asyncConsumer = (IAsyncBasicConsumer)consumer;
         }
 
-        public async Task Execute(ModelBase model)
+        public Task Execute(ModelBase model)
         {
-            try
-            {
-                await Task.Yield();
-                await Execute(model, _asyncConsumer).ConfigureAwait(false);
-            }
-            catch (Exception)
-            {
-                // intentionally caught
-            }
+            return Execute(model, _asyncConsumer);
         }
 
         protected abstract Task Execute(ModelBase model, IAsyncBasicConsumer consumer);


bollhals:
I don't see a reason for the task.yield.

From Testapp:
Before:
image
After:
image

=> 71 MB -> 65 MB

michaelklishin:
So an 8.5% improvement. This is certainly meaningful. I will trust your and @stebet's judgment on the need to yield.

michaelklishin (approved):

lukebakken:
@bollhals actually this has been discussed. Searching this repo for the string `yield` turns up this issue ...

https://github.com/rabbitmq/rabbitmq-dotnet-client/issues/341

...and this PR...

https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/760

I will revert the changes here.

lukebakken:
Reverted by aa5e8cb

bollhals:
> @bollhals actually this has been discussed. Searching this repo for the string `yield` turns up this issue ...
> 
> #341
> 
> ...and this PR...
> 
> #760
> 
> I will revert the changes here. We can't remove the yield until the client is refactored to be async, basically.

I'd like to understand how this pervents the deadlock, but I haven't been able to find a test case to reproduce it. (I'd like to add one if there isn't. So it is clear to everyone why it is needed)
From the comments in #341 I wasn't able to break it, even with [the code provided here](https://gist.github.com/karoberts/bc10a6aa57d9bcfeda10fca86b68ffca).

bollhals:
Now that I think of it, Is it possible that #844 fixed the issue so it isn't needed anymore?

lukebakken:
This is something @danielmarbach and / or @bording should weigh in on if they have time. No hurry at all.

danielmarbach:
I looked at the current implementation on master and I think the changes proposed here should be fine. If I recall correctly (and I doubt that given my bad memories) we basically introduced the yield as a precaution. Around the timeframe when the first version of the code was introduced we didn't really have a good possibility to run continuations async other than doing some nasty reflection tricks, or wrap the task completion source TrySet methods with Task.Run or enforce async continuations by doing the `Task.Yield`. With the current code either using `RunContinuationsAsynchronously` or setting the channels continuation behavior to `AllowSynchronousContinuations = false` this problem should be mitigated I think although I haven't spend a lot of time digging through everything.

PR #858, opened 2020-06-02
Unify on IModel

danielmarbach:
## Proposed Changes

The worker service did mostly use `IModel` exception when the model was added with safely casted it to `ModelBase` but never actually verified if the model implements `ModelBase`. Any model not deriving from `ModelBase` would cause a NullRef. Given that the `ModelBase` method is only used in the exception case I moved the conversion to `ModelBase` into the worker implementation. 

Of course this change has the tradeoff of making now everything an interface dispatch as was as safe casting on the exception path (which is slow anyway) so I'll leave it up to you if you think this is good or not. I'd suggest though if this PR is not taken to throw a meaningful exception if the model cannot be converted to model base.

## Types of Changes

## Checklist

## Further Comments

-

### DIFF
--- a/projects/RabbitMQ.Client/client/impl/BasicDeliver.cs
+++ b/projects/RabbitMQ.Client/client/impl/BasicDeliver.cs
@@ -18,12 +18,12 @@ namespace RabbitMQ.Client.Impl
         readonly IBasicProperties _basicProperties;
         readonly ReadOnlyMemory<byte> _body;
 
-        public BasicDeliver(IBasicConsumer consumer, 
-            string consumerTag, 
-            ulong deliveryTag, 
-            bool redelivered, 
-            string exchange, 
-            string routingKey, 
+        public BasicDeliver(IBasicConsumer consumer,
+            string consumerTag,
+            ulong deliveryTag,
+            bool redelivered,
+            string exchange,
+            string routingKey,
             IBasicProperties basicProperties,
             ReadOnlyMemory<byte> body) : base(consumer)
         {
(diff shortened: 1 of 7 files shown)


projects/RabbitMQ.Client/client/impl/BasicDeliver.cs
```diff
@@ -18,12 +18,12 @@ sealed class BasicDeliver : Work
         readonly IBasicProperties _basicProperties;
         readonly ReadOnlyMemory<byte> _body;
 
-        public BasicDeliver(IBasicConsumer consumer,
```
danielmarbach, line 21: Happy to roll back if you want to keep the white space

michaelklishin:
@danielmarbach the callback exception path should not generally be executed very often. However, maybe we should borrow a complete enough (using real publishers and consumers) benchmark from one of the recent PRs to have some data to make a more informed decision?

Just curious, was some kind of mock implementation of `IModel` used in your case?

danielmarbach:
No mock. I looked at it from a more open/close and liskov principle. I'm not disappointed if this is thrown away and I also don't want to spend too much time on this. For the recent ping I had to dig into the worker services and I found this weird state of IModel vs ModelBase. Give I found no guidance which one should be defacto standard I figured I send in a suggestion that uses the most abstract one which is the interface

danielmarbach:
Any thoughts on this, is it worth a rebase or not?

projects/RabbitMQ.Client/client/impl/BasicDeliver.cs
michaelklishin: Nah, turning down a PR because of a single whitespace change sounds unreasonable :)

danielmarbach:
Ok rebased

michaelklishin:
@danielmarbach can you please submit this based on the `6.x` branch? There are conflicts that are not obvious to me.

danielmarbach:
Here we go https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/898

PR #866, opened 2020-06-15
Ability to do concurrent dispatches both on the async as well as the sync consumer

danielmarbach:
## Proposed Changes

An alternative to https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/806

If the interface is not extended then this can be added in a non breaking way on the level of the connection factory and later could be propagated to the interfaces in v7. I currently named the property `ProcessingConcurrency` to convey the intent for both the sync and the async case

## Types of Changes

## Checklist

## Further Comments

considered, etc.

### DIFF
--- a/projects/RabbitMQ.Client/client/api/ConnectionFactory.cs
+++ b/projects/RabbitMQ.Client/client/api/ConnectionFactory.cs
@@ -175,6 +175,16 @@ namespace RabbitMQ.Client
         /// </summary>
         public bool DispatchConsumersAsync { get; set; } = false;
 
+        /// <summary>
+        /// Set to a value greater than one to enable concurrent processing. For a concurrency greater than one <see cref="IBasicConsumer"/>
+        /// will be offloaded to the worker thread pool so it is important to choose the value for the concurrency wisely to avoid thread pool overloading.
+        /// <see cref="IAsyncBasicConsumer"/> can handle concurrency much more efficiently due to the non-blocking nature of the consumer.
+        /// Defaults to 1.
+        /// </summary>
+        /// <remarks>For concurrency greater than one this removes the guarantee that consumers handle messages in the order they receive them.
+        /// In addition to that consumers need to be thread/concurrency safe.</remarks>
+        public int ProcessingConcurrency { get; set; } = 1;
+
         /// <summary>The host to connect to.</summary>
         public string HostName { get; set; } = "localhost";
 
--- a/projects/RabbitMQ.Client/client/impl/AsyncConsumerWorkService.cs
+++ b/projects/RabbitMQ.Client/client/impl/AsyncConsumerWorkService.cs
@@ -9,7 +9,12 @@ namespace RabbitMQ.Client.Impl
     internal sealed class AsyncConsumerWorkService : ConsumerWorkService
     {
         private readonly ConcurrentDictionary<IModel, WorkPool> _workPools = new ConcurrentDictionary<IModel, WorkPool>();
-        private readonly Func<IModel, WorkPool> _startNewWorkPoolFunc = model => StartNewWorkPool(model);
+        private readonly Func<IModel, WorkPool> _startNewWorkPoolFunc;
+
+        public AsyncConsumerWorkService(int concurrency) : base(concurrency)
+        {
+            _startNewWorkPoolFunc = model => StartNewWorkPool(model);
+        }
 
         public void Schedule<TWork>(ModelBase model, TWork work) where TWork : Work
         {
@@ -22,9 +27,9 @@ namespace RabbitMQ.Client.Impl
             workPool.Enqueue(work);
         }
 
-        private static WorkPool StartNewWorkPool(IModel model)
+        private WorkPool StartNewWorkPool(IModel model)
         {
-            var newWorkPool = new WorkPool(model as ModelBase);
+            var newWorkPool = new WorkPool(model as ModelBase, _concurrency);
             newWorkPool.Start();
             return newWorkPool;
         }
@@ -44,16 +49,29 @@ namespace RabbitMQ.Client.Impl
             readonly Channel<Work> _channel;
             readonly ModelBase _model;
             private Task _worker;
+            private readonly int _concurrency;
+            private SemaphoreSlim _limiter;
+            private CancellationTokenSource _tokenSource;
 
-            public WorkPool(ModelBase model)
+            public WorkPool(ModelBase model, int concurrency)
             {
+                _concurrency = concurrency;
                 _model = model;
                 _channel = Channel.CreateUnbounded<Work>(new UnboundedChannelOptions { SingleReader = true, SingleWriter = false, AllowSynchronousContinuations = false });
             }
 
             public void Start()
             {
-                _worker = Task.Run(Loop, CancellationToken.None);
+                if (_concurrency == 1)
+                {
+                    _worker = Task.Run(Loop, CancellationToken.None);
+                }
+                else
+                {
+                    _limiter = new SemaphoreSlim(_concurrency);
+                    _tokenSource = new CancellationTokenSource();
+                    _worker = Task.Run(() => LoopWithConcurrency(_tokenSource.Token), CancellationToken.None);
+                }
             }
 
             public void Enqueue(Work work)
@@ -83,9 +101,55 @@ namespace RabbitMQ.Client.Impl
                 }
             }
 
+            async Task LoopWithConcurrency(CancellationToken cancellationToken)
+            {
+                try
+                {
+                    while (await _channel.Reader.WaitToReadAsync(cancellationToken).ConfigureAwait(false))
+                    {
+                        while (_channel.Reader.TryRead(out Work work))
+                        {
+                            // Do a quick synchronous check before we resort to async/await with the state-machine overhead.
+                            if(!_limiter.Wait(0))
+                            {
+                                await _limiter.WaitAsync(cancellationToken).ConfigureAwait(false);
+                            }
+
+                            _ = HandleConcurrent(work, _model, _limiter);
+                        }
+                    }
+                }
+                catch (OperationCanceledException)
+                {
+                    // ignored
+                }
+            }
+
+            static async Task HandleConcurrent(Work work, ModelBase model, SemaphoreSlim limiter)
+            {
+                try
+                {
+                    Task task = work.Execute(model);
+                    if (!task.IsCompleted)
+                    {
+                        await task.ConfigureAwait(false);
+                    }
+                }
+                catch (Exception)
+                {
+
+                }
+                finally
+                {
+                    limiter.Release();
+                }
+            }
+
             public Task Stop()
             {
                 _channel.Writer.Complete();
+                _tokenSource?.Cancel();
+                _limiter?.Dispose();
                 return _worker;
             }
         }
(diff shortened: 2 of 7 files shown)


danielmarbach:
@stebet what do you think about this spike

projects/RabbitMQ.Client/client/impl/AsyncConsumerWorkService.cs
```diff
@@ -9,7 +9,14 @@ namespace RabbitMQ.Client.Impl
...
     {
         private readonly ConcurrentDictionary<IModel, WorkPool> _workPools = new ConcurrentDictionary<IModel, WorkPool>();
-        private readonly Func<IModel, WorkPool> _startNewWorkPoolFunc = model => StartNewWorkPool(model);
+        private readonly Func<IModel, WorkPool> _startNewWorkPoolFunc;
+        private readonly int _concurrency;
+
+        public AsyncConsumerWorkService(int concurrency)
+        {
+            _concurrency = concurrency;
+            _startNewWorkPoolFunc = model => StartNewWorkPool(model);
```
danielmarbach, line 18: This would introduce a capture of `this`

stebet:
I'll take a look at this tomorrow :)

projects/RabbitMQ.Client/client/impl/AsyncConsumerWorkService.cs
```diff
@@ -83,9 +103,51 @@ async Task Loop()
...
 
+            async Task LoopWithConcurrency(CancellationToken cancellationToken)
+            {
+                try
+                {
+                    while (await _channel.Reader.WaitToReadAsync(cancellationToken).ConfigureAwait(false))
+                    {
+                        while (_channel.Reader.TryRead(out Work work))
+                        {
+                            await _limiter.WaitAsync(cancellationToken);
```
stebet, line 114: for a fast path this could be:
```csharp
// Do a quick synchronous check before we resort to async/await with the state-machine overhead.
if(!_limiter.Wait(0))
{
    await _limiter.WaitAsync(cancellationToken).ConfigureAwait(false);
}
```

projects/RabbitMQ.Client/client/api/ConnectionFactory.cs
```diff
@@ -175,6 +175,12 @@ public sealed class ConnectionFactory : ConnectionFactoryBase, IAsyncConnectionF
         /// </summary>
         public bool DispatchConsumersAsync { get; set; } = false;
 
+        /// <summary>
+        /// TODO.
+        /// Defaults to 1.
+        /// </summary>
+        public int AsyncConsumerConcurrency { get; set; } = 1;
```
danielmarbach, line 182: This is just a working name. Could be anything that is more meaningful.

stebet:
This looks good to me. I'm wondering if it'd be better to default AsyncConsumerConcurrency to Environment.ProcessorCount if it's enabled. That'd mean two properties though, AsyncConsumerConcurrencyEnabled (false when AsyncConsumerConcurrencyCount == 1, when set to true, would set AsyncConsumerConcurrencyCount = Environment.ProcessorCount if AsyncConsumerConcurrencyCount == 1).

Maybe that's just making things a bit convoluted perhaps... what do you think?

danielmarbach:
My thought process was that `Concurrency = 1` is just a special case of `Concurrency = N` that's why I went without having an additional bool, because then you only need to set it to something sensible which than automatically switches to fan out.

projects/RabbitMQ.Client/client/api/ConnectionFactory.cs
stebet: Would be good to set the description to clearly say that it defines the parallelism of the consumer processing, perhaps providing an example of setting it to Environment.ProcessorCount?

danielmarbach: I force pushed a more thorough explanation. Have a look

projects/RabbitMQ.Client/client/impl/AsyncConsumerWorkService.cs
```diff
@@ -83,9 +101,55 @@ async Task Loop()
...
+                    {
+                        while (_channel.Reader.TryRead(out Work work))
+                        {
+                            // Do a quick synchronous check before we resort to async/await with the state-machine overhead.
+                            if(!_limiter.Wait(0))
+                            {
+                                await _limiter.WaitAsync(cancellationToken).ConfigureAwait(false);
+                            }
+
+                            _ = HandleConcurrent(work, _model, _limiter);
```
stebet, line 118: <3 the clarity of intention by using the discard :)

projects/RabbitMQ.Client/client/impl/ConsumerWorkService.cs
```diff
@@ -105,10 +124,54 @@ async Task Loop()
...
+                    {
+                        // Do a quick synchronous check before we resort to async/await with the state-machine overhead.
+                        if(!_limiter.Wait(0))
+                        {
+                            await _limiter.WaitAsync(cancellationToken).ConfigureAwait(false);
+                        }
+
+                        _ = OffloadToWorkerThreadPool(action, _limiter);
+                    }
+                }
```
stebet, line 151: Will these two loops spin continuously (causing CPU load) since there is no Delay/Sleep between iterations when the queue is empty? Not sure of the mechanisms in the ConcurrentQueue implementation.

stebet: Would it make sense to have the `Enqueue` method trigger the loop when needed via. AutoResetEvent instead?

projects/RabbitMQ.Client/client/impl/AsyncConsumerWorkService.cs
danielmarbach: Some teams have extensions methods like

```
public static void IntentionallyNotAwaited|Ignored(this Task task) {
 //
}
```

but given we have the language constructs I think discard is fine. Glad you like it ;)

projects/RabbitMQ.Client/client/impl/ConsumerWorkService.cs
danielmarbach: The behavior of the `LoopWithConcurrency` is the same as `Loop` which is the existing behavior. The loop sits idle until something is in the queue. Once there is something in the queue it consumes as long as there is something to consume and then is idle again until there is something enqueued. It is correct that continues scheduling of work will "capture" the worker in the inner while. At this stage it is by design so I feel any discussions about changing that behavior should probably be separate from this PR?

stebet: Yeah, was just wondering because the AsyncConsumerWorkService has a Channel as a work item buffer where it can properly await until a message arrives in the channel, where the normal ConsumerWorkService is using the ConcurrentQueue instead. Just wondering if the rapid iteration in TryDequeue (which will return false if the queue is empty, causing a spinning loop until there is an item in there) will cause unneeded CPU usage. Couldn't that just as well be replaced with the Channel implementation as well, making the WorkServices pretty much identical apart from that one execute a synchronous action and the other one a Task based action?

danielmarbach: It doesn't case a spinning loop if there is nothing in the queue because the inner while is exited and then 

```
await _syncSource.Task.ConfigureAwait(false);
```

kick in which is not reset because there was nothing added to the queue. Therefore it just sits there. In case there was something added to the queue it falls through, set the new TCS and then dequeues until done.

> Couldn't that just as well be replaced with the Channel implementation as well, making the WorkServices pretty much identical apart from that one execute a synchronous action and the other one a Task based action?

Sure that would be possible and always was even when it was decided to diverge the async service to use channel. I'm fine to do this at a later stage, still I found it not relevant right now for the scope of this PR. If you have good arguments why those two changes should be bundled I'm happy to listen though

stebet: > It doesn't case a spinning loop if there is nothing in the queue because the inner while is exited and then
> 
> ```
> await _syncSource.Task.ConfigureAwait(false);
> ```

Guess I'm not awake yet... I completely missed that await :) The other changes can wait from my point of view, this should work just fine.

danielmarbach:
I've pushed a few changes to the tests to make them more reliable.

danielmarbach:
@michaelklishin @lukebakken any thoughts? This could go in 6.2 if you guys are willing to take it

lukebakken:
I've been busy with other work. I'll try to get to this PR today. We have some bug fixes that are planned for `6.1.1` but it sounds like `6.2.0` should be the next release.

danielmarbach:
No stress @lukebakken

lukebakken:
Thanks @danielmarbach and @stebet for the review.

danielmarbach:
@lukebakken where should we track whether this new property should be moved to the interface or not? I did deliberately not add it there to not add a breaking change. Ideally an issue that is added to the next major version with a breaking change label would be awesome

lukebakken:
@danielmarbach we can create a new issue for that. I'm mid-review / testing of #868 / #878 right now :smile:

shaneqld:
Have `BasicAck` and `BasicNack` been made thread-safe yet?  If not, users will run into strange exceptions (mostly protocol errors) when they work on highly concurrent workloads sans workarounds.

stebet:
> Have `BasicAck` and `BasicNack` been made thread-safe yet? If not, users will run into strange exceptions (mostly protocol errors) when they work on highly concurrent workloads sans workarounds.

Do you have a repro you can share? Would love to dig into if this is still a problem.

danielmarbach:
I'd be interested to see and hear this as well. We are doing `BasicAck` and `BasicNack` since multiple versions concurrently. Is that something that has regressed recently?

michaelklishin:
@shaneqld @danielmarbach @stebet acknowledgments of single deliveries are safe. Acknowledgments with `multiple` set to `true` can indeed cause [protocol channel exceptions](https://www.rabbitmq.com/confirms.html#consumer-acks-double-acking). But this can happen even with a single thread of execution. This is also not a novel problem, although I see that [our doc section on concurrency of consumers](https://www.rabbitmq.com/dotnet-api-guide.html#concurrency) does not cover this very well.

So I don't see how this approach changes anything for `basic.ack` and `basic.{nack,reject}` operations.

danielmarbach:
BTW how can I contribute to the doco for the new feature I added. I don't want you guys to need to do all the grunt workAm 25.06.2020 17:49 schrieb Michael Klishin <notifications@github.com>:
@shaneqld @danielmarbach @stebet concurrent acknowledgments of single deliveries are safe. Acknowledgments with multiple set to true can indeed cause protocol channel exceptions. But this can happen even with a single thread of execution. This is also not a novel problem, although I see that our doc section on concurrency of consumers does not cover this very well.
So I don't see how this approach changes anything for basic.ack and basic.{nack,reject} operations.

shaneqld:
> I'd be interested to see and hear this as well. We are doing `BasicAck` and `BasicNack` since multiple versions concurrently. Is that something that has regressed recently?

I ran into this in a previous version with a different implementation of concurrency, but have been unable to reproduce in this latest version.  When I have a moment I will update a project that has this workaround for the protocol error (and remove said workaround) and let you know if I run into any issues.

PR #887, opened 2020-07-02
use cached empty BasicProperties when null

bollhals:
## Proposed Changes

Instead of creating a new BasicProperties when the passed one is null, a cached empty one is used. This avoids creating a new instance every time.

## Types of Changes

## Checklist

## Further Comments

edited as I accidentally posted it too early

### DIFF
--- a/projects/RabbitMQ.Client/client/impl/BasicPublishBatch.cs
+++ b/projects/RabbitMQ.Client/client/impl/BasicPublishBatch.cs
@@ -57,7 +57,6 @@ namespace RabbitMQ.Client.Impl
 
         public void Add(string exchange, string routingKey, bool mandatory, IBasicProperties basicProperties, byte[] body)
         {
-            IBasicProperties bp = basicProperties ?? _model.CreateBasicProperties();
             var method = new BasicPublish
             {
                 _exchange = exchange,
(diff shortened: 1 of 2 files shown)


michaelklishin:
Makes perfect sense to me. Thank you!

michaelklishin:
Backported to `6.x`.

AStove:
Just because `internal readonly IBasicProperties _emptyBasicProperties;` is "readonly" doesn't mean you can't change the properties of the object that referenced by `_emptyBasicProperties`. It just means you can't assign a new object to it outside of the constructor.

I think it's fine now but it could introduce some bugs later if the objects decide to make modifications to the properties they receive and thus your `_emptyBasicProperties` gets changed aswell. Solution would be to clone the object on passing but then what have we achieved with caching it.

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/readonly
> Because reference types contain a reference to their data, a field that is a readonly reference type must always refer to the same object. That object isn't immutable. The readonly modifier prevents the field from being replaced by a different instance of the reference type. However, the modifier doesn't prevent the instance data of the field from being modified through the read-only field.

stebet:
The cached object is only used when sending data and no properties have been set, so it should never be able to be modified, except through reflection.

PR #888, opened 2020-07-04
add size hint variable for PublishBatch creation

bollhals:
## Proposed Changes

add a new method with a size hint for the creation of the list to hold the commands. This can reduce the allocations due to the storage of the commands by half. 

## Types of Changes

## Checklist

## Further Comments

This only adds a new overloaded method to the public IModel interface.

### DIFF
--- a/projects/RabbitMQ.Client/client/api/IModel.cs
+++ b/projects/RabbitMQ.Client/client/api/IModel.cs
@@ -292,6 +292,12 @@ namespace RabbitMQ.Client
         IBasicPublishBatch CreateBasicPublishBatch();
 
         /// <summary>
+        ///  Creates a BasicPublishBatch instance
+        /// </summary>
+        [AmqpMethodDoNotImplement(null)]
+        IBasicPublishBatch CreateBasicPublishBatch(int sizeHint);
+
+        /// <summary>
         /// Construct a completely empty content header for use with the Basic content class.
         /// </summary>
         [AmqpContentHeaderFactory("basic")]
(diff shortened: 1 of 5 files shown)


bollhals:
In the test app of stebet, before:
image

after
image

michaelklishin:
I updated `APIApproval` expectations. Thank you for your ongoing contributions @bollhals! 🥇

michaelklishin:
Backported to `6.x`.

bording:
@michaelklishin This is a breaking change, so it should not have been backported to the `6.x` branch.

stebet:
Is there also a good reason to have interfaces for the basics like IConnection, IModel etc.? Don't really see a reason why anyone would go through the trouble of implementing their own version of those.

bording:
> Is there also a good reason to have interfaces for the basics like IConnection, IModel etc.? Don't really see a reason why anyone would go through the trouble of implementing their own version of those.

When I see interfaces used like that, it's usually an attempt to make testing easier since you can provide fake testing implementations.

michaelklishin:
I'm not going to lie it is seriously annoying to me that we cannot ship new small features like this quickly in this client because of the adopted semantic versioning interpretation.

There is so much code out there that uses interfaces that removing them seems unrealistic.

michaelklishin:
I suggest then that we only release new majors and patch releases, no minors. Minors make no sense in the world of strict semantic versioning: even small features often require some kind of interface changes, even minor usability improvements can be interpreted as breaking behavior changes. If almost no change is appropriate for a minor release, why bother with them at all?

rwkarg:
> Is there also a good reason to have interfaces for the basics like IConnection, IModel etc.? Don't really see a reason why anyone would go through the trouble of implementing their own version of those.

For context, we have a SharedPublishConnection and a SharedConsumerConnection implementation of IConnection that we use to enforce that publish and subscribe happen on separate connections/sockets (to avoid TCP backpressure on consumers in case of backlogs) as well as to enforce a single publisher and single consumer connection per endpoint/connection config (to avoid unnecessary TCP connections).

PR #893, opened 2020-07-06
Do DNS resolution before connection attempt

AStove:
## Proposed Changes

When starting an application (with all exceptions on) rabbitmq always gives a couple of exceptions because it first tries to connect to IPv6 (and fails) but then succeeds to connect to IPv4. RabbitMQ is using exceptions as normal control flow. This is annoying. It makes the user wonder what he did wrong, and has to wressle through the startup of his program clicking away all the exceptions. Usually one will disable the recurring exceptions at the risk of missing it when it's relevant.

As asked in https://groups.google.com/g/rabbitmq-users/c/sHeiZMBwfWU 
Also see https://stackoverflow.com/a/59495708/1560347 .

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue)

## Checklist

## Further Comments

This pull request probably need some work. This is more of a proof of concept, please let me know what you think.

It's probably a big no-no to change the api, however it's impossible to pass an endpoint without doing the DNS resolution internally.


projects/RabbitMQ.Client/client/impl/SocketFrameHandler.cs
```diff
@@ -101,21 +101,38 @@ class SocketFrameHandler : IFrameHandler
             _channelReader = channel.Reader;
             _channelWriter = channel.Writer;
 
-            if (ShouldTryIPv6(endpoint))
+            // Resolve the hostname to know if it's even possible to even try IPv6
+            IPAddress[] adds = Dns.GetHostAddresses(endpoint.HostName);
```
michaelklishin, line 105: This is a great opportunity to introduce a resolver interface similar to [that in the Java client](https://github.com/rabbitmq/rabbitmq-java-client/blob/master/src/main/java/com/rabbitmq/client/AddressResolver.java). This would allow the user to enforce, say, IPv6 connectivity.

AStove: This goes beyond my capabilities.

michaelklishin: Fair enough.

stebet:
Just one thing I was wondering, is there a reason the RabbitMQ client doesn't just let the operating system decide if it should connect via IPv4 or IPv6 when it's given a hostname?

AStove:
Well most systems are dual stack now so the OS has the capabilities to connect using IPv6 and IPv4. However when someone sets up a server they will still only configure IPv4 entries in the DNS. The OS has no way of knowing what DNS entries there will be until it resolves the hostname. And even then the server can be up on IPv4 and not on IPv6 or vice versa, which the OS can also not know.

michaelklishin:
@stebet unfortunately, the client cannot know what address family should be used or preferred. Local OS cannot know that either since target nodes can be configured to only listen for connections that use IPv6. Trying both families is the best default behavior.

michaelklishin:
This changes an interface so can only go into `7.0`.

PR #897, opened 2020-07-07
Missing ConfigureAwait in TcpClientAdapter

danielmarbach:
## Types of Changes

## Checklist

## Further Comments

@bording I also saw this

```
#if NET461
            await Task.Run(() => _sock.Connect(ep, port)).ConfigureAwait(false);
#else
            await _sock.ConnectAsync(ep, port).ConfigureAwait(false);
#endif
```

any reasons why we are not using

```
#if NETFRAMEWORK
            await Task.Run(() => _sock.Connect(ep, port)).ConfigureAwait(false);
#else
            await _sock.ConnectAsync(ep, port).ConfigureAwait(false);
#endif
```

and then what I also wondered is if the else case ever exists? I can't find that method overload. Judging by the history it started with

https://github.com/rabbitmq/rabbitmq-dotnet-client/blame/696ed9be4e337f8472b32f85bd95bc46a48ca781/projects/client/RabbitMQ.Client/src/client/impl/TcpClientAdapter.cs#L35-L39

Any ideas? Something seems off

### DIFF
--- a/projects/RabbitMQ.Client/client/impl/TcpClientAdapter.cs
+++ b/projects/RabbitMQ.Client/client/impl/TcpClientAdapter.cs
@@ -27,7 +27,7 @@ namespace RabbitMQ.Client.Impl
                 throw new ArgumentException($"No ip address could be resolved for {host}");
             }
 
-            await ConnectAsync(ep, port);
+            await ConnectAsync(ep, port).ConfigureAwait(false);
         }
 
         public virtual async Task ConnectAsync(IPAddress ep, int port)


bording:
@danielmarbach The else case is an extension method from the `SocketTaskExtensions` class which is in `net471`+. That's why the ifdef is currently targeting `net461`. If the target framework is changed/updated, the new method will be used even if the ifdef was overlooked.

projects/RabbitMQ.Client/client/impl/TcpClientAdapter.cs
```diff
@@ -27,7 +27,7 @@ public virtual async Task ConnectAsync(string host, int port)
                 throw new ArgumentException($"No ip address could be resolved for {host}");
             }
 
-            await ConnectAsync(ep, port);
+            await ConnectAsync(ep, port).ConfigureAwait(false);
```
sungam3r, line 30: Is there any difference in using `ConfigureAwait(false)` at the end of the method? After all, after this line there is no code and there is no next step of the state machine. Or am I missing something?

sungam3r: OK. Also I use ca2007 in everyday work. I have to always disable its warnings for test projects.

PR #902, opened 2020-07-07
Getting rid of Command allocations

bollhals:
## Proposed Changes

Replaced the Command class with two structs for IncomingCommand and OutgoingCommand. 

## Types of Changes

## Checklist

## Further Comments
Before
image
After
image


projects/RabbitMQ.Client/client/impl/ModelBase.cs
```diff
@@ -608,7 +608,7 @@ public void HandleBasicCancelOk(string consumerTag)
                 _consumers.Remove(consumerTag);
             }
             ConsumerDispatcher.HandleBasicCancelOk(k.m_consumer, consumerTag);
-            k.HandleCommand(null); // release the continuation.
+            k.HandleCommand(IncomingCommand.Empty); // release the continuation.
```
danielmarbach, line 611: I don't want to imply anything needs to be changed. I'm asking for my own benefit.

Why `.Empty` instead of just `default`? My preference would have been `default` because I don't think `.Empty` conveys much especially given `.Empty` is actually `default`

projects/RabbitMQ.Client/client/impl/OutgoingCommand.cs
```diff
@@ -40,39 +40,35 @@
 
 using System;
 using System.Buffers;
-using System.Runtime.InteropServices;
 using RabbitMQ.Client.Framing.Impl;
 
 namespace RabbitMQ.Client.Impl
 {
-    class Command : IDisposable
+    internal readonly struct OutgoingCommand
```
danielmarbach, line 47:
```
Marshal.SizeOf(typeof(OutgoingCommand));
```

shows it is 32 bytes large which technically violates the principle of structs should be around 16-24 bytes. Am I assuming correctly the potential perf impact of that is largely mitigated by passing them around with `in`,`out` or `ref`?

projects/RabbitMQ.Client/client/impl/IncomingCommand.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using System.Buffers;
+
+namespace RabbitMQ.Client.Impl
+{
+    internal readonly struct IncomingCommand : IDisposable
```
danielmarbach, line 6:
```
Marshal.SizeOf(typeof(IncomingCommand));
```

shows it is 40 bytes large which technically violates the principle of structs should be around 16-24 bytes. Am I assuming correctly the potential perf impact of that is largely mitigated by passing them around with `in`,`out` or `ref`?

projects/RabbitMQ.Client/client/impl/ISession.cs
```diff
@@ -44,12 +44,14 @@
 
 namespace RabbitMQ.Client.Impl
 {
-    interface ISession
+    delegate void CommandReceivedAction(in IncomingCommand cmd);
```
danielmarbach, line 47: for consistency reasons should this have the `internal` keyword?

projects/RabbitMQ.Client/client/impl/Session.cs
```diff
@@ -43,24 +43,24 @@
...
         {
-            Command cmd = _assembler.HandleFrame(in frame);
-            if (cmd != null)
+            var cmd = _assembler.HandleFrame(in frame);
+            if (!cmd.IsEmpty)
             {
                 using (cmd)
                 {
-                    OnCommandReceived(cmd);
+                    CommandReceived.Invoke(in cmd);
```
danielmarbach, line 63: Why not

```
CommandReceived(in cmd)
```

?

danielmarbach: I would have prefered

```
            if (cmd.IsEmpty)
            {
                return;
            }

            using (cmd)
            {
                CommandReceived.Invoke(in cmd);
            }
```

danielmarbach: Couldn't this technically be `null`? Should we always assign a default delegate so that we never need to null check?

projects/RabbitMQ.Client/client/impl/SessionManager.cs
```diff
@@ -99,7 +99,7 @@ public ISession CreateInternal(int channelNumber)
         {
             lock (_sessionMap)
             {
-                ISession session = new Session(_connection, channelNumber);
+                ISession session = new Session(_connection, (ushort)channelNumber);
```
danielmarbach, line 102: Since we are now casting to ushort wouldn't it make sense to go one step further to change the `SessionManager`, `IntAllocator` and the rest of the infra as well to get rid of the cast?

michaelklishin: `IntAllocator` can be generalised to be a `ushort` allocator :)

projects/RabbitMQ.Client/client/impl/IncomingCommand.cs
bollhals: Yes, while this struct is passed around, the copying overhead is neglected by the `in` passing.

projects/RabbitMQ.Client/client/impl/OutgoingCommand.cs
bollhals: Also here, yes the `in` modifier neglects it.

projects/RabbitMQ.Client/client/impl/ISession.cs
bollhals: yes, good catch

projects/RabbitMQ.Client/client/impl/Session.cs
bollhals: Just my style, I don't mind adapting to whatever style there is preferred here.

> Couldn't this technically be `null`? Should we always assign a default delegate so that we never need to null check?

technically yes, as it's not set in the constructor, but it's done in Session.Initialise which is called from the constructor. In that sense no it can not.

projects/RabbitMQ.Client/client/impl/SessionManager.cs
bollhals: Yes that's possible. I didn't want to change too much in one go.
Also I think there is some public API that exposes it as int, so this is most likely going to stay.

projects/RabbitMQ.Client/client/impl/ModelBase.cs
bollhals: Honestly I didn't think too much about it. I do it for all "empty instances", but this one is actually a struct, so I'm uncertain about the consequences here. I'll need to investigate but my assumption is that default will (apart from the visibility) even be better overall.

bollhals: hmm interesting. My benchmark suggests that the Empty is faster here. Probably it is due to the fact of the `in` of the HandleCommand. The default case will have to create a temporary local to pass in and thus zeroing the stack for the "big" struct. Where as in the Empty case, the method just takes a reference of the static Empty field.

PR #911, opened 2020-07-10
cache the frameHeaderBuffer & rentedArray

bollhals:
## Proposed Changes

This change adds a byte[6] cache to the SocketFrameHandler instead of stack allocating every time and then calling the Read extension method that then rented another byte[] which then copied the result to the stack allocated span.
=> Simplifies & improves performance.

We might reconsider the stack allocation again if we've moved to NS2.1 (I think this contains a Stream.Read(Span) function)

Also I extended the InboundFrame with the rented array, so one does not have to retrieve it again in the dispose. 

## Types of Changes

## Checklist

## Further Comments

(The total execution duration is not relevant here, as it contains the waiting time for input in the socket. Look at the remaining functions. Before ~300ms, after ~200ms)
Performance before:
image

after:
image


michaelklishin:
Good catch, thank you for your ongoing contributions @bollhals 💯

PR #919, opened 2020-07-16
reduce the amount of times we rent / return from arraypool

### LINKED ISSUE #918: BasicGet Payload buffer is "never" returned
For "BasicDeliver" the rules got changed so that after you return from "BasicDeliver", the body reference shall not be used anymore from the outside. As far as I can tell this is due to the buffer being returned.

For BasicGet & BasicReturn the same actually applies as well. But instead of the library code doing it, it's the users responsiblity as once we returned the class, the library lost all control over it.
Are users aware of this? Would we need to make it more obvious (E.g. by implementing IDisposable on it?)

bollhals:
## Proposed Changes

We used to copy around the byte buffers multiple times until they were finally consumed. This PR changes it so they're forwarded whenever possible. 
This doesn't mean we allocate less, just that we less different memory blocks until we're done with an operation.

This PR also fixes #918 and does some minor cleanup here and there.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #918 )

## Checklist

## Further Comments

In order to take over the buffer from Incoming Command, I had to extend the Apigen, I'm not sure whether this is done as you'd want it, but also do we get rid of it now or not? :)


bollhals:
I'll update the post when I get around to compare the perf measurements between this and master. (Should be marginally faster due to not needing to copy around buffers multiple times.)

bollhals:
Before:
image

After:
image

Relevant bits are highlighted for comparison.
1. ParseBodyFrame: If the whole body is within 1 payload (I'd argue the usual case) then we don't rent / copy a new buffer, we  pass on the existing buffer.
2. HandleBasicDeliver: Instead of renting a new one and copying it over, we pass it along.
3. IncomingCommand.Dispose: No need to dispose / return anymore, since we always pass the buffer along if there was a body at all.
4. InboundFrame.Dispose: Only the body of the frames that haven't been taken over are now returned. (Before 25k+145k = 170k, After 15k+105k = 120k => missing 50k are "BasicDeliver payload frames"

bollhals:
Failing test is the breaking public API (Adding the IDisposable).

michaelklishin:
@bollhals please update the `APIApproved.txt` file with the generated one as this is a change we intend to make. Thank you.

bollhals:
> @bollhals please update the `APIApproved.txt` file with the generated one as this is a change we intend to make. Thank you.

/fixed

michaelklishin:
Thank you!

PR #927, opened 2020-08-18
change AmqpVersion to struct

bollhals:
## Proposed Changes

Simple small change to change it to a struct. Avoids allocation.

## Types of Changes

## Checklist

## Further Comments

I've let resharper implement the equality members, which also changed the GetHashCode. But as the type is internal, this is not breaking.

### DIFF
--- a/projects/RabbitMQ.Client/client/impl/AmqpVersion.cs
+++ b/projects/RabbitMQ.Client/client/impl/AmqpVersion.cs
@@ -29,6 +29,8 @@
 //  Copyright (c) 2007-2020 VMware, Inc.  All rights reserved.
 //---------------------------------------------------------------------------
 
+using System;
+
 namespace RabbitMQ.Client.Framing.Impl
 {
     /// <summary>Represents a version of the AMQP specification.</summary>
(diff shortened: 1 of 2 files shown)


projects/RabbitMQ.Client/client/impl/AmqpVersion.cs
```diff
@@ -70,27 +72,36 @@ public AmqpVersion(int major, int minor)
...
         /// <summary>
         /// Implement hashing as for value-equality.
         /// </summary>
         public override int GetHashCode()
         {
-            return 31*Major.GetHashCode() + Minor.GetHashCode();
+            unchecked
+            {
+                return (Major * 397) ^ Minor;
+            }
```
sungam3r, line 104:
```suggestion
            return (Major, Minor).GetHashCode();
```

michaelklishin (approved):

projects/RabbitMQ.Client/client/impl/AmqpVersion.cs
michaelklishin: That seems to require an extra dependency on `System.Tuple_2`, though.

michaelklishin: Thank you for the suggestion, it is definitely a nicer option but I'm not sure if it's worth an extra dependency (if it does require one, I'm unable to compile this on .NET Core 3.1 on MacOS as is).

sungam3r: ok

bollhals: Due to .NET 4.6.1 this syntax isn't yet possible to use without additional dependencies. (Similar to that we can't use anonymous (int, int) return values)

PR #931, opened 2020-08-20
Code cleanups

stebet:
## Proposed Changes

Loads of code cleanups:
* Explicit types instead of var where type is not apparent.
* Removed private setters from properties where they were not used.
* Simplified null-checks to `x is null` instead of `x == null`. Also prevents accidental bugs in case `==` operator would ever be overloaded.
* Made private variables `readonly` where possible.
* Simplified Properties and Methods to be expression-based where possible.
* Made null-checks implit via property where it made sense to reduce code clutter.
* Changed multiple `if`s to `switch` where possible.

## Types of Changes

## Checklist

### DIFF
--- a/projects/RabbitMQ.Client/util/DebugUtil.cs
+++ b/projects/RabbitMQ.Client/util/DebugUtil.cs
@@ -99,55 +99,62 @@ namespace RabbitMQ.Util
         ///<summary>Dump properties of objects to the supplied writer.</summary>
         public static void DumpProperties(object value, TextWriter writer, int indent)
         {
-            if (value == null)
+            switch (value)
             {
-                writer.WriteLine("(null)");
-            }
-            else if (value is string)
-            {
-                writer.WriteLine($"\"{((string)value).Replace("\"", "\\\"")}\"");
-            }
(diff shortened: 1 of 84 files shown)


stebet:
For an example of the most drastic changes and code reductions, see projects/RabbitMQ.Client/client/impl/AutorecoveringModel.cs for example.

projects/RabbitMQ.Client/util/DebugUtil.cs
```diff
@@ -99,55 +99,62 @@ public static void DumpKeyValue(string key, object value, TextWriter writer, int
         ///<summary>Dump properties of objects to the supplied writer.</summary>
         public static void DumpProperties(object value, TextWriter writer, int indent)
         {
-            if (value == null)
+            switch (value)
```
danielmarbach, line 102: Should we lock the csharp version to not accidentally use something that would not be compatible?

michaelklishin:
It would be very nice to have a version of this for `7.x`: this makes the delta with master noticeably wider and complicates backporting of subsequent PRs.

PR #935, opened 2020-08-26
Expose events ConsumerTagChangeAfterRecovery and QueueNameChangeAfterRecovery

paulomf:
## Proposed Changes

After the merge #490 the events ConsumerTagChangeAfterRecovery and QueueNameChangeAfterRecovery became inaccessible. I propose to declare those events in the interface IAutorecoveringConnection.

## Types of Changes

## Checklist

### DIFF
--- a/projects/RabbitMQ.Client/client/api/IAutorecoveringConnection.cs
+++ b/projects/RabbitMQ.Client/client/api/IAutorecoveringConnection.cs
@@ -46,5 +46,8 @@ namespace RabbitMQ.Client
     {
         event EventHandler<EventArgs> RecoverySucceeded;
         event EventHandler<ConnectionRecoveryErrorEventArgs> ConnectionRecoveryError;
+
+        event EventHandler<ConsumerTagChangedAfterRecoveryEventArgs> ConsumerTagChangeAfterRecovery;
+        event EventHandler<QueueNameChangedAfterRecoveryEventArgs> QueueNameChangeAfterRecovery;
     }
 }


michaelklishin:
@paulomf why do you need to observe those events? They are very much internal to the operation of the recovery mechanism.

paulomf:
@michaelklishin When using a server named queue, a new queue name may get generated after an automatic network recovery. Without observing such an event, I'm unable to retrieve the new name.

Unfortunately, the automatic recovery may not be able to recover the queue when using a constant queue name if the client starts a new connection without the server clearing the previous one and it's correlated exclusive queue, which is why I'm using the server-named queue.

michaelklishin (approved):

michaelklishin:
This could be backported for `6.3.0`, let's see what others say and how far off `7.0` may be.

Dommicentl:
Any chance that this gets backported to `6.X`? It is the only missing functionality that keeps us from upgrading.

michaelklishin:
@Dommicentl from upgrading to what version? This change has never shipped in a release.

Dommicentl:
We are currently on the latest `5.X` version. But we would like to upgrade to the latest `6.X`. However we also use the QueueNameChangeAfterRecovery event which got removed in the `6.0` release. So we need this change, which again exposes this event, before we are able to upgrade. That is why I wondered if it could be backported to the `6.X` branch.

michaelklishin:
Backported to `7.x` and `6.x`.

Dommicentl:
Super, thanks! Is there a new release planned of the `6.X` branch in the near future?

Dommicentl:
@michaelklishin, any update on a release that will contain this fix?

pdoteter:
> Backported to `7.x` and `6.x`.

Nice, I was looking in some issue when I came across this thread, should solved the same issue we have.  Any word when the new 6.x branch with this fix will be released?

Issue #515, opened 2019-02-12
Consider removing implementations for unused events in Connection


antiduh:
Split from #512 .

The `Connection` class exposes events that it never fires - `RecoverySucceeded` and `ConnectionRecoveryError`. It seems these events are defined in `Connection` because they're part of the `IConnection` interface, which seems to define these events on behalf of `AutorecoveringConnection`.

As an example, here are all of the callsites to the `RecoverySucceeded` event and its backing field:

```
        public event EventHandler<EventArgs> RecoverySucceeded
        {
            add
            {
                lock (m_eventLock)
                {
                    m_recoverySucceeded += value;
                }
            }
            remove
            {
                lock (m_eventLock)
                {
                    m_recoverySucceeded -= value;
                }
            }
        }


        private void Dispose(bool disposing)
        {
           ...
                    m_recoverySucceeded = null;
           ...
         }
```

Since the events are never fired from within `Connection`, but they need to be implemented by `Connection` to meet the `IConnection` interface, the events should be replaced with empty add-remove handlers, and all backing fields should be removed.

```
        public event EventHandler<EventArgs> RecoverySucceeded
        {
            add { }
            remove { }
        }

        public event EventHandler<ConnectionRecoveryErrorEventArgs> ConnectionRecoveryError
        {
            add { }
            remove {}
        }
```

michaelklishin:
We can consider extracting an `IAutorecoverableConnection`, it sounds like a natural thing to do to me.

antiduh:
That was an option I was thinking of, but I didn't want to overstep my bounds; I was aiming for fixes that are low-impact / non-breaking, but I'm ok with separating the concerns into different interfaces.

michaelklishin:
@antiduh that's fine and appreciated. Larger changes can always follow later.

michaelklishin:
We will undo this in https://github.com/rabbitmq/rabbitmq-dotnet-client/issues/940 for `7.0`. Using empty handlers and providing `IConnection` users a way to register recovery events makes more sense. Casting to `IAutorecoveringConnection` is not intuitive at all.

PR #943, opened 2020-09-17
Breaking public API change: fold IAutorecoveringConnection [back] into IConnection

### LINKED ISSUE #940: Discussion: recovery events for automatically recovering connections (IAutoRecoveringConnection)
Reposting from https://groups.google.com/g/rabbitmq-users/c/k9eHUaptlRg

In the move from v5.0.0 to v6.0.0 of the .NET client, `IAutoRecoveringConnection` was created to hold connection recovery events in https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/758. However it isn't possible to get hold of one via the public API, as `ConnectionFactory.CreateConnection()` returns an `IConnection`. 

Is it worth considering a change to the public API to allow for returning an `IAutoRecoveringConnection`? I can see two ways to do this:

- Non breaking: create another method `ConnectionFactory.CreateAutoRecoveringConnection()` which returns an `IAutoRecoveringConnection`. This would ignore whatever value `ConnectionFactory.AutomaticRecoveryEnabled` is set to, however this could be confusing to API users.

- Breaking: Remove `ConnectionFactory.AutomaticRecoveryEnabled`, have `ConnectionFactory.CreateConnection()` create a `Connection` and return an `IConnection`, and add `ConnectionFactory.CreateAutoRecoveringConnection()` which creates an `AutoRecoveringConnection` and returns an `IAutoRecoveringConnection`.

I'm not sure what would be preferred. Of course there may be a better solution that I haven't thought of.

At the moment we can work around this by checking if our `IConnection` object is actually an `IRecoveringConnection`, casting it and subscribing to the events, but I think it would be better to provide a proper way of getting an `IRecoveringConnection`.

michaelklishin:
This undoes #758 (introduced in response to #515).

The events make more sense in the primary IConnection interface.
Most users enable automatic connection recovery and asking them to
cast to IAutorecoveringConnection is counter-intuitive.

As suggested in #515, we add no-op add and remove handlers
so that IConnection is easier to implement/mock.

See #940 for the discussion.

Closes #940, expected to go into `7.0` only.


lukebakken (approved):

michaelklishin:
Merging because we seem to have reached a consensus in #940: this is a better way forward compared to #758.

Issue #944, opened 2020-09-24
v6.2.1: CallbackException is not invoked for unhandled exceptions from AsyncEventingBasicConsumer.Received


qrli:
rabbitmq-dotnet-client v6.2.1
rabbitmq server v3.8.3

When an unhandled exception thrown from within `AsyncEventingBasicConsumer.Received`, IModel.CallbackException will not be triggered in v6.2.1. It had always been working up to v6.1.0. 

I see you have [unit tests for this scenario for the sync BasicConsumer](https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/275b4ebcc6f00df414612703ea0c3e367f233c26/projects/Unit/TestConsumerExceptions.cs), but not for the Async consumer. So the issue may be only for the async API.

michaelklishin:
A contribution would be welcome. How unhandled exceptions are treated by async APIs in general can be very different from "regular" code.

qrli:
I dug in the code base for sometime, searching for clues. It is my first time reading it, so still quite unfamilar.
I kind of believe I found the cause in [AsyncConsumerWorkService.cs](https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/d9c6f478511f448326e02d33aa1e6f294b8a0036/projects/RabbitMQ.Client/client/impl/AsyncConsumerWorkService.cs):
```cs
                        try
                        {
                            Task task = work.Execute();
                            if (!task.IsCompleted)
                            {
                                await task.ConfigureAwait(false);
                            }
                        }
                        catch (Exception e)
```
This code tries to optimize in case Task has already finished. However, `IsCompleted` does not mean success. It can be failure with an exception. In case the handler runs fast and throws, the above logic just eats the exception inside `Task.Exception` silently.

In addition, if the task is not `await`ed, the exception will not be unwrapped, going back to the early days issues of AggregatedException parsing. 

To my knowledge, `await` has some builtin optimization in case the task has already completed. I'm not exactly sure how it compares with the manual optimization above, but from correctness point of view, I think it is easier and more robust to simply `await` the task than trying to micro-optimize here.

michaelklishin:
Addressed in #946 contributed by @BarShavit 👏

PR #946, opened 2020-09-25
Make sure OnCallbackException is executed for AsyncConsumers

BarShavit:
## Proposed Changes

Always call to await when executing a task in AsyncConsumerWorkService. Previously, the wait called only if the job hasn't done yet. It lead to missing exception and missed the invoke of OnCallbackException event.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #944 )

## Checklist

### DIFF
--- a/projects/RabbitMQ.Client/client/impl/AsyncConsumerWorkService.cs
+++ b/projects/RabbitMQ.Client/client/impl/AsyncConsumerWorkService.cs
@@ -90,10 +90,7 @@ namespace RabbitMQ.Client.Impl
                         try
                         {
                             Task task = work.Execute();
-                            if (!task.IsCompleted)
-                            {
-                                await task.ConfigureAwait(false);
-                            }
+                            await task.ConfigureAwait(false);
                         }
                         catch (Exception e)
                         {
@@ -126,7 +123,7 @@ namespace RabbitMQ.Client.Impl
                         while (_channel.Reader.TryRead(out Work work))
                         {
                             // Do a quick synchronous check before we resort to async/await with the state-machine overhead.
-                            if(!_limiter.Wait(0))
+                            if (!_limiter.Wait(0))
                             {
                                 await _limiter.WaitAsync(cancellationToken).ConfigureAwait(false);
                             }
@@ -146,10 +143,7 @@ namespace RabbitMQ.Client.Impl
                 try
                 {
                     Task task = work.Execute();
-                    if (!task.IsCompleted)
-                    {
-                        await task.ConfigureAwait(false);
-                    }
+                    await task.ConfigureAwait(false);
                 }
                 catch (Exception e)
                 {
--- a/projects/RabbitMQ.Client/client/impl/AsyncConsumerWorkService.cs
+++ b/projects/RabbitMQ.Client/client/impl/AsyncConsumerWorkService.cs
@@ -90,7 +90,16 @@ namespace RabbitMQ.Client.Impl
                         try
                         {
                             Task task = work.Execute();
-                            await task.ConfigureAwait(false);
+                            if (!task.IsCompleted)
+                            {
+                                await task.ConfigureAwait(false);
+                            }
+
+                            // Without await, the exception will disappear
+                            if(task.Exception != null)
+                            {
+                                throw task.Exception.InnerException;
+                            }
                         }
                         catch (Exception e)
                         {
@@ -143,7 +152,16 @@ namespace RabbitMQ.Client.Impl
                 try
                 {
                     Task task = work.Execute();
-                    await task.ConfigureAwait(false);
+                    if (!task.IsCompleted)
+                    {
+                        await task.ConfigureAwait(false);
+                    }
+
+                    // Without await, the exception will disappear
+                    if (task.Exception != null)
+                    {
+                        throw task.Exception.InnerException;
+                    }
                 }
                 catch (Exception e)
                 {
--- a/projects/RabbitMQ.Client/client/impl/AsyncConsumerWorkService.cs
+++ b/projects/RabbitMQ.Client/client/impl/AsyncConsumerWorkService.cs
@@ -94,11 +94,10 @@ namespace RabbitMQ.Client.Impl
                             {
                                 await task.ConfigureAwait(false);
                             }
-
-                            // Without await, the exception will disappear
-                            if(task.Exception != null)
+                            else
                             {
-                                throw task.Exception.InnerException;
+                                // to materialize exceptions if any
+                                task.GetAwaiter().GetResult();
                             }
                         }
                         catch (Exception e)
@@ -156,11 +155,10 @@ namespace RabbitMQ.Client.Impl
                     {
                         await task.ConfigureAwait(false);
                     }
-
-                    // Without await, the exception will disappear
-                    if (task.Exception != null)
+                    else
                     {
-                        throw task.Exception.InnerException;
+                        // to materialize exceptions if any
+                        task.GetAwaiter().GetResult();
                     }
                 }
                 catch (Exception e)
--- a/projects/Unit/TestAsyncConsumerExceptions.cs
+++ b/projects/Unit/TestAsyncConsumerExceptions.cs
@@ -114,7 +114,7 @@ namespace RabbitMQ.Client.Unit
             {
             }
 
-            public override async Task HandleBasicDeliver(string consumerTag,
+            public override Task HandleBasicDeliver(string consumerTag,
                 ulong deliveryTag,
                 bool redelivered,
                 string exchange,
@@ -122,8 +122,7 @@ namespace RabbitMQ.Client.Unit
                 IBasicProperties properties,
                 ReadOnlyMemory<byte> body)
             {
-                await Task.Delay(0);
-                throw new Exception("oops");
+                return Task.FromException(new Exception("oops"));
             }
         }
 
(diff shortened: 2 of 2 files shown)


BarShavit:
Related to issue #944

projects/RabbitMQ.Client/client/impl/AsyncConsumerWorkService.cs
```diff
@@ -90,10 +90,7 @@ private async Task Loop()
                         try
                         {
                             Task task = work.Execute();
-                            if (!task.IsCompleted)
```
danielmarbach, line 93: How about doing

```
if (!task.IsCompleted)
{
    await task.ConfigureAwait(false);
}
else
{
    // to materialize exceptions if any
    task.GetAwaiter().GetResult();
}
```

to not undo @stebet optimizations he was trying to achieve here?

projects/RabbitMQ.Client/client/impl/AsyncConsumerWorkService.cs
```diff
@@ -146,10 +143,7 @@ private static async Task HandleConcurrent(Work work, IModel model, SemaphoreSli
                 try
                 {
                     Task task = work.Execute();
-                    if (!task.IsCompleted)
```
danielmarbach, line 149: Same comment as above

https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/946/files#r495070886

projects/RabbitMQ.Client/client/impl/AsyncConsumerWorkService.cs
danielmarbach: Btw. I usually prefer to always use the async state machinery to avoid the problems we are seeing here but I totally respect the previous attempts to optimize this code since it is living on the hot path and the optimization was targetting synchronously completing workers

BarShavit: The await command should works just the same.
Another solution is check if task.Exception exists, and then throw a new exception.
It will make it more readable.

danielmarbach: Readability should not be the reason the change the stack trace at all. Plus the task exception is an AggregateException. My proposal properly unpacks the exception should we want to keep the current synchronously completed task optimization

projects/RabbitMQ.Client/client/impl/AsyncConsumerWorkService.cs
```diff
@@ -90,7 +90,16 @@ private async Task Loop()
...
                             Task task = work.Execute();
-                            await task.ConfigureAwait(false);
+                            if (!task.IsCompleted)
+                            {
+                                await task.ConfigureAwait(false);
+                            }
+
+                            // Without await, the exception will disappear
+                            if(task.Exception != null)
+                            {
```
danielmarbach, line 100: This rewrites the stack trace

projects/Unit/TestAsyncConsumerExceptions.cs
```diff
@@ -0,0 +1,165 @@
...
+            }
+        }
+
+        protected void TestExceptionHandlingWith(IBasicConsumer consumer,
+            Action<IModel, string, IBasicConsumer, string> action)
+        {
+            object o = new object();
+            bool notified = false;
+            string q = _model.QueueDeclare();
+
```
danielmarbach, line 115: Nit 🔥 extra line

projects/Unit/TestAsyncConsumerExceptions.cs
```diff
@@ -0,0 +1,165 @@
...
+using System.Threading.Tasks;
+using NUnit.Framework;
+using RabbitMQ.Client.Events;
+
+namespace RabbitMQ.Client.Unit
+{
+    [TestFixture]
+    public class TestAsyncConsumerExceptions : IntegrationFixture
+    {
+        private class ConsumerFailingOnDelivery : AsyncEventingBasicConsumer
```
danielmarbach, line 43: I would prefer these below the tests but I leave this up to @michaelklishin to decide what's best

danielmarbach (approved):

bollhals:
Are you sure the tests are testing the correct thing?
(They are not very strict about what they verify the success of. Just that a exception was thrown.)

If I debug them, I get the exception thrown by `throw new InvalidOperationException("Should never be called.");` from AsyncDefaultBasicConsumer.cs

Also if I use the tests on master, they're green with or without your changes.

danielmarbach:
@bollhals good catch. Honestly I was a bit lazy in reviewing the tests and mostly looked at the implementation side of things. Will take extra care next time (although I confess I was just about going on leave yesterday and I think that cause me to take shortcuts)

BarShavit:
Copied the tests from the sync version.
Fixed that, now it fails without the changes.
The bug happened only if the registered received method is async. Also need to mark "DispatchConsumersAsync" as true, or it go to the default class and throws the exception `throw new InvalidOperationException("Should never be called.");`.

The task.delay(0) is just to avoid warnings (async without await in the method).

projects/Unit/TestAsyncConsumerExceptions.cs
```diff
@@ -40,20 +40,89 @@ namespace RabbitMQ.Client.Unit
...
-            public override Task HandleBasicDeliver(string consumerTag,
+            public override async Task HandleBasicDeliver(string consumerTag,
                 ulong deliveryTag,
                 bool redelivered,
                 string exchange,
                 string routingKey,
                 IBasicProperties properties,
                 ReadOnlyMemory<byte> body)
             {
+                await Task.Delay(0);
```
danielmarbach, line 125: I find this hard to understand because the thing you intend to test is the throwing on the synchronous path, but then there is still the async statemachine at play and the reader if the test need to now know that Task Delay with zero always returns a completed task and therefore sync completes. Still though the presence of the async key work essentially changes the exception semantics at play

danielmarbach: https://dotnetfiddle.net/1DUM2W

You can observe the difference by executing the above fiddle. Once as is and once with the async keyword on Throws

BarShavit: The purpose of those tests is to reproduce the bug #944. For cases which the async change the exception semantics.
This line needed only to clean the compile warning, not changing the functionality. The other option was suppress  the warning, which lead to more confusing then this.
Or calling another random async function and await, but still - it's a random action just to remove the warning.

danielmarbach: My point was any async / await code that yields you add changes the behaviour of the code. The description of the issue you reference talks about sync completion which is the case when IsCompleted is true right after scheduling. It is almost impossible to run with yielding into the sync completion case there you would always land on the path that already today awaits the task. Or in other words are you sure the test you wrote are correctly testing the else part?

danielmarbach: You could use Task.FromException instead if I'm not mistaken 
https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.fromexception?view=netcore-3.1

BarShavit: Yes, it tests what it should be. The test classes throws exception without waiting, so with an average CPU it will get task done flag without a problem.
Changed to Task.FromException anyway, it's more elegant way.

danielmarbach (approved):

bollhals:
I did some quick tests regarding performance of this change

|          Method |     Mean |    Error |   StdDev |   Median | Ratio |
|---------------- |---------:|---------:|---------:|---------:|------:|
|     IsCompleted | 23.24 ns | 0.138 ns | 0.202 ns | 23.24 ns |  1.00 |
|           Await | 27.88 ns | 0.055 ns | 0.080 ns | 27.86 ns |  1.20 |
|       GetResult | 23.62 ns | 0.199 ns | 0.297 ns | 23.47 ns |  1.02 |
| RanToCompletion | 25.08 ns | 0.072 ns | 0.106 ns | 25.08 ns |  1.08 |

So the currently suggested change seems like the most promising one!

projects/Unit/TestAsyncConsumerExceptions.cs
```diff
@@ -0,0 +1,177 @@
...
+        {
+            object o = new object();
+            bool notified = false;
+            string q = _model.QueueDeclare();
+
+
+            _model.CallbackException += (m, evt) =>
+            {
+                notified = true;
+                Monitor.PulseAll(o);
```
bollhals, line 54: Due to the sync call, the pulse will always happen before the wait, hence it always runs in the test timeout, but since its not checked in WaitOn-Method, it still passes but the test takes 5s.

My suggestion: Use a ResetEvent or a TaskCompletionSource to notify. (If you use a ResetEvent, you still have the timeout available)

projects/Unit/TestAsyncConsumerExceptions.cs
```diff
@@ -0,0 +1,177 @@
...
+            Action<IModel, string, IBasicConsumer, string> action)
+        {
+            object o = new object();
+            bool notified = false;
+            string q = _model.QueueDeclare();
+
+
+            _model.CallbackException += (m, evt) =>
+            {
+                notified = true;
```
bollhals, line 53: My personal preference: Verify the exception raised against an expected one to be sure we got what we want. And only then set notified to true

michaelklishin:
Thank you for sticking with it @BarShavit!

bollhals:
This one can and should probably go back into 6.x

michaelklishin (approved):

michaelklishin:
Backported to `6.x`.

PR #949, opened 2020-10-13
Add System.IO.Pipelines to be able to better support Async

stebet:
## Proposed Changes

This PR adds [System.IO.Pipelines](https://devblogs.microsoft.com/dotnet/system-io-pipelines-high-performance-io-in-net/) and replaces the Stream+Channels implementation in the SocketFrameHandler. It does so via. Pipelines.Socket.Unofficial which is written and maintained by @mgravell from Stack Overflow and used for example in their StackExchange.Redis library.

No public facing API changes.

This is a good preparation to be able to add Async support since we can now do the network IO completely asynchronously.

Notable changes:
* Removes the channels in the SocketFrameHandler and the custom Task needed to handle writes.
* Reintroduces locking since we need to make sure only one thread at a time fetches a memory buffer to write network data into and flush. Locks are only needed in two places though, when sending heartbeats and when writing commands.
* Removes the need to manage and return memory buffers for data written to the wire, since pipelines can provide a buffer to serialize data into.
* Makes reading data from the network completely asynchronous!
* Can make writing data completely asynchronous as well (although it's blocking currently due to the API being blocking, this will allow us to remove that blocker.
* Less memory copies and buffering needed overall. This is not a significant impact on memory usage overall since we were reusing buffers anyways but should result in less manual buffer manipulation and hopefully reduced CPU usage as well.

## Types of Changes

- [X] Bug fix (non-breaking change which fixes issue #694 and prepares for #843)

## Checklist

### DIFF
--- a/projects/RabbitMQ.Client/client/api/ConnectionFactoryBase.cs
+++ b/projects/RabbitMQ.Client/client/api/ConnectionFactoryBase.cs
@@ -31,6 +31,9 @@
 
 using System;
 using System.Net.Sockets;
+
+using Pipelines.Sockets.Unofficial;
+
 using RabbitMQ.Client.Impl;
 
 namespace RabbitMQ.Client
@@ -49,12 +52,8 @@ namespace RabbitMQ.Client
         /// <returns>New instance of a <see cref="TcpClient"/>.</returns>
         public static ITcpClient DefaultSocketFactory(AddressFamily addressFamily)
         {
-            var socket = new Socket(addressFamily, SocketType.Stream, ProtocolType.Tcp)
-            {
-                NoDelay = true,
-                ReceiveBufferSize = 65536,
-                SendBufferSize = 65536
-            };
+            var socket = new Socket(addressFamily, SocketType.Stream, ProtocolType.Tcp);
+            SocketConnection.SetRecommendedClientOptions(socket);
             return new TcpClientAdapter(socket);
         }
     }
--- a/projects/RabbitMQ.Client/client/impl/Connection.cs
+++ b/projects/RabbitMQ.Client/client/impl/Connection.cs
@@ -32,6 +32,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.IO.Pipelines;
 using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
@@ -58,6 +59,8 @@ namespace RabbitMQ.Client.Framing.Impl
         private ShutdownEventArgs? _closeReason;
         public ShutdownEventArgs? CloseReason => Volatile.Read(ref _closeReason);
 
+        private readonly SemaphoreSlim _writeSemaphore = new SemaphoreSlim(1);
+
         public Connection(ConnectionConfig config, IFrameHandler frameHandler)
         {
             _config = config;
@@ -71,7 +74,7 @@ namespace RabbitMQ.Client.Framing.Impl
 
             _sessionManager = new SessionManager(this, 0);
             _session0 = new MainSession(this);
-            _model0 = new Model(_config, _session0); ;
+            _model0 = new Model(_config, _session0);
 
             ClientProperties = new Dictionary<string, object?>(_config.ClientProperties)
             {
@@ -79,7 +82,7 @@ namespace RabbitMQ.Client.Framing.Impl
                 ["connection_name"] = ClientProvidedName
             };
 
-            _mainLoopTask = Task.Factory.StartNew(MainLoop, TaskCreationOptions.LongRunning);
+            _mainLoopTask = Task.Run(MainLoop);
             try
             {
                 Open();
@@ -373,7 +376,7 @@ namespace RabbitMQ.Client.Framing.Impl
 
         private bool SetCloseReason(ShutdownEventArgs reason)
         {
-            return System.Threading.Interlocked.CompareExchange(ref _closeReason, reason, null) is null;
+            return Interlocked.CompareExchange(ref _closeReason, reason, null) is null;
         }
 
         private void LogCloseError(string error, Exception ex)
@@ -394,9 +397,135 @@ namespace RabbitMQ.Client.Framing.Impl
             _callbackExceptionWrapper.Invoke(this, args);
         }
 
-        internal void Write(ReadOnlyMemory<byte> memory)
+        internal void Write(ReadOnlySpan<byte> payload)
         {
-            _frameHandler.Write(memory);
+            if (!_frameHandler.IsClosed)
+            {
+                _writeSemaphore.Wait();
+                WriteSpan(payload);
+                FlushAndRelease().AsTask().Wait(); // Sync-over-async, not great, not terrible. All we can do for the sync path.
+            }
+        }
+
+        internal ValueTask<FlushResult> WriteAsync(ReadOnlySpan<byte> payload)
+        {
+            if (!_frameHandler.IsClosed)
+            {
+                if (!_writeSemaphore.Wait(0))
+                {
+                    return WriteAsyncSlow(payload.ToArray());
+                }
+
+                WriteSpan(payload);
+                return FlushAndRelease();
+            }
+
+            return new ValueTask<FlushResult>(new FlushResult(false, true));
+        }
+
+        internal async ValueTask<FlushResult> WriteAsyncSlow(ReadOnlyMemory<byte> memory)
+        {
+            await _writeSemaphore.WaitAsync().ConfigureAwait(false);
+            WriteSpan(memory.Span);
+            return await FlushAndRelease().ConfigureAwait(false);
+        }
+
+        private void WriteSpan(ReadOnlySpan<byte> payload)
+        {
+            Span<byte> span = _frameHandler.FrameWriter.GetSpan(payload.Length);
+            payload.CopyTo(span);
+            _frameHandler.FrameWriter.Advance(payload.Length);
+        }
+
+        internal void Write<TMethod>(ref TMethod method, ushort channelNumber) where TMethod : struct, IOutgoingAmqpMethod
+        {
+            if (!_frameHandler.IsClosed)
+            {
+                _writeSemaphore.Wait();
+                Client.Impl.Framing.SerializeToFrames(ref method, _frameHandler.FrameWriter, channelNumber);
+                FlushAndRelease().AsTask().Wait(); // Sync-over-async, not great, not terrible. All we can do for the sync path.
+            }
+        }
+
+        internal ValueTask<FlushResult> WriteAsync<TMethod>(ref TMethod method, ushort channelNumber) where TMethod : struct, IOutgoingAmqpMethod
+        {
+            if (!_frameHandler.IsClosed)
+            {
+                if (!_writeSemaphore.Wait(0))
+                {
+                    return WriteAsyncSlow(method, channelNumber);
+                }
+
+                Client.Impl.Framing.SerializeToFrames(ref method, _frameHandler.FrameWriter, channelNumber);
+                return FlushAndRelease();
+            }
+
+            return new ValueTask<FlushResult>(new FlushResult(false, true));
+        }
+
+        private async ValueTask<FlushResult> WriteAsyncSlow<TMethod>(TMethod method, ushort channelNumber) where TMethod : struct, IOutgoingAmqpMethod
+        {
+            await _writeSemaphore.WaitAsync().ConfigureAwait(false);
+            Client.Impl.Framing.SerializeToFrames(ref method, _frameHandler.FrameWriter, channelNumber);
+            return await FlushAndRelease().ConfigureAwait(false);
+        }
+
+        internal void Write<TMethod, THeader>(ref TMethod method, ref THeader header, ReadOnlyMemory<byte> body, ushort channelNumber, int maxBodyPayloadBytes)
+            where TMethod : struct, IOutgoingAmqpMethod
+            where THeader : IAmqpHeader
+        {
+            if (!_frameHandler.IsClosed)
+            {
+                _writeSemaphore.Wait();
+                Client.Impl.Framing.SerializeToFrames(ref method, ref header, body, _frameHandler.FrameWriter, channelNumber, maxBodyPayloadBytes);
+                FlushAndRelease().AsTask().Wait(); // Sync-over-async, not great, not terrible. All we can do for the sync path.
+            }
+        }
+
+        internal ValueTask<FlushResult> WriteAsync<TMethod, THeader>(ref TMethod method, ref THeader header, ReadOnlyMemory<byte> body, ushort channelNumber, int maxBodyPayloadBytes)
+            where TMethod : struct, IOutgoingAmqpMethod
+            where THeader : IAmqpHeader
+        {
+            if (!_frameHandler.IsClosed)
+            {
+                if (!_writeSemaphore.Wait(0))
+                {
+                    return WriteAsyncSlow(method, header, body, channelNumber, maxBodyPayloadBytes);
+                }
+
+                Client.Impl.Framing.SerializeToFrames(ref method, ref header, body, _frameHandler.FrameWriter, channelNumber, maxBodyPayloadBytes);
+                return FlushAndRelease();
+            }
+
+            return new ValueTask<FlushResult>(new FlushResult(false, true));
+        }
+
+        private async ValueTask<FlushResult> WriteAsyncSlow<TMethod, THeader>(TMethod method, THeader header, ReadOnlyMemory<byte> body, ushort channelNumber, int maxBodyPayloadBytes)
+            where TMethod : struct, IOutgoingAmqpMethod
+            where THeader : IAmqpHeader
+        {
+            await _writeSemaphore.WaitAsync().ConfigureAwait(false);
+            Client.Impl.Framing.SerializeToFrames(ref method, ref header, body, _frameHandler.FrameWriter, channelNumber, maxBodyPayloadBytes);
+            return await FlushAndRelease().ConfigureAwait(false);
+        }
+
+        private ValueTask<FlushResult> FlushAndRelease()
+        {
+            async ValueTask<FlushResult> AwaitFlushAndRelease(ValueTask<FlushResult> flushTask)
+            {
+                FlushResult result = await flushTask.ConfigureAwait(false);
+                _writeSemaphore.Release();
+                return result;
+            }
+
+            ValueTask<FlushResult> flushTask = _frameHandler.FrameWriter.FlushAsync();
+            if (flushTask.IsCompletedSuccessfully)
+            {
+                _writeSemaphore.Release();
+                return flushTask;
+            }
+
+            return AwaitFlushAndRelease(flushTask);
         }
 
         public void Dispose()
--- a/projects/RabbitMQ.Client/client/impl/Frame.cs
+++ b/projects/RabbitMQ.Client/client/impl/Frame.cs
@@ -31,10 +31,7 @@
 
 using System;
 using System.Buffers;
-using System.IO;
-using System.Net.Sockets;
 using System.Runtime.CompilerServices;
-using System.Runtime.ExceptionServices;
 
 using RabbitMQ.Client.Exceptions;
 using RabbitMQ.Client.Framing.Impl;
@@ -131,41 +128,30 @@ namespace RabbitMQ.Client.Impl
             /// <summary>
             /// Compiler trick to directly refer to static data in the assembly, see here: https://github.com/dotnet/roslyn/pull/24621
             /// </summary>
-            private static ReadOnlySpan<byte> Payload => new byte[]
+            internal static ReadOnlySpan<byte> Payload => new byte[]
             {
                 Constants.FrameHeartbeat,
                 0, 0, // channel
                 0, 0, 0, 0, // payload length
                 Constants.FrameEnd
             };
-
-            public static Memory<byte> GetHeartbeatFrame()
-            {
-                // Is returned by SocketFrameHandler.WriteLoop
-                byte[] buffer = ArrayPool<byte>.Shared.Rent(FrameSize);
-                Payload.CopyTo(buffer);
-                return new Memory<byte>(buffer, 0, FrameSize);
-            }
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static ReadOnlyMemory<byte> SerializeToFrames<T>(ref T method, ushort channelNumber)
-            where T : struct, IOutgoingAmqpMethod
+        public static void SerializeToFrames<TMethod, TBufferWriter>(ref TMethod method, TBufferWriter pipeWriter, ushort channelNumber) where TMethod : struct, IOutgoingAmqpMethod where TBufferWriter : IBufferWriter<byte>
         {
             int size = Method.FrameSize + method.GetRequiredBufferSize();
-
-            // Will be returned by SocketFrameWriter.WriteLoop
-            var array = ArrayPool<byte>.Shared.Rent(size);
-            int offset = Method.WriteTo(array, channelNumber, ref method);
-
-            System.Diagnostics.Debug.Assert(offset == size, $"Serialized to wrong size, expect {size}, offset {offset}");
-            return new ReadOnlyMemory<byte>(array, 0, size);
+            Span<byte> outputBuffer = pipeWriter.GetSpan(size);
+            int offset = Method.WriteTo(outputBuffer, channelNumber, ref method);
+            System.Diagnostics.Debug.Assert(offset == size, "Serialized to wrong size", "Serialized to wrong size, expect {0:N0}, offset {1:N0}", size, offset);
+            pipeWriter.Advance(size);
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static ReadOnlyMemory<byte> SerializeToFrames<TMethod, THeader>(ref TMethod method, ref THeader header, ReadOnlyMemory<byte> body, ushort channelNumber, int maxBodyPayloadBytes)
+        public static void SerializeToFrames<TMethod, THeader, TBufferWriter>(ref TMethod method, ref THeader header, ReadOnlyMemory<byte> body, TBufferWriter pipeWriter, ushort channelNumber, int maxBodyPayloadBytes)
             where TMethod : struct, IOutgoingAmqpMethod
             where THeader : IAmqpHeader
+            where TBufferWriter : IBufferWriter<byte>
         {
             int remainingBodyBytes = body.Length;
             int size = Method.FrameSize + Header.FrameSize +
@@ -173,20 +159,20 @@ namespace RabbitMQ.Client.Impl
                        BodySegment.FrameSize * GetBodyFrameCount(maxBodyPayloadBytes, remainingBodyBytes) + remainingBodyBytes;
 
             // Will be returned by SocketFrameWriter.WriteLoop
-            var array = ArrayPool<byte>.Shared.Rent(size);
+            Span<byte> outputBuffer = pipeWriter.GetSpan(size);
 
-            int offset = Method.WriteTo(array, channelNumber, ref method);
-            offset += Header.WriteTo(array.AsSpan(offset), channelNumber, ref header, remainingBodyBytes);
+            int offset = Method.WriteTo(outputBuffer, channelNumber, ref method);
+            offset += Header.WriteTo(outputBuffer.Slice(offset), channelNumber, ref header, remainingBodyBytes);
             var bodySpan = body.Span;
             while (remainingBodyBytes > 0)
             {
                 int frameSize = remainingBodyBytes > maxBodyPayloadBytes ? maxBodyPayloadBytes : remainingBodyBytes;
-                offset += BodySegment.WriteTo(array.AsSpan(offset), channelNumber, bodySpan.Slice(bodySpan.Length - remainingBodyBytes, frameSize));
+                offset += BodySegment.WriteTo(outputBuffer.Slice(offset), channelNumber, bodySpan.Slice(bodySpan.Length - remainingBodyBytes, frameSize));
                 remainingBodyBytes -= frameSize;
             }
 
-            System.Diagnostics.Debug.Assert(offset == size, $"Serialized to wrong size, expect {size}, offset {offset}");
-            return new ReadOnlyMemory<byte>(array, 0, size);
+            System.Diagnostics.Debug.Assert(offset == size, "Serialized to wrong size", "Serialized to wrong size, expect {0:N0}, offset {1:N0}", size, offset);
+            pipeWriter.Advance(size);
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
@@ -201,7 +187,7 @@ namespace RabbitMQ.Client.Impl
         }
     }
 
-    internal readonly ref struct InboundFrame
+    internal readonly struct InboundFrame
     {
         public readonly FrameType Type;
         public readonly int Channel;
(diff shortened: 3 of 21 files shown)


stebet:
FYI @bollhals

mgravell:
I haven't looked at the code, but just an advisory: don't expose any of the Pipelines.Sockets.Unofficial types in the public API. The reason for this is that "bedrock" (i.e. the official transports) should add client connections in full at some point, at which point you'll probably want to disentangle my lib and use that instead.

stebet:
> I haven't looked at the code, but just an advisory: don't expose any of the Pipelines.Sockets.Unofficial types in the public API. The reason for this is that "bedrock" (i.e. the official transports) should add client connections in full at some point, at which point you'll probably want to disentangle my lib and use that instead.

Thanks for the pointer Marc. None of them are exposed in the public API, and we should make sure to keep it that way :)

michaelklishin:
@mgravell thank you for chiming in :)

projects/RabbitMQ.Client/client/impl/Frame.cs
```diff
@@ -136,25 +139,17 @@ internal static class Heartbeat
             /// <summary>
             /// Compiler trick to directly refer to static data in the assembly, see here: https://github.com/dotnet/roslyn/pull/24621
             /// </summary>
-            private static ReadOnlySpan<byte> Payload => new byte[]
+            internal static ReadOnlySpan<byte> Payload => new byte[]
```
bollhals, line 142: changing this most likely does fail the compiler trick (see comment)

projects/RabbitMQ.Client/client/impl/SocketFrameHandler.cs
```diff
@@ -68,31 +69,16 @@ internal class SocketFrameHandler : IFrameHandler
...
-                new UnboundedChannelOptions
-                {
-                    AllowSynchronousContinuations = false,
-                    SingleReader = true,
-                    SingleWriter = false
-                });
-
-            _channelReader = channel.Reader;
-            _channelWriter = channel.Writer;
+            _frameHeaderBuffer = new byte[8];
```
bollhals, line 81: Why 8? Thought we always read 7 bytes.

projects/RabbitMQ.Client/client/impl/Frame.cs
stebet: It shouldn't. The examples I saw used both private and public modifiers.

projects/RabbitMQ.Client/client/impl/SocketFrameHandler.cs
stebet: Was planning on filling the 8th when validating the protocol but wasn't needed. Doesn't make all the difference though.

bollhals:
I just quickly scanned through the PR, but I quickly realized I need to read up on the pipelines more to get a better understanding of the impact of this change.

projects/RabbitMQ.Client/client/impl/SocketFrameHandler.cs
```diff
@@ -232,14 +212,21 @@ public void Close()
...
+
+        public ValueTask<InboundFrame> ReadFrame()
+        {
+            return InboundFrame.ReadFrom(_pipeReader, _frameHeaderBuffer);
+        }
 
         public void SendHeader()
         {
-            byte[] headerBytes = new byte[8];
+            byte[] headerBytes = ArrayPool<byte>.Shared.Rent(8);
```
JanEggers, line 229: if you have a fixed small number of elements you should use stackalloc instead of rent its faster

projects/RabbitMQ.Client/client/impl/Frame.cs
```diff
@@ -169,22 +164,22 @@ private InboundFrame(FrameType type, int channel, ReadOnlyMemory<byte> payload,
...
                 }
 
-                int transportHigh = reader.ReadByte();
-                int transportLow = reader.ReadByte();
-                int serverMajor = reader.ReadByte();
-                int serverMinor = reader.ReadByte();
+                int transportHigh = protocolError.Span[3];
+                int transportLow = protocolError.Span[4];
+                int serverMajor = protocolError.Span[5];
+                int serverMinor = await reader.ReadByteAsync().ConfigureAwait(false); ;
```
JanEggers, line 182: this is bad you should read all bytes you need at minimum before calling ProcessProtocolHeader and make it sync

projects/RabbitMQ.Client/client/impl/Frame.cs
```diff
@@ -228,31 +224,27 @@ internal static InboundFrame ReadFrom(Stream reader, byte[] frameHeaderBuffer)
                     throw new EndOfStreamException("Reached the end of the stream. Possible authentication failure.");
                 case 'A':
                     // Probably an AMQP protocol header, otherwise meaningless
-                    ProcessProtocolHeader(reader);
+                    await ProcessProtocolHeader(reader, frameHeaderBuffer.Slice(1, 6)).ConfigureAwait(false);
```
JanEggers, line 227: why not Slice(1, 7) and avoid the async code in ProcessProtocolHeader

projects/RabbitMQ.Client/client/impl/SocketFrameHandler.cs
```diff
@@ -345,5 +312,20 @@ private void ConnectOrFail(ITcpClient socket, IPEndPoint endpoint, TimeSpan time
...
+            return _pipeWriter.GetSpan(sizeHint);
+        }
+
+        public void Commit(int bytesToCommit)
+        {
+            _pipeWriter.Advance(bytesToCommit);
+            ValueTask<FlushResult> task = _pipeWriter.FlushAsync();
+            if (!task.IsCompleted)
+            {
+                task.AsTask().Wait();
```
JanEggers, line 327: use at least getawaiter().getResult() to unwrap the correct exception in case of an error

projects/RabbitMQ.Client/util/PipeReaderExtensions.cs
```diff
@@ -0,0 +1,92 @@
+﻿using System;
+using System.Buffers;
+using System.IO;
+using System.IO.Pipelines;
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+
+namespace RabbitMQ.Client
+{
+    internal static class PipeReaderExtensions
```
JanEggers, line 10: whats the reason for all these extensions? why is the copy on read needed? looks like a waste to me

projects/RabbitMQ.Client/client/impl/IFrameHandler.cs
```diff
@@ -57,10 +58,12 @@ internal interface IFrameHandler
         ///<summary>Read a frame from the underlying
         ///transport. Returns null if the read operation timed out
```
JanEggers, line 59: this is not true for the new impl. if this behavior is desired the impl needs to add cancellationtokens to handle timeouts and return null.

projects/RabbitMQ.Client/client/impl/Frame.cs
JanEggers: this swich could also use a default case that throws because of protocol violation?!

projects/RabbitMQ.Client/client/impl/SocketFrameHandler.cs
stebet: stackalloc creates a Span, and that doesn't work with Encoding.GetBytes on net461.

projects/RabbitMQ.Client/client/impl/Frame.cs
stebet: This is an edge case that'll happen at most once during a connections lifetime, and it'll shut down right after, so I'm not worried about that here.

projects/RabbitMQ.Client/client/impl/Frame.cs
stebet: I might then read more than I want and complicate the code that comes later. Also, this is an edge case :)

projects/RabbitMQ.Client/client/impl/SocketFrameHandler.cs
stebet: You are right, I'll fix that :)

projects/RabbitMQ.Client/client/impl/IFrameHandler.cs
stebet: I'll take a closer look at this. It's not surfaced in any tests it seems.

projects/RabbitMQ.Client/client/impl/Frame.cs
stebet: Works the same :)
image

projects/RabbitMQ.Client/util/PipeReaderExtensions.cs
stebet: The short version, Pipelines reuse internal buffers, so you have to copy the content you want to keep from them befire calling AdvanceTo, since that'll return the buffer for reuse within the Pipelines later.

danielmarbach:
I'm a bit swamped right now so I haven't looked in details but have you considered leaving the code as is with the channels in between (which would also allow more sophisticated batching to be implemented later right?) and just float task completion sources so that things become await able on the upper layers where needed? This is would allows to build upon the existing optimizations without having to take a dependency just jet to pipelines. Again maybe my idea is mute anyway because I missed some details by just glancing over but anyway I wanted to drop the idea as early as possible.

This input was actually the thing I was hoping we could align on in the call I suggested so that all the reviewers share a common understanding of where things are heading.

But maybe it is enough to outline the rough plan and alternatives on an issue or a markdown in the repo

stebet:
The pipelines maintain internal buffers so they take care of the batching, and they are very low-allocation becauseibterbslly they reuse ValueTask objects and memory pools so all that maintenance goes away from the codebase, and instead it becomes a relatively simple network de(serializer) with the IO taken care of by the pipelines.

They also implement async Socket/Stream reading/writing with minimal allocations which isn't trivial to do in .NET Framework, due to the async APIs on streams implementing Task instead of ValueTask, which was improved in .NET Core.

And with proper async interfaces on the IO level, implementing low-alloc Async API for the client becomes a lot easier, as well as freeing us from manual reusable buffer maintenance for the IO which had to be done manually with the ArrayPools.

That was the biggest driver for this PR since it was relatively trivial to do:) But I'm loving this discussion!

danielmarbach:
Yeah and with the TCS approach ValueTask is definitely a bit out of the window.

danielmarbach:
To connect to the previous discussion as far as I'm aware pipelines are handy for single reader/writer scenarios which essentially means we have to synchronize access to it as you did. By doing so we make it possible to concurrently use the model but we pay the overhead again of the underlying synchronization. Are we making the right tradeoffs from the perspective of the impact this will have on the higher level abstractions (the internal benefits have already been outlined)?

stebet:
> To connect to the previous discussion as far as I'm aware pipelines are handy for single reader/writer scenarios which essentially means we have to synchronize access to it as you did. By doing so we make it possible to concurrently use the model but we pay the overhead again of the underlying synchronization. Are we making the right tradeoffs from the perspective of the impact this will have on the higher level abstractions (the internal benefits have already been outlined)?

The benchmarks I've done have not surfaced any throughput degradation in the optimal scenario (RabbitMQ server running on my local machine) and profiling shows little time spent waiting for locks. Would be great to have more benchmarks if anyone has a remote RabbitMQ server he can hammer on for more tests :)

stebet:
With the latest changes, it is now pretty trivial to create a fully Async version of BasicPublish, and pretty much all commands that do not require a response (which utilize the ModelSend method), so that's one stop closer to an async client :) I can provide an example later.

Commands that require a response need some more infrastructure to work though, but are perfectly doable, however it's better to just do one step at a time :)

bollhals:
I've read up a bit on pipelines and read your explanations, very interesting I must say =) 

What I do not quite get is the benefits they bring that we couldn't do with channels+streams or could do simpler now with pipelines. 
The one thing that bugs me is, I'm not sure with the protocol we use, we're getting the a lot benefit out of it. (E.g. We're always know how much we need to read / write per frame. There is no unknown size or "read until you find x" where manual bufferhandling would be required. We read the fixed size header and then the known sized payload.)

stebet:
It's the ValueTask async implementation I'm mostly after. That is not trivial to get right on .NET Framework, and also the IO buffer management which is taken care of by Pipelines. It basically unlocks a lot of async implementation paths that would otherwise require a lot of clever ValueTask implementations to get right without incurring A LOT of allocation overhead, especially on .NET Framework where there are no ValueTask implementations for Streams for example.

The reduced ArrayPool.Rent and Return is just an added benefit. Also, the automatic back-pressure handling which avoids possible OOM exceptions in the case of slow connections + rapid generation of big events due to the current unconstrained channels.

I'd highly suggest reading up on Marc Gravells excellent blog posts on the subject as well, they are very insightful and show the benefits of the ValueTask implementations clearly, as well as the effort involved in getting reusable ValueTasks right.

JanEggers:
are you planning to add more benchmarks? right now I only found only benchmarks for the serializer but nothing regarding to networking?

stebet:
Networking benchmarks are almost impossible to do reliably due their nature, but I'll add a couple of profiling results showing before and after allocations and CPU profiles.

JanEggers:
oki i will create some myself then. 

just a suggestion: I think the message processing should be changed to be synchronous so that the copy of the payload on receive is avoided.

with pipelines in place there is really no need to copy the payload. the transport side of the pipe will always maintain a read ahead buffer and the application side can just process the data synchronous.

stebet:
> just a suggestion: I think the message processing should be changed to be synchronous so that the copy of the payload on receive is avoided.

What message processing are you referring to?

stebet:
# Initial benchmarks

## .NET Framework 4.8
*Memory traffic before:*
Before_net48

*Memory traffic after:*
After_net48

*CPU tracing before:*
CPU_Before_net48

*CPU tracing after:*
CPU_After_net48

## .NET Core 3.1
*Memory traffic before:*
Before_netcoreapp31

*Memory traffic after:*
After_netcoreapp31

*CPU tracing before:*
CPU_Before_netcoreapp31

*CPU tracing after:*
CPU_After_netcoreapp31

JanEggers:
thx for the screenshots. 

but what im looking for is some kind of bounded benchmark. like sending 100K msg and consume 100K msg with some kind of static payloads. so you can actually compare them. with benchmark.net you can then easily see whats the memory overhead of that payload is and how is the latency. when you do manual profiling its always hard to compare because you are not sure if the workload is actually the same. 

with that kind of benchmarks you could verify performance related pr´s and I will try to create some. the goal should be zero allocation overhead per msg without degrading latency or other perf chracteristics. or changing existing the public api.

> What message processing are you referring to?

whatever the client does with the received msg.

I assume that the readerextensions copy the bytes from readresult into another buffer because there is some kind of detached processing going on in the client.

all im saying is that when all code path of handling the message are awaited / synchronized there should be no need to copy the bytes from the readresult into another buffer.

danielmarbach:
@stebet I think @JanEggers might have a point. With the introduction of the concurrency settings of the consumers can't we assume we are now in full control of the payload and thus reduce the copying? Maybe the previous code still made assumptions that are no longer true

projects/RabbitMQ.Client/client/impl/Connection.cs
```diff
@@ -871,7 +877,12 @@ public void HeartbeatWriteTimerCallback(object state)
             {
                 if (!_closed)
                 {
-                    Write(Client.Impl.Framing.Heartbeat.GetHeartbeatFrame());
+                    _writeLock.Wait();
+                    Span<byte> span = _frameHandler.PipeWriter.GetSpan(Client.Impl.Framing.Heartbeat.Payload.Length);
+                    Client.Impl.Framing.Heartbeat.Payload.CopyTo(span);
+                    _frameHandler.PipeWriter.Advance(Client.Impl.Framing.Heartbeat.Payload.Length);
+                    _frameHandler.Flush();
```
JanEggers, line 884: you should add try finally to the locking in case any method throws. this ofcourse applies to all places where the lock is used.

stebet: yeah, I've noticed some room for improvements after the profiling as well, will make some changes today :) Going to try a few scenarios.

stebet:
> all im saying is that when all code path of handling the message are awaited / synchronized there should be no need to copy the bytes from the readresult into another buffer.

Don't forget that the SocketFrameHandler reads individual frames, frames are then assembled from multiple parts (due to max-frame-size for example) into Commands(messages), which are then delivered to consumers, so copying will always need to take place. Frames from different channels could theoretically also be interleaved with other channels, but they are always in order for individual channels.

JanEggers:
> frames are then assembled from multiple parts (due to max-frame-size for example) into Commands(messages)

I wasnt aware of this. thx for clarifiing. 

this is out of scope for this pr but I had a quick look at the code and from my point of view CommandAssembler is responsible for combining frames into commands so it should decide weather the payload needs to be copied or not. that should at least avoid the copy for all commands that fit into a single frame.

stebet:
> > frames are then assembled from multiple parts (due to max-frame-size for example) into Commands(messages)
> 
> I wasnt aware of this. thx for clarifiing.
> 
> this is out of scope for this pr but I had a quick look at the code and from my point of view CommandAssembler is responsible for combining frames into commands so it should decide weather the payload needs to be copied or not. that should at least avoid the copy for all commands that fit into a single frame.

There is another caveat, the ReadResult buffers are resued as well, so we have to return them, preferably as early as possible so the Pipeline can reuse it for subsequent reads. That's what the `AdvanceTo` method does, it's to indicate what buffers you are done reading and/or have inspected. That's why the copies are being made immediately, but there might be some room for optimization there.

projects/RabbitMQ.Client/client/impl/Connection.cs
```diff
@@ -71,6 +70,7 @@ internal sealed class Connection : IConnection
         private volatile bool _running = true;
         private readonly MainSession _session0;
         private SessionManager _sessionManager;
+        internal SemaphoreSlim _writeLock = new SemaphoreSlim(1, 1);
```
danielmarbach, line 73: I haven't measured it but I seem to remember Semaphore is up to 10x slower than a lock. I think I read this once in

http://www.albahari.com/threading/part2.aspx

ah yeah here it is

image

but these numbers are a bit old and the picture might have changed. I guess as long as we are only doing sync work it might make sense to use the lock instead. The semaphore might come handy when we need to do async work and ensure single writer access

stebet: Yeah, I'm looking at better solutions, it might be that our previous Channels implementation with the Pipelines just replacing the Stream functionality (for a proper ReadAsync and FlushAsync with minimal allocations) is the best solution. Trying out that approach currently.

stebet:
I have made some further improvements in the latest checking, and I believe I'm getting closer to the sweet spot.

We now have Channels between both incoming and outgoing frames, and pipelines are working asynchronously in the background to write outgoing frames and push incoming frames. I'll clean up the code after the weekend and add before/after tests, but it's looking pretty damn good now both memory and performance wise and this now provides fully asynchronous IO :)

danielmarbach:
This starts to look very promising! Glad we have now a good proposal between what I suggested here https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/949#issuecomment-708557474 as well as the pipeline benefits.

projects/RabbitMQ.Client/client/impl/Frame.cs
```diff
@@ -164,96 +152,88 @@ private InboundFrame(FrameType type, int channel, ReadOnlyMemory<byte> payload,
...
             {
-                await reader.ReadAsync(frameHeaderBuffer.Slice(0, 7)).ConfigureAwait(false);
-
-                type = frameHeaderBuffer.Span[0];
+                frame = default;
+                return false;
             }
-            catch (IOException ioe)
+
+            if (buffer.First.Span[0] == 'A')
```
JanEggers, line 181: accessing chunks this way is  wrong.  just because the ReadOnlySequence conatins 8 bytes the must not be all in the first chunk. the first chunk may onls contain just one byte.

have a look at SequenceReader

JanEggers: https://docs.microsoft.com/en-us/dotnet/api/system.buffers.sequencereader-1?view=netcore-3.1

UnreadSpan is the right thing if you want to get the available data that should also eliminate the copies you do below

JanEggers: https://github.com/dotnet/runtime/blob/b6d428c3208cf3cc849a5f1f8c77d8663c9de4ea/src/libraries/System.Memory/src/System/Buffers/SequenceReader.cs

projects/RabbitMQ.Client/client/impl/Frame.cs
```diff
@@ -169,99 +152,88 @@ private InboundFrame(FrameType type, int channel, ReadOnlyMemory<byte> payload,
...
+
+            if (buffer.First.Length >= 7)
             {
-                case -1:
-                    throw new EndOfStreamException("Reached the end of the stream. Possible authentication failure.");
-                case 'A':
-                    // Probably an AMQP protocol header, otherwise meaningless
-                    ProcessProtocolHeader(reader);
-                    break;
+                channel = NetworkOrderDeserializer.ReadUInt16(buffer.First.Span.Slice(1, 2));
```
JanEggers, line 202: there are alos extension methods on the sequence reader like 
https://docs.microsoft.com/en-us/dotnet/api/system.buffers.sequencereaderextensions.tryreadbigendian?view=netcore-3.1

so

projects/RabbitMQ.Client/client/impl/SocketFrameHandler.cs
```diff
@@ -226,37 +225,32 @@ public void Close()
...
+                    _closed = true;
                 }
             }
         }
 
         public void SendHeader()
         {
-            byte[] headerBytes = ArrayPool<byte>.Shared.Rent(8);
-            Encoding.ASCII.GetBytes("AMQP", 0, 4, headerBytes, 0);
+            Span<byte> headerBytes = stackalloc byte[8];
```
JanEggers, line 249: why stackalloc + copy you could write directly into the pipe

projects/RabbitMQ.Client/client/impl/SocketFrameHandler.cs
```diff
@@ -283,28 +279,51 @@ public void Write(OutgoingFrame frame)
...
 
-                ReadOnlySequence<byte> buffer = readResult.Buffer;
-                if (buffer.IsEmpty)
-                {
-                    _frameWriter.Complete(new EndOfStreamException());
-                    return;
-                }
+                    int handled = 0;
+                    ReadOnlySequence<byte> buffer = readResult.Buffer;
+                    if (!buffer.IsEmpty)
```
JanEggers, line 295: this is redundand

projects/RabbitMQ.Client/client/impl/SocketFrameHandler.cs
stebet: You are right :) will fix that.

projects/RabbitMQ.Client/client/impl/SocketFrameHandler.cs
stebet: It might still return 0 even if the buffer is not empty, since we might not have an entire frame to read.

stebet:
Here are benchmarks from the latest runs after improvements and code fixes.

**Benchmark: Send and receive 500.000 messages with a 512 byte payload with a Guid set as correlation id.**

# .NET Framework 4.8

** Memory before **
net48_before_memory

** Memory after **
net48_after_memory

** CPU before **
net48_before_cpu

** CPU after **
net48_after_cpu

# .NET Core 3.1

** Memory before **
netcoreapp31_before_memory

** Memory after **
netcoreapp31_after_memory

** CPU before **
netcoreapp31_before_cpu

** CPU after **
netcoreapp31_after_cpu

stebet:
Benchmark code:
```csharp
class Program
{
    private static int itemsPerBatch = 5000;
    private static ManualResetEventSlim manualResetEventSlim = new ManualResetEventSlim();
    {

        var asyncListener = new AsyncEventingBasicConsumer(subscriber);
        subscriber.QueueBind("testqueue", "test", "");
        subscriber.BasicConsume("testqueue", true, "testconsumer", asyncListener);
        byte[] payload = new byte[512];

        {
            {
                var header = publisher.CreateBasicProperties();
                header.CorrelationId = $"{Guid.NewGuid()}";
                publisher.BasicPublish("test", "", false, header, payload);
                messagesSent++;
            }
        });

        await batchPublish.ConfigureAwait(false);
        manualResetEventSlim.Wait();

        subscriber.Dispose();
        publisher.Dispose();
        connection.Dispose();
        connection2.Dispose();
    }

    private static Task AsyncListener_Received(object sender, BasicDeliverEventArgs @event)
    {
        if (messagesReceived == batchesToSend * itemsPerBatch)
        {
            manualResetEventSlim.Set();
        }

        return Task.CompletedTask;
    }
}
```

stebet:
Rebased and squished the commits to clean stuff up

projects/RabbitMQ.Client/client/api/ConnectionFactoryBase.cs
```diff
@@ -49,12 +52,8 @@ public class ConnectionFactoryBase
         /// <returns>New instance of a <see cref="TcpClient"/>.</returns>
         public static ITcpClient DefaultSocketFactory(AddressFamily addressFamily)
         {
-            var socket = new Socket(addressFamily, SocketType.Stream, ProtocolType.Tcp)
-            {
-                NoDelay = true,
-                ReceiveBufferSize = 65536,
```
bollhals, line 55: You played around with the buffer size, what's the final verdict now?

projects/RabbitMQ.Client/client/impl/SocketFrameHandler.cs
```diff
@@ -68,25 +68,25 @@ internal class SocketFrameHandler : IFrameHandler
...
         {
             Endpoint = endpoint;
-            _frameHeaderBuffer = new byte[6];
-            var channel = Channel.CreateUnbounded<ReadOnlyMemory<byte>>(
+            _frameHeaderBuffer = new byte[8];
+            var channel = Channel.CreateUnbounded<OutgoingFrame>(
                 new UnboundedChannelOptions
                 {
-                    AllowSynchronousContinuations = false,
+                    AllowSynchronousContinuations = true,
```
bollhals, line 89: What's the resulting change for this?
Does publish now possibly write to the pipe/socket? If so is this what we want?

bollhals:
> It's the ValueTask async implementation I'm mostly after. That is not trivial to get right on .NET Framework, and also the IO buffer management which is taken care of by Pipelines. It basically unlocks a lot of async implementation paths that would otherwise require a lot of clever ValueTask implementations to get right without incurring A LOT of allocation overhead, especially on .NET Framework where there are no ValueTask implementations for Streams for example.

Question: How important is .NET Framework for 7.0?

> Here are benchmarks from the latest runs after improvements and code fixes.

From what I can read out of the images, it seems that:
- There are a lot less byte[] allocation => Mainly due to delaying the serialization from before writing it to channel to in the processing of the channel.
(Keep in mind this kind of allocations only happen, if we mass publish. If you put in a small sleep to simulate other work every 50 publishes, the byte[] allocation drops to ~2 MB. I'd argue that mass publish is rather unlikely. (We can also change to a bounded channel to prevent mass allocation.))
- Apart from byte[] allocations on .NET Core, there are only few minor changes (e.g. OutgoingFrame[] due to the channel and the size of the struct being bigger than ReadOnlyMemory<byte>) 
- Apart from byte[] allocations on .NET Framework, there are a couple of new allocations (e.g. ExecutionContext, IOCompletionCallback, ...)
- It looks like it uses more threads in parallel, but generally also a lot more CPU time. (Slightly more total time as well)


So while I really like the pipeline (thanks for forcing me to finally read up on them 👍 ) I have trouble seeing the benefits while also the code is getting harder to understand (at least for now, since not a lot are experts on pipes). Also on .NET Framework I struggle to see big benefits.

PS: I have a local change, that modifies all "Method"-Classes to be structs, and hence dropping the allocations. But in order to do that, It needs to serialize the data before writing it to the channel. (Structs can't be inherited and the different structs can't be put in the OutputFrame). I don't see a way of dropping these allocations with this change (excluding maybe pooling).

projects/RabbitMQ.Client/client/impl/SocketFrameHandler.cs
stebet: Publish will write to the channel, but it might run synchronously if the WriteLoop completes synchronously, for example Pipeline had enough empty buffer to receive the data. This is what we want to avoid the async overhead for synchronous completions since the Pipelines take care of async.

projects/RabbitMQ.Client/client/api/ConnectionFactoryBase.cs
stebet: Leave it at default values, Pipelines have their own internal buffer and manage it themselves. Too big of a buffer at the socket level can confuse Pipelines and result in operations hanging.

stebet:
I have yet to do more benchmarks, as the ones I've done are not very parallel and therefore induce a lot of overhead in async calls, but again, I mainly see this PR as a stepping stone to be able to provide a real low-allocation async API for the client. ValueTask is already available in .NET Framework. Also, having pipelines in place makes the client a lot more ready for "Bedrock" which are the new Client/Server abstractions being introduced to .NET.

Yes the code is more convoluted, but that's mostly isolated to the IO part, which frankly can be very hard to do efficiently and asynchronoysly in a clean way (see Task based Streams in .NET Framework for reference, which have HUGE allocation penalties, especially for small reads).

I also learned a lot, and there is no pressure on accepting this PR, but I can show some further examples of how this can provide us with a true async API, hopefully side-by-side with the existing API, which I'm planning on experimenting with next :)

JanEggers:
> I have a local change, that modifies all "Method"-Classes to be structs

nice one, I was also thinking about that one to get heap allocations down

projects/RabbitMQ.Client/client/impl/SocketFrameHandler.cs
bollhals: But this also means that the thread that called publish might be blocked for a longer time, isn't-it?

projects/RabbitMQ.Client/client/impl/SocketFrameHandler.cs
```diff
@@ -262,51 +267,105 @@ public void SendHeader()
...
         public void Write(OutgoingFrame frame)
         {
-            _channelWriter.TryWrite(frame);
+            LockToken token = default;
+            try
+            {
+                token = _singleWriterMutex.TryWait(WaitOptions.NoDelay);
+                if (!token.Success)
+                {
+                    // Didn't get the lock immediately, let's backlog the write and start the backlog writer task.
```
bollhals, line 276: General question, this makes it possible that the order of published messages could get changed. Is this "allowed"? Does any documentation say something about this?

projects/RabbitMQ.Client/client/impl/SocketFrameHandler.cs
```diff
@@ -262,51 +267,105 @@ public void SendHeader()
...
+                    if (_channelWriter.TryWrite(frame))
+                    {
+                        if (_writerTask == null || _writerTask.Status == TaskStatus.RanToCompletion)
+                        {
+                            _writerTask = Task.Run(WriteLoop, CancellationToken.None);
```
bollhals, line 281: This (although unlikely) could lead to a race condition where two Tasks are created.

stebet: Yeah, I have yet to clean that up a bit, thanks for pointing it out :)

projects/RabbitMQ.Client/client/impl/SocketFrameHandler.cs
stebet: Can they? Considering that we are using a write lock, which only one can hold at a time, if we can't get the lock synchronously, we'll add the write to the backlog, since the backlog is the one holding the lock. I'll take a closer look at this as well, I'm shamelessly stealing this logic from StackExchange.Redis :)

bollhals: Good question,  my thought process was:
- T1 has the lock, WriteFrameToPipe
- T2 didn't get the lock, wrote frame to channel, about to start the Task.Run
- T3 just entered the method

=> If T1 releases the lock & T3 gets the lock before the channel gets it, it will publish T3 before T2's message.

But now that I think of it, this is most probably irrelevant, as it uses multiple threads which the order will never be guaranteed anyway...

lukebakken:
@stebet I'll find time to rebase this branch on `main` as well as figure out how this PR relates to #982

stebet:
It might also be time to have it updated a bit. I can maybe take a look at it this week.

lukebakken:
@stebet just so we don't duplicate work, I'm going to manually apply these changes to a new branch. I'll push it up as soon as I make some progress.

Doing a merge or rebase and then resolving conflicts makes my eyes cross.

stebet:
I'm working on refreshing this PR and rebuilding it on-top of the latest master. I'll probably force-update it once done :)

lukebakken:
@stebet thank you.

stebet:
So, I'm working in updating this PR, changed a bit and rebased on the latest `main` build , so consider everything above to be outdated. Working on benchmarks now!

stebet:
Due to the length of this PR convo, I'm going to close this PR and create a fresh one considering the feedback from this one to explain things in more details and to clear up all the clutter.

PR #954, opened 2020-10-19
Minor improvements to GitHub Actions CI build.

stebet:
## Proposed Changes

* Using the RabbitMQ alpine image for faster download and test init.
* Adding more verbose test output to output passed tests to better follow progress.

## Types of Changes

## Checklist

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -12,7 +12,7 @@ jobs:
 
     services:
       rabbitmq:
-        image: rabbitmq:3.8.9-management-alpine
+        image: rabbitmq:3.8.9-management
         ports:
          - 5672:5672
          - 15672:15672
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -12,7 +12,7 @@ jobs:
 
     services:
       rabbitmq:
-        image: rabbitmq:3.8.9-management
+        image: rabbitmq:3.8.9-management-alpine
         ports:
          - 5672:5672
          - 15672:15672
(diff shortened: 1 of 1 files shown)


.github/workflows/test-linux.yml
```diff
@@ -12,7 +12,7 @@ jobs:
 
     services:
       rabbitmq:
-        image: rabbitmq:3.8.9-management
+        image: rabbitmq:3.8.9-management-alpine
```
michaelklishin, line 15: @stebet our team uses Ubuntu or Debian across the board (where we have the choice). What kind of difference does this make to build times in practice?

@gerhard would you agree that we should not place any bets on the Alpine version of the image being around in the long term?

stebet: Since this is just the service container, I thought running the alpine image to run the tests made more sense since it's 67 megs compared to 83. It can easily be changed back at any time though, if the alpine image is not something that will be supported going forward.

gerhard: I understand the preference for smaller image sizes that Alpine images have. Because Alpine uses musl, a newer and smaller libc implementation that targets architectures with small chaches, we cannot endorse it as strongly as Ubuntu, which uses glibc (GNU C library), and is also what most other Linux distributions use.

While musl is cleaner, lighter, and apparently has a better dev reputation than glibc, Erlang BEAM which has been around since 1993, couldn't be compiled on musl without patching until OTP v21 shipped in 2018. This would explain the anecdotal evidence about Erlang instabilities on musl which we have not seen on glibc.

I would recommend sticking with the Ubuntu variant and avoiding a host of potential musl-related issues. Long-term & time permitting, team RabbitMQ would like to contribute a `rabbitmq` container image which is based on [Tiny](https://github.com/paketo-buildpacks/stacks#tiny), meaning distroless-like Ubuntu bionic with glibc + openssl + CA certs. This will be similar in size to the Alpine variant, but with none of the musl drawbacks.

stebet: Great explanation, thx. I'll change it back to the other image.

michaelklishin (approved):

michaelklishin:
Thank you!

PR #962, opened 2020-10-28
Structify the Framing.Impl methods

bollhals:
## Proposed Changes

The main focus of this change was to change all classes of Framing.Impl to be structs.

## Types of Changes

## Checklist

## Further Comments

Allocation before (Master)
image

Allocations now
image

Side effect is a reduction of the dll size of 11KB
image

Performance is hard to measure, the test app we otherwise use won't show a meaningful difference, so I used to benchmarks provided by @JanEggers:

2nd Line is always this PR

New Connection with 10000 messages
|              Method |     Mean |   Error |  StdDev | Ratio |     Gen 0 |    Gen 1 | Gen 2 | Allocated |
|-------------------- |---------:|--------:|--------:|------:|----------:|---------:|------:|----------:|
| Publish_Hello_World | 190.6 ms | 3.71 ms | 6.49 ms |  1.00 | 1000.0000 | 333.3333 |     - |   4.97 MB |
| Publish_Hello_World | 187.8 ms | 3.75 ms | 3.86 ms |  1.00 | 666.6667 |     - |     - |   4.12 MB |

Existing Connection but with 1000 messages
|              Method |     Mean |    Error |   StdDev | Ratio |    Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------- |---------:|---------:|---------:|------:|---------:|------:|------:|----------:|
| Publish_Hello_World | 22.37 ms | 0.429 ms | 0.511 ms |  1.00 | 93.7500 |     - |     - | 508.24 KB |
| Publish_Hello_World | 21.41 ms | 0.408 ms | 0.362 ms |  1.00 | 62.5000 |     - |     - | 420.11 KB |


projects/RabbitMQ.Client/client/impl/AutorecoveringModel.cs
```diff
@@ -297,177 +296,6 @@ private void Dispose(bool disposing)
             }
         }
 
-        public void ConnectionTuneOk(ushort channelMax,
```
bollhals, line 300: These were all never actually called

projects/RabbitMQ.Client/client/impl/SessionBase.cs
```diff
@@ -159,31 +159,48 @@ public void Notify()
...
-        public virtual void Transmit(in OutgoingCommand cmd)
+        public virtual void Transmit<T>(in T cmd) where T : struct, IOutgoingCommand
         {
             if (CloseReason != null)
             {
-                lock (_shutdownLock)
+                CheckCanSendWhileClosed(cmd.ProtocolCommandId);
+            }
+
+            Connection.Write(Framing.SerializeToFrames(cmd, ChannelNumber));
```
bollhals, line 169: @stebet Here is the conflict that I've mentioned in your PR for the pipelines. 

I get T as input and there is no common type we could use to pass it on to the connection (except the ROM<byte>) without allocation. (If you use the interface IOutgoingCommand then due to T being a struct it will be boxed to to be stored in the interface field, as it doesn't know the actual size of the struct so it stores it as a reference.

projects/RabbitMQ.Client/client/impl/AutorecoveringModel.cs
michaelklishin: Not every method is called by the client, some are received from the server. It's weird to see this particular one on a channel since `connection.tune` is an early connection negotiation step. So I suspect it genuinely was not used.

projects/Benchmarks/WireFormatting/WireFormatting_Read_BasicAck.cs
```diff
@@ -16,9 +16,9 @@ public WireFormatting_Read_BasicAck()
         }
 
         [Benchmark(Baseline = true)]
-        public object ReadFromSpan()
+        public void ReadFromSpan()
```
stebet, line 19: This might not have the behavior you want. The compiler could optimize away the logic since a new variable is being created without being used. Having the method return the new object it will force it to stay. See here: https://benchmarkdotnet.org/articles/guides/good-practices.html#avoid-dead-code-elimination

projects/Benchmarks/WireFormatting/WireFormatting_Read_BasicDeliver.cs
```diff
@@ -12,13 +12,17 @@ public class WireFormatting_Read_BasicDeliver
...
+            int offset = RabbitMQ.Client.Impl.WireFormatting.WriteShortstr(_buffer, string.Empty);
+            offset += RabbitMQ.Client.Impl.WireFormatting.WriteLonglong(_buffer.AsSpan(offset), 0);
+            offset += RabbitMQ.Client.Impl.WireFormatting.WriteBits(_buffer.AsSpan(offset), false);
+            offset += RabbitMQ.Client.Impl.WireFormatting.WriteShortstr(_buffer.AsSpan(offset), string.Empty);
+            RabbitMQ.Client.Impl.WireFormatting.WriteShortstr(_buffer.AsSpan(offset), string.Empty);
         }
 
         [Benchmark(Baseline = true)]
-        public object ReadFromSpan()
+        public void ReadFromSpan()
```
stebet, line 23: Possible dead code elimination.

projects/Benchmarks/WireFormatting/WireFormatting_Read_ChannelClose.cs
```diff
@@ -16,9 +16,9 @@ public WireFormatting_Read_ChannelClose()
         }
 
         [Benchmark(Baseline = true)]
-        public object ReadFromSpan()
+        public void ReadFromSpan()
```
stebet, line 19: Possible dead code elimination.

projects/RabbitMQ.Client/client/impl/AutorecoveringModel.cs
stebet: On that note, in one of my PRs I simply removed the negotiation methods from the public facing API since they are just called during negotiation and should never be called explicitly. Something to consider for 7.0.0.

projects/RabbitMQ.Client/client/impl/Frame.cs
```diff
@@ -47,7 +48,7 @@ internal static class Framing
          * +------------+---------+----------------+---------+------------------+
          * | 1 byte     | 2 bytes | 4 bytes        | x bytes | 1 byte           |
          * +------------+---------+----------------+---------+------------------+ */
-        private const int BaseFrameSize = 1 + 2 + 4 + 1;
+        public const int BaseFrameSize = 1 + 2 + 4 + 1;
```
stebet, line 51: would internal suffice?

projects/RabbitMQ.Client/client/impl/Frame.cs
```diff
@@ -152,6 +153,56 @@ public static Memory<byte> GetHeartbeatFrame()
...
+            var size = Method.FrameSize + method.GetRequiredBufferSize();
+            var array = ArrayPool<byte>.Shared.Rent(size);
+            var offset = Method.WriteTo(array, channelNumber, method);
+
+            if (offset != size)
+            {
+                throw new InvalidOperationException($"Serialized to wrong size, expect {size}, offset {offset}");
+            }
+
+            return new ReadOnlyMemory<byte>(array, 0, size);
```
stebet, line 168: use .AsMemory(0, size) extension method for simplicity?

projects/RabbitMQ.Client/client/impl/ISession.cs
```diff
@@ -73,8 +73,8 @@ internal interface ISession
         void Close(ShutdownEventArgs reason, bool notify);
         bool HandleFrame(in InboundFrame frame);
         void Notify();
-        void Transmit(in OutgoingCommand cmd);
-        void Transmit(IList<OutgoingCommand> cmds);
-
+        void Transmit<T>(in T cmd) where T : struct, IOutgoingCommand;
```
stebet, line 76: remove in favor of the one below with default parameters?

projects/RabbitMQ.Client/client/impl/ISession.cs
```diff
@@ -73,8 +73,8 @@ internal interface ISession
...
+        void Transmit<T>(in T cmd, ContentHeaderBase header, ReadOnlyMemory<byte> body) where T : struct, IOutgoingCommand;
```
stebet, line 77: change to `void Transmit<T>(in T cmd, ContentHeaderBase header = default, ReadOnlyMemory<byte> body = default) where T : struct, IOutgoingCommand;` to cover both Transmit cases?

projects/RabbitMQ.Client/client/impl/SessionBase.cs
stebet: That's ok, worst case, even though ugly, we could simply create an overload for `Connection.Write` for each possible struct. It's ugly, but it gets the job done and still maintains the benefits of this PR.

stebet (commented):
I really like the changes in this PR :) It simplifies a lot of code.

projects/RabbitMQ.Client/client/impl/SessionBase.cs
bollhals: The issue isn't the Write method itself, it's the channel's storage type (Outputframe vs ROM<byte>). The Outputframe struct can't store the different structs without boxing them => Benefits of this PR are lost.

bollhals:
> I really like the changes in this PR :) It simplifies a lot of code.

Thanks for your review!

projects/Benchmarks/WireFormatting/WireFormatting_Read_BasicAck.cs
bollhals: I was thinking about it as well, the benchmark didn't change though.
Why did I change it? object return would box and I can't put the actual struct there, as it is internal. I haven't found a way to declare benchmarks as internal and being able to execute it. Is there?

projects/RabbitMQ.Client/client/impl/Frame.cs
bollhals: The Framing class is internal. I usually don't use internals in internal classes.

projects/RabbitMQ.Client/client/impl/Frame.cs
bollhals: AsMemory creates a Memory<byte> not a ReadOnlyMemory<byte> => implicit cast

projects/RabbitMQ.Client/client/impl/ISession.cs
bollhals: possible, but there are 2 benefits by not doing so
- You either have just cmd or cmd, header & body. Default parameters would enable the possibility of just leaving out the body. 
- Performance, as the less that needs to be passed, the better. Also Framing.SerializeToFrames is more performant due to the reduced parameters

bollhals:
My plan  would be to merge this one before the first 8.0 PRs, as this might be also relevant for backporting to 6.x which will be simpler before 8.0 PR hitting master.

I'll update the code in the next days and trying to get this in a mergable state.

bollhals:
> My plan would be to merge this one before the first 8.0 PRs, as this might be also relevant for backporting to 6.x which will be simpler before 8.0 PR hitting master.
> 
> I'll update the code in the next days and trying to get this in a mergable state.

Done, rebased to master and fixed all failing tests.

stebet:
Will try to review this soon. A bit busy with a newborn a.t.m :)

bollhals:
> Will try to review this soon. A bit busy with a newborn a.t.m :)

congrats =)

bollhals:
rebased to master. this is ready for review

bollhals:
Updated with latest master and ran new benchmarks & updated initial post.

bollhals:
@stebet @michaelklishin Can we get this reviewed? Otherwise this is going to end up in another conflicting state with master due to the next PR (like #1024).

bollhals:
Rebased again against master and resolved all conflicts. Would you mind taking a look so we finally can get this done? @stebet @michaelklishin

stebet:
> Rebased again against master and resolved all conflicts. Would you mind taking a look so we finally can get this done? @stebet @michaelklishin 

Will do

projects/RabbitMQ.Client/client/framing/BasicAck.cs
```diff
@@ -35,14 +35,10 @@
...
+    internal readonly struct BasicAck : IOutgoingAmqpMethod
     {
-        public ulong _deliveryTag;
-        public bool _multiple;
-
-        public BasicAck()
-        {
-        }
+        public readonly ulong _deliveryTag;
+        public readonly bool _multiple;
```
stebet, line 41: Would be good since these are public properties to `PascalCase` them. Same applies to all the other *Method structs.

projects/RabbitMQ.Client/client/impl/ModelBase.cs
```diff
@@ -332,38 +334,64 @@ private void HandleCommand(in IncomingCommand cmd)
             }
         }
 
-        protected T ModelRpc<T>(MethodBase method) where T : MethodBase
+        protected void ModelRpc<TMethod>(in TMethod method, ProtocolCommandId returnCommandId)
```
stebet, line 337: What's the reason for removing the ProtocolCommandId from the method classes and adding it as an extra parameter?

projects/RabbitMQ.Client/client/impl/ModelBase.cs
```diff
@@ -332,38 +334,64 @@ private void HandleCommand(in IncomingCommand cmd)
...
+            reply.ReturnMethodBuffer();
+
+            if (reply.CommandId != returnCommandId)
+            {
+                throw new UnexpectedMethodException(reply.CommandId, returnCommandId);
+            }
+        }
+
+        protected TReturn ModelRpc<TMethod, TReturn>(in TMethod method, ProtocolCommandId returnCommandId, Func<ReadOnlyMemory<byte>, TReturn> createFunc)
+            where TMethod : struct, IOutgoingAmqpMethod
```
stebet, line 358: You should be able to add a constraint to the TReturn as well as a `struct, IAmqpMethod` right?

projects/RabbitMQ.Client/client/framing/TxSelect.cs
```diff
@@ -34,22 +34,16 @@
 
 namespace RabbitMQ.Client.Framing.Impl
 {
-    internal sealed class TxSelect : Client.Impl.MethodBase
+    internal readonly struct TxSelect : IOutgoingAmqpMethod
```
stebet, line 37: For empty structs like the `Tx*` structs, we might as well create a static singleton since there is no reason to even new them up?

projects/RabbitMQ.Client/client/framing/BasicAck.cs
bollhals: Yes, but I kept it the same for now. Can be done in a future PR.

projects/RabbitMQ.Client/client/impl/ModelBase.cs
bollhals: There are a lot of methods which are not interested in parsing the returned method (or is empty) so by passing the id, we can do the check if it is the correct response. If you want to parse the result, there are extra methods for it.

projects/RabbitMQ.Client/client/impl/ModelBase.cs
bollhals: Correct, but maybe you don't want to return the method itself but rather a subset of properties, this is why I left it unconstrained. (Look at QueuePurge I think, which returns only a subset of it (messages deleted))

projects/RabbitMQ.Client/client/framing/TxSelect.cs
bollhals: Possible, but the benefits are very limited (new for this type of struct is "basically a noop" as it has a size of 0, the change in useage makes it more confusing than benefiting I think.

bollhals:
rebased to master again. ready for final reviews

bollhals:
ping

bollhals:
@michaelklishin Can we finish this PR please?

bollhals:
🚨 Anything I can help to get this done? 🚨 

PS: What is the plan for 7.0?

stebet:
After the latest reviews I'm all good for my part :)

bollhals:
Who could I ping to get this done? :) feeling a bit left out here

michaelklishin:
@bollhals sorry, there were significant changes on our team in the last few months (mostly positive) and .NET client has slipped under the radar.

Issue #965, opened 2020-10-29
Proposal: LazyBasicProperties / IReadonlyBasicProperties 


JanEggers:
https://user-images.githubusercontent.com/661658/96521641-37aacc00-1261-11eb-89d9-abe85fbc2dec.png

shows that currently there are large allocations of BasicProperties and subsequent strings contained in the BasicProperties 

My proposal is to either split IBasicProperties or create IReadonlyBasicProperties with only get accessors.

(i didnt do the research yet maybe IBasicProperties can remain and just remove all the setters)

That way we can provide a new Implementation for incoming messages that just deserializes the properties that are actually accessed resulting in much lower computational and memory overhead.

michaelklishin:
We cannot use such a read-only interface on the publisher side but we should indeed be able to on the consumer/delivery side. Makes sense.

stebet:
I was thinking **the exact same thing** yesterday. My thought was that requires the BasicProperties to by IDisposable though, since we'd want to keep a copy of the bytes ready, but only deserialize on access, possibly incurring just a memory-copy cost (which can be pooled) but possibly avoiding the deserialization overhead. But lazily deserializing them seems like the correct approach i.m.o if we can easily take care of the disposal.

JanEggers:
from my pov no dispose / copy is needed.

we just need to add to the docs that BasicProperties are only valid until the handler returns just like for the body bytes.

stebet:
It's created out of a header frame so we could take over the payload just like we do with body frames, which will require disposing of it when done. I'll do a little test and report back.

bollhals:
Question is, would it be possible to implement it as a "readonly struct" (We'd have to investigate whether we could even make it ref, to prevent storing it anywhere.) We'd also have to adapt all the passing of the type to use "in" to prevent copying.
Or do we have to implement it as class? (E.g. should the public API use an interface or a struct)

JanEggers:
I think the public api is and should be an interface. 

And I think it should be implemented as a class to avoid boxing and copiing. Also the instance itself is not the problem, the real problems are the many properties. If we only store one property of type memory<T> and introduce the lazy getters it should get much smaller. Im still thinking about if its better to have one memory<T> property that contains all the data or one slice of memory per property and just dont create the strings.  

[...]

PR #967, opened 2020-10-30
Reducing string allocations when reading BasicProperties.

stebet:
## Proposed Changes

Reduces string allocations in BasicProperties by only deserializing the strings when needed.

## Types of Changes

- [X] Bug fix (non-breaking change which fixes issue #965)

## Checklist

## Further Comments

This works by storing a reference to the bytes read for header frames in BasicProperties, and lazily reading the strings as they are referenced. For consumer workloads that do not access all the data in BasicProperties, this saves CPU and allocations.


stebet:
Test app:
```csharp
class Program
{
    private static int itemsPerBatch = 5000;
    private static ManualResetEventSlim manualResetEventSlim = new ManualResetEventSlim();
    {
        var connectionFactory = new ConnectionFactory() { DispatchConsumersAsync = false, Uri = connectionString };
        connectionFactory.ClientProvidedName = "BenchMark";

        var listener = new EventingBasicConsumer(subscriber);
        listener.Received += Listener_Received;
        subscriber.QueueBind("testqueue", "test", "");
        subscriber.BasicConsume("testqueue", true, "testconsumer", listener);
        byte[] payload = new byte[512];

        {
            var header = publisher.CreateBasicProperties();
            header.CorrelationId = $"{Guid.NewGuid()}";
            publisher.BasicPublish("test", "", false, header, payload);
            messagesSent++;
        }

        manualResetEventSlim.Wait();

        Console.ReadLine();
        subscriber.Dispose();
        publisher.Dispose();
        connection.Dispose();
        connection2.Dispose();
    }

    private static void Listener_Received(object sender, BasicDeliverEventArgs e)
    {
        if (messagesReceived == batchesToSend * itemsPerBatch)
        {
            manualResetEventSlim.Set();
        }
    }
}
```

**Allocations before:**


**Allocations after:**

JanEggers:
who creates the remaining 1.5 mio strings?

stebet:
> what creates the remaining 1.5 mio strings?

The benchmark app :)

JanEggers:
ahh you are setting the correlationid

JanEggers:
maybe we can pool the BasicProperties on the receiving end and return them when returning the bytearray

projects/RabbitMQ.Client/client/framing/BasicProperties.cs
```diff
@@ -166,69 +284,57 @@ public override string ClusterId
...
+        public override bool IsHeadersPresent() => _headers != default || (_readOnlyMemory.Length > 0 && (_readOnlyMemory.Span[0] & 0b0010_0000) != 0);
 
-        public override bool IsContentEncodingPresent() => _contentEncoding != default;
+        public override bool IsDeliveryModePresent() => _deliveryMode != default || (_readOnlyMemory.Length > 0 && (_readOnlyMemory.Span[0] & 0b0001_0000) != 0);
 
-        public override bool IsHeadersPresent() => _headers != default;
+        public override bool IsPriorityPresent() => _priority != default || (_readOnlyMemory.Length > 0 && (_readOnlyMemory.Span[0] & 0b0000_1000) != 0);
 
-        public override bool IsDeliveryModePresent() => _deliveryMode != default;
+        public override bool IsCorrelationIdPresent() => _correlationId != default || (_readOnlyMemory.Length > 0 && (_readOnlyMemory.Span[0] & 0b0000_0100) != 0);
```
danielmarbach, line 297: Can we somehow collapse the hierarchy and thus I line those methods?

projects/RabbitMQ.Client/client/framing/BasicProperties.cs
```diff
@@ -56,85 +61,198 @@ internal sealed class BasicProperties : RabbitMQ.Client.Impl.BasicProperties
...
 
         public override IDictionary<string, object> Headers
         {
-            get => _headers;
+            get
+            {
+                if (_headers == default && IsHeadersPresent())
+                {
+                    WireFormatting.ReadDictionary(_readOnlyMemory.Span.Slice(HeadersOffset), out Dictionary<string, object> tempDirectory);
+                    _headers = tempDirectory;
```
danielmarbach, line 97: Temp Directory?

projects/RabbitMQ.Client/client/framing/BasicProperties.cs
stebet: I was wondering about that actually, I was thinking that calculating the IsPresent and Offsets to read the properties could be loaded all at once when the first property is accessed to also prevent calculations from being repeated. It should be easy to read and calculate those without allocating on first access.

projects/RabbitMQ.Client/client/framing/BasicProperties.cs
stebet: IDictionary<,> vs Dictionary<,> :( Unfortunately we are exposing IDictionary<,> interfaces in a few places and it isn't co/contravariant :(

bollhals:
I know it's not about the test app, but 
`header.CorrelationId = $"{Guid.NewGuid()}";`
Doesn't that allocate 2 strings? (1 for Guid.NewGuid().ToString() and 1 to put the brackets around? => Guid.NewGuid().ToString("B") would get the same output if I'm not mistaken.

I'd be interested in the CPU aspect of this change! The whole "figuring out where my string should start and whether I already have it stored"-part doesn't sound cheap to me.

Downside is that the BasicProperty class got even bigger (+8 MB / 1m objects) and that we pay the prize as well when we have the class to send and not when receiving. 
I like the idea in principal, but I think this would make much more sense if we split the receiving property type from the sending property type.

projects/RabbitMQ.Client/client/framing/BasicProperties.cs
YulerB: tempDirectory, can we rename to tempDictionary.

stebet:
> Are you sure? That is, the application will gradually gain memory while reading unique data, such as `correlationid` header.
> 
> https://docs.microsoft.com/en-us/dotnet/api/system.string.intern?view=netcore-3.1#performance-considerations:
> 
> изображение

Spot on. That's why I thought I'd only intern certain strings, which will certainly be duplicated, such as exchange names for example.

projects/RabbitMQ.Client/client/framing/BasicProperties.cs
YulerB: It's possible to read all the is present values into a flags enum bypassing all the read bits related code.
For an example:

https://github.com/YulerB/rabbitmq-dotnet-client/blob/master/projects/client/RabbitMQ.Client/src/client/Protocol/BasicProperties.cs

stebet:
> I know it's not about the test app, but
> `header.CorrelationId = $"{Guid.NewGuid()}";`
> Doesn't that allocate 2 strings? (1 for Guid.NewGuid().ToString() and 1 to put the brackets around? => Guid.NewGuid().ToString("B") would get the same output if I'm not mistaken.

It's not adding brackets, it's just string interpolation.

> I'd be interested in the CPU aspect of this change! The whole "figuring out where my string should start and whether I already have it stored"-part doesn't sound cheap to me.

I haven't done a benchmark yet, but I'm willing to bet it's cheaper than doing all the deserialization beforehand. Offsets and presence bits could be precalculated once for all properties on first access.

> Downside is that the BasicProperty class got even bigger (+8 MB / 1m objects) and that we pay the prize as well when we have the class to send and not when receiving.
> I like the idea in principal, but I think this would make much more sense if we split the receiving property type from the sending property type.

Or we utilize object pooling for the BasicProperties, having a pool of them which we can reset after use. That should almost kill the allocations and coupled with lazy loading, could severely reduce the string allocations overhead as well.

Another option would be to create on "Message" object, where the BasicProperties is a property (instead of a separate field, which makes sense to me) and is also lazy loaded. That'd mean that BasicDeliver methods on consumers only have one parameter containing all the information required, and it can be lazy loaded/deserialized from the memory buffers on-demand. This is for example what Kestrel does with incoming HTTP requests. All the information is stored in the HttpRequest object, and lazy-loaded/parsed as it's accessed for maximum efficiency.

JanEggers:
> Another option would be to create on "Message" object
 
+1

bollhals:
> > I know it's not about the test app, but
> > `header.CorrelationId = $"{Guid.NewGuid()}";`
> > Doesn't that allocate 2 strings? (1 for Guid.NewGuid().ToString() and 1 to put the brackets around? => Guid.NewGuid().ToString("B") would get the same output if I'm not mistaken.
> 
> It's not adding brackets, it's just string interpolation.
> 
woops, silly me, sorry.

> Another option would be to create on "Message" object, where the BasicProperties is a property (instead of a separate field, which makes sense to me) and is also lazy loaded. That'd mean that BasicDeliver methods on consumers only have one parameter containing all the information required, and it can be lazy loaded/deserialized from the memory buffers on-demand. This is for example what Kestrel does with incoming HTTP requests. All the information is stored in the HttpRequest object, and lazy-loaded/parsed as it's accessed for maximum efficiency.

I like the idea of this. Especially the reference to the HttpRequest, people are already familiar to this and therefore it will be easily adoptable.

projects/RabbitMQ.Client/client/framing/BasicProperties.cs
michaelklishin: @stebet I think folks are simply pointing out that there is a local variable naming typo :)

michaelklishin:
@stebet @bollhals @JanEggers is there anything else that we'd like to do in this PR? Now that we have benchmarking project merged into master, perhaps we can adopt it for this change. Anything else?

stebet:
I'd like to go over it again to be sure about a few things perf-wise, also add a few benchmarks while doing so. I'll update it soon-ish (on paternity leave a.t.m) if that's ok? :)

michaelklishin:
@stebet sure. Sorry to bothering you on your paternity leave 🙏👨 👶 🥇

stebet:
> @stebet sure. Sorry to bothering you on your paternity leave 🙏👨 👶 🥇   

No worries :) I have my notifications turned off, just checking in every now and then.

bollhals:
I have a local change that is based on #962 and extends on #990 , which caches a few common strings like QueueName, ExchangeName, ConsumerTag, AppId. This should make the LRUCache part obsolete.

danielmarbach:
Made a review comment on https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/990/files#r551733728 that maybe influences your work sandro

Issue #974, opened 2020-11-12
"Object reference not set to an instance of an object" when connecting using Ssl.Enabled = true


danielsvalefelt:
**Description**
I do not know exactly what happens (I am struggling to understand the SSL settings), but the problem is that, SslHelper.TcpUpgrade throws an exception in SocketFrameHandler constructor, it is calling Close(), but _writerTask is not created yet.

Close calling GetResult without checking.

So, basically, you get a "Object reference not set to an instance of an object" instead of the real exception..

**Solution:**
Change:
_writerTask.GetAwaiter().GetResult(); 
to
_writerTask?.GetAwaiter().GetResult();

michaelklishin:
We cannot conclude whether this will be a "solution" or just mask an exception without understanding what's going on. Please submit pull requests instead of suggesting changes in an issue.

RabbitMQ has a dedicated doc guide that explains how to [narrow down TLS connectivity issues](https://www.rabbitmq.com/troubleshooting-ssl.html), and an extensive [TLS guide](https://www.rabbitmq.com/ssl.html) which explains the basics of TLS, including how peer verification is performed.

There really isn't much else that can be suggested with this much information. We do not guess in this community.

PR #976, opened 2020-11-12
Be more defensive when closing SocketFrameHandler

michaelklishin:
Suggested by @danielsvalefelt in #974. We don't have much context but this makes the change "reviewable" compared to a suggestion comment.

@bollhals @stebet do you see any downsides to doing this? There isn't anything a `SocketFrameHandler` can do to recover when it's being closed, so being more defensive makes sense to me.

### DIFF
--- a/projects/RabbitMQ.Client/client/impl/SocketFrameHandler.cs
+++ b/projects/RabbitMQ.Client/client/impl/SocketFrameHandler.cs
@@ -210,7 +210,7 @@ namespace RabbitMQ.Client.Impl
                     try
                     {
                         _channelWriter.Complete();
-                        _writerTask.GetAwaiter().GetResult();
+                        _writerTask?.GetAwaiter().GetResult();
                     }
                     catch(Exception)
                     {


danielsvalefelt:
I could give you connection string to my server of you want to reproduce it.

Daniel Svalefelt
Svalefelt Consulting LTD

________________________________

stebet:
Makes sense to me :)

bollhals:
I disagree here. The _writerTask is readonly and should never be null (under normal conditions). So adding a null check for the close only makes that every normal close "pays more".
The real issue I see here
```
                try
                {
                    netstream = SslHelper.TcpUpgrade(netstream, endpoint.Ssl);
                }
                catch (Exception)
                {
                    Close(); //<----------------- Issue
                    throw;
                }
```

I think we should just close the stream / socket here instead of calling Close().

danielsvalefelt:
> I disagree here. The _writerTask is readonly and should never be null (under normal conditions). So adding a null check for the close only makes that every normal close "pays more".
> The real issue I see here
> 
> ```
>                 try
>                 {
>                     netstream = SslHelper.TcpUpgrade(netstream, endpoint.Ssl);
>                 }
>                 catch (Exception)
>                 {
>                     Close(); //<----------------- Issue
>                     throw;
>                 }
> ```
> 
> I think we should just close the stream / socket here instead of calling Close().

Normally, I would agree. No reason to call a complete Close() when connection is half open. 
BUT, as we also need to complete the _channelWriter and set the _closed to true and it would be nice to also have some try-catch, so it is a lot of duplicated code..
So, it is much cleaner, and less code to just be more defensive in Close(),

Just my opinion... I have no idea of you coding approaches.

bollhals:
The thing is, the exception raised in the ssl gets rethrown in the catch block, which means that the constructor (which we are in) will not finish. Thus the callee does not have an instance of a sockethandler to call something on. So whats the point of doing more than clean up what we opened in the constructor? 

The ? approach would be all correct if the constructor would succeed, but since it doesn't I think it's clearer to future devs looking at that code and thinking why do we have a null check for a readonly field initialized to not null in the ctor. When we enable nullable reference types in the future, this would be even more confusing, as we'd have a null check on a non nullable field.

danielsvalefelt:
True.
And I do not think I would have put that code in the constructor in the first place, but either set up the connection and handed it over to the constructor when (and if) it is ready, or called a "connect" method once the constructor is done.

But I do not think this minor bug merits a refactoring, so I leave it over to you... I just started to use the code two days ago, and it works well for me (except this).

michaelklishin:
@bollhals so what would you do to avoid exception trace noise in cases of failing TLS connections (which effectively are closed before they are truly open)?

bollhals:
honestly I'd just replace 
```cs
                catch (Exception)
                {
                    Close();
                    throw;
                }
```
with this and call it a day.
```cs
                catch (Exception)
                {
                    // Just close the socket, everything else is not relevant as we're in a constructor and the instance isn't valid anyway after the throw below
                    _socket.Close();
                    throw;
                }
```

PR #978, opened 2020-11-19
Remove unsafe, please

fadi-botros:
## Proposed Changes

Memory safety is the main reason of moving from native code to JVM or CLR,
so I think we must fully unleash memory safety when working in C#, just easily
by not using `unsafe` and disallowing unsafe code in the library.
Performance difference would be very trivial, tried a sample (`ReadShortstr`)
on the .NET benchmark and the performance difference were about 
100 milliseconds when repeating the operation 10000000 times.
So I don't think it deserves using `unsafe`, the unsafe code seems sane but
at least removing unsafe now would make it easier to audit any further unsafe
code that maybe added.

## Types of Changes

## Checklist

## Further Comments

Consider leaving it as an issue and removing the PR if the PR itself is not appropriate

Please note, it seems don't work on old versions of .NET Framework, so either we
should remove support from old versions (like what happens in this PR), or, use
the unsafe only when addressing old frameworks, by adding some precompiler
directives.  I don't think keeping the unsafe just for compatibility for old versions
would be good idea, maybe keep an old version for who needs something old?

### DIFF
--- a/projects/RabbitMQ.Client/RabbitMQ.Client.csproj
+++ b/projects/RabbitMQ.Client/RabbitMQ.Client.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
     <NoWarn>$(NoWarn);CS1591</NoWarn>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <AssemblyTitle>RabbitMQ Client Library for .NET</AssemblyTitle>
--- a/projects/RabbitMQ.Client/RabbitMQ.Client.csproj
+++ b/projects/RabbitMQ.Client/RabbitMQ.Client.csproj
@@ -26,7 +26,6 @@
     <MinVerVerbosity>minimal</MinVerVerbosity>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <PackageOutputPath>..\..\packages</PackageOutputPath>
-    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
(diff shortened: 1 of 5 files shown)


projects/RabbitMQ.Client/RabbitMQ.Client.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
```
sungam3r, line 4: Wow! Why to remove Netstandard?

fadi-botros: Due to most Span related APIs not available in it, and I don't think that we may need to support old .NET Framework versions.
An alternative solution would be to provide the Span API ourselves for `netstandard2.0` as a separate project (yes it would be unsafe), only compilable for `netstandard2.0`, so we won't have any unsafe on the main project, and that in `netcoreapp3.1` we would be using the standard provided APIs that should be safe and we can disable the unsafe in the main project.

michaelklishin:
Thank you for your time. We cannot accept this pull request for a few reasons:

 * It removes support for .NET Standard 2.0. This is potentially a big deal for some users and we haven't discussed when would be a good time to do this.
 * I don't remember seeing any other complaints about the use of this feature; it was adopted before come users would like to optimize the publishing code path as much as possible. It seems to be an acceptable trade-off for most.
 * This PR has three merge commits and "remove unsafe" in four different commits. It's just not a particularly polished contribution, even though we are not particularly strict about such things.

projects/RabbitMQ.Client/RabbitMQ.Client.csproj
sungam3r: netstandard2.0 does not mean `old .NET Framework versions`.

PR #999, opened 2021-01-09
7.0 - implement WaitForConfirmsAsync

### LINKED ISSUE #959: Using blocking publisher confirms with concurrent publishers
Hello.
We're using rabbitmq library version 5 in web applications to publish messages from request. As you may guess it's highly concurrent environment and it's quite common to have 100-500 requests per second sending a message to rabbit.

Trying to update to version 6 of the library (6.2.1 at this time) shows a lot of timeout errors while sending the message. The problem can be reproduced locally and I've attached a simple solution with two projects:

- RabbitMQ (which is using 5.2.0)
- RabbitMQ2 (which is using 6.2.1)

Also I've added simple docker compose file to spin single local rabbitmq node on 127.0.0.11

Each project contains identical code which performs this steps:

1. Opens single (static) connection
2. Creates exchange "tests" (durable with type = topic)
3. Starts 100 parallel tasks trying to publish a message to "test" exchange with publish confirm

Step 3 implemented in 3 different ways and can be switched by passing specific argument when running project:

- (default) always creates new channel for each publish and disposes it
- (per_thread) creates channel per thread (using ThreadStatic)
- (per_thread_with_lock) creates channel per thread (using ThreadStatic) but performs locking so only 1 thread opens up a channel at a time
- (single) only one channel created and every task is under single lock to accessing it

Running all tests on v5 gets me a very good performance - 7-9 msec per message (100 messages total in parallel).
Running all tests on v6 never completed without timeout errors.

Can someone please look into this? Maybe I'm doing something wrong? How to use rabbitmq v6 in high concurrent environment like web?

[RabbitMQ.zip](https://github.com/rabbitmq/rabbitmq-dotnet-client/files/5434284/RabbitMQ.zip)

bollhals:
## Proposed Changes

implements a async WaitForConfirms => does not block a thread anymore while waiting.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #959 )

## Checklist

## Further Comments

This change partially fixes #959. Partially due to it only fixes the wait part and not the channel open part.
(Meaning as long as not all thread pool threads are blocked in opening channels, the publish works as expected. This can be achieved by either precreating channels or pooling the channels or making sure not all threads are started at the same time, so all create a channel at the same time (I'd argue this is already given by real world scenarios))

### DIFF
--- a/projects/RabbitMQ.Client/client/impl/Connection.cs
+++ b/projects/RabbitMQ.Client/client/impl/Connection.cs
@@ -105,7 +105,7 @@ namespace RabbitMQ.Client.Framing.Impl
             _connectionBlockedWrapper = new EventingWrapper<ConnectionBlockedEventArgs>("OnConnectionBlocked", onException);
             _connectionUnblockedWrapper = new EventingWrapper<EventArgs>("OnConnectionUnblocked", onException);
 
-                _sessionManager = new SessionManager(this, 0);
+            _sessionManager = new SessionManager(this, 0);
             _session0 = new MainSession(this) { Handler = NotifyReceivedCloseOk };
             _model0 = (ModelBase)Protocol.CreateModel(_session0);
 
@@ -426,15 +426,6 @@ namespace RabbitMQ.Client.Framing.Impl
             _model0.FinishClose();
         }
 
-        /// <remarks>
-        /// We need to close the socket, otherwise attempting to unload the domain
-        /// could cause a CannotUnloadAppDomainException
-        /// </remarks>
-        public void HandleDomainUnload(object sender, EventArgs ea)
-        {
-            Abort(Constants.InternalError, "Domain Unload");
-        }
-
         public void HandleMainLoopException(ShutdownEventArgs reason)
         {
             if (!SetCloseReason(reason))
@@ -750,7 +741,7 @@ namespace RabbitMQ.Client.Framing.Impl
 
         public void StartMainLoop()
         {
-            _mainLoopTask = Task.Run((Action)MainLoop);
+            _mainLoopTask = Task.Factory.StartNew(MainLoop, TaskCreationOptions.LongRunning);
         }
 
         public void HeartbeatReadTimerCallback(object state)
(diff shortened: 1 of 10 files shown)


projects/RabbitMQ.Client/client/impl/Connection.cs
```diff
@@ -426,15 +426,6 @@ public void FinishClose()
             _model0.FinishClose();
         }
 
-        /// <remarks>
-        /// We need to close the socket, otherwise attempting to unload the domain
-        /// could cause a CannotUnloadAppDomainException
-        /// </remarks>
-        public void HandleDomainUnload(object sender, EventArgs ea)
```
bollhals, line 433: was unused

projects/RabbitMQ.Client/client/impl/Connection.cs
```diff
@@ -750,7 +741,7 @@ public void MaybeStartHeartbeatTimers()
 
         public void StartMainLoop()
         {
-            _mainLoopTask = Task.Run((Action)MainLoop);
+            _mainLoopTask = Task.Factory.StartNew(MainLoop, TaskCreationOptions.LongRunning);
```
bollhals, line 744: This forces the creation of a new thread instead of blocking one of the threadpool

projects/Unit/TestConcurrentAccessWithSharedConnection.cs
```diff
@@ -150,20 +93,9 @@ public void TestConcurrentChannelOpenCloseLoop()
...
         {
-            TestConcurrentChannelOpenAndPublishingWithBodyOfSize(length, 30);
-        }
-
-        internal void TestConcurrentChannelOpenAndPublishingWithBodyOfSize(int length, int iterations)
-        {
-            string s = "payload";
-            if (length > s.Length)
-            {
-                s.PadRight(length);
```
bollhals, line 163: This didn't work as a string is immutable and the returned new string was unused. So it always sent "payload" for all length inputs. => I removed most of the different cases (when I changed it to send the actual size it ran into timeouts due to the huge amount of data it tried to send.)

PR #1003, opened 2021-01-17
add two test applications

bollhals:
## Proposed Changes

adds two test applications to execute when you want to profile (performance or memory) some changes.
I think it would be good to have them in git so everyone has access and we're talking about the same thing.

Apps added:
- Message batch publish / receive (originated from @stebet)
- Open / Close Channels

## Types of Changes

## Checklist


michaelklishin:
We have a benchmarks project now, is this really different from why it was added?

bollhals:
> We have a benchmarks project now, is this really different from why it was added?

If you want to speifically analyze e.g what allocation happend (the images shown in the past PRs) you need an small application to profile. 

The benchmark only yields overall results, but won't give you details about whether or not the thing you tried to eliminate is still allocated. 

Different apps because you may want to look at different things. I've added the batch publish one because that one we used since @stebet used it to track allocations. I've added the channel one because I wanted to see what exactly gets allocated when we crate a channel.

michaelklishin (approved):

PR #1004, opened 2021-01-19
Simplify connection recovery

bollhals:
## Proposed Changes

Simplifies how to recover a connection by

- getting rid of the channel for the recovery commands => replaced by recovery method
- only allocates when a recovery is required => reducing overall startup allocations
- splitting recovery in its own partial class cs file for grouping purpose

## Types of Changes

## Checklist

## Further Comments

I kind of like the split to partial classes for grouping for cases where we can't break up the implementation (big classes like model or connection). If this is not what we want, I can copy them back.


michaelklishin (approved):

michaelklishin:
Thank you!

PR #1008, opened 2021-01-24
Reduce close / abort methods in public API

bollhals:
## Proposed Changes

There are 3 close / abort functions in the public API. This forces a lot of similar implementations in AutorecoveryConnection. 
This PR removes all of them except one in the public API and replaces it with extension methods with the same parameters. (This means a recompile should be enough for customers, if they haven't mocked the interface itself)

## Types of Changes

## Checklist

### DIFF
--- a/projects/RabbitMQ.Client/client/impl/ModelBase.cs
+++ b/projects/RabbitMQ.Client/client/impl/ModelBase.cs
@@ -201,10 +201,10 @@ namespace RabbitMQ.Client.Impl
 
         public void Close(ushort replyCode, string replyText, bool abort)
         {
-            _ = Close(new ShutdownEventArgs(ShutdownInitiator.Application, replyCode, replyText), abort);
+            _ = CloseAsync(new ShutdownEventArgs(ShutdownInitiator.Application, replyCode, replyText), abort);
         }
 
-        private async Task Close(ShutdownEventArgs reason, bool abort)
+        private async Task CloseAsync(ShutdownEventArgs reason, bool abort)
         {
             var k = new ShutdownContinuation();
             ModelShutdown += k.OnConnectionShutdown;
@@ -1355,14 +1355,14 @@ namespace RabbitMQ.Client.Impl
                     return;
                 }
 
-                await Close(
+                await CloseAsync(
                     new ShutdownEventArgs(ShutdownInitiator.Application, Constants.ReplySuccess, "Nacks Received",
                         new IOException("nack received")),
                     false).ConfigureAwait(false);
             }
             catch (TaskCanceledException exception)
             {
-                await Close(new ShutdownEventArgs(ShutdownInitiator.Application,
+                await CloseAsync(new ShutdownEventArgs(ShutdownInitiator.Application,
                         Constants.ReplySuccess,
                         "Timed out waiting for acks",
                         exception),
--- a/projects/RabbitMQ.Client/client/impl/ModelBase.cs
+++ b/projects/RabbitMQ.Client/client/impl/ModelBase.cs
@@ -795,11 +795,8 @@ namespace RabbitMQ.Client.Impl
         {
             if (m_connectionStartCell is null)
             {
-                var reason =
-                    new ShutdownEventArgs(ShutdownInitiator.Library,
-                        Constants.CommandInvalid,
-                        "Unexpected Connection.Start");
-                ((Connection)Session.Connection).Close(reason);
+                var reason = new ShutdownEventArgs(ShutdownInitiator.Library, Constants.CommandInvalid, "Unexpected Connection.Start");
+                Session.Connection.Close(reason, false, Timeout.InfiniteTimeSpan);
             }
             var details = new ConnectionStartDetails
             {
--- a/projects/RabbitMQ.Client/client/impl/ModelBase.cs
+++ b/projects/RabbitMQ.Client/client/impl/ModelBase.cs
@@ -199,14 +199,12 @@ namespace RabbitMQ.Client.Impl
             _recoveryWrapper.Takeover(other._recoveryWrapper);
         }
 
-        public Task Close(ushort replyCode, string replyText, bool abort)
+        public void Close(ushort replyCode, string replyText, bool abort)
         {
-            return Close(new ShutdownEventArgs(ShutdownInitiator.Application,
-                replyCode, replyText),
-                abort);
+            _ = Close(new ShutdownEventArgs(ShutdownInitiator.Application, replyCode, replyText), abort);
         }
 
-        public async Task Close(ShutdownEventArgs reason, bool abort)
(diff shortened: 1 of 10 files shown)


bollhals:
Hmm it works with SDK V 5.0, but build still uses 3.1.x which somehow doesn't like the `this.Abort()` where Abort is an extension method 🤔 
Shall I change it to call the Close method directly, or can we update the SDK?

michaelklishin:
There are some known users to do it. Reducing the minimum required API surface here sounds reasonable to me.

@stebet @bording @danielmarbach @lukebakken WDYT?

michaelklishin:
> Hmm it works with SDK V 5.0, but build still uses 3.1.x which somehow doesn't like the `this.Abort()` where Abort is an extension method 🤔
> Shall I change it to call the Close method directly, or can we update the SDK?

Are there any downsides besides all contributors having to upgrade? I'd say requiring .NET Core 5 for our own development is fine.

bording:
> Reducing the minimum required API surface here sounds reasonable to me.

I'm all for making the Public API simpler, though a quick look at the PR, I'm not quite sure how this PR does that. It seems to just be shuffling the APIs around, not actually reducing anything.

> Are there any downsides besides all contributors having to upgrade? I'd say requiring .NET Core 5 for our own development is fine.

It should be perfectly fine to move to a newer SDK because that has no impact on the consumers of the client.

However, something seems off here. I can't think of a reason why the SDK version would impact whether or not the code compiles.

michaelklishin:
It reduces the part that is mandatory for implementing. We can consider removing the `Abort` overload that provide access to the reply code and message. I mean, if you are wiling to force close and ignore all errors, that context is likely not very relevant anyway.

bollhals:
> > Reducing the minimum required API surface here sounds reasonable to me.
> 
> 
> 
> I'm all for making the Public API simpler, though a quick look at the PR, I'm not quite sure how this PR does that. It seems to just be shuffling the APIs around, not actually reducing anything.
> 
The reduction happens at the IConnection and IModel interface. For convenience and to reduce the update effort the existing api are now extension methods. 

=> interface smaller, easier to implement it while keeping the public facing api usage the same. 

Our benefits are the reduced number of implementation in e.g Connection & AutorecoveringConnection while for the customer nothing changes (except he may calls close now through an extension method)
> 
> > Are there any downsides besides all contributors having to upgrade? I'd say requiring .NET Core 5 for our own development is fine.
> 
> 
> 
> It should be perfectly fine to move to a newer SDK because that has no impact on the consumers of the client.
> 
> 
> 
> However, something seems off here. I can't think of a reason why the SDK version would impact whether or not the code compiles.

It is indeed strange, but I have had it happened to me once or twice that they fixed some bug in a newer SDK. This feels just like it. 

You can look at the code in question. It's a instance method of e.g. Connection, that calls this.Abort. (This implementing an IConnection and Abort being an extension method on IConnection) So this should be fine, but somehow the old SDK doesn't like it. 🤷‍♂

bollhals:
> It reduces the part that is mandatory for implementing. We can consider removing the `Abort` overload that provide access to the reply code and message. I mean, if you are wiling to force close and ignore all errors, that context is likely not very relevant anyway.

Happy to update it if it's wanted. With this change it's now just deleting the extension method 😊
Keep in mind that through the interface close method customer could still call an abort wirh custom code. So we wouldn't prevent it

projects/Unit/APIApproval.Approve.verified.txt
```diff
@@ -325,19 +325,23 @@ namespace RabbitMQ.Client
...
-        void Close(System.TimeSpan timeout);
-        void Close(ushort reasonCode, string reasonText);
-        void Close(ushort reasonCode, string reasonText, System.TimeSpan timeout);
+        void Close(ushort reasonCode, string reasonText, System.TimeSpan timeout, bool abort);
         RabbitMQ.Client.IModel CreateModel();
         void HandleConnectionBlocked(string reason);
         void HandleConnectionUnblocked();
         void UpdateSecret(string newSecret, string reason);
     }
+    public static class IConnectionExtensions
```
danielmarbach, line 334: Did you forget to add and push this file? It seems so and that explains the compilation error

bollhals: Whoooops 😅 you are most certainly correct. Let me add it...
So the SDK update won't be needed then, but could still be done if wanted.

bollhals: api/ was still ignored from when it was generated. this was why it was missing. Thanks for spotting it.

michaelklishin (approved):

michaelklishin:
Thank you!

projects/RabbitMQ.Client/client/impl/ModelBase.cs
```diff
@@ -199,14 +199,12 @@ protected void TakeOver(ModelBase other)
             _recoveryWrapper.Takeover(other._recoveryWrapper);
         }
 
-        public Task Close(ushort replyCode, string replyText, bool abort)
+        public void Close(ushort replyCode, string replyText, bool abort)
         {
-            return Close(new ShutdownEventArgs(ShutdownInitiator.Application,
-                replyCode, replyText),
-                abort);
+            _ = CloseAsync(new ShutdownEventArgs(ShutdownInitiator.Application, replyCode, replyText), abort);
```
danielmarbach, line 204: Is it safe to drop the task here because it means we are no longer materializing any exceptions or waiting for the operation to complete

was this only ever called by the upper layers that are now in the extension methods?

bollhals: Correct. But I agree that this should probably be fixed in another PR. (I didn't want to mix cleanup and fixing in a pr)

bollhals: Also the visibility is why I renamed it to CloseAsync and explicitely dropped the task with _ =. Glad it already worked out😛

projects/RabbitMQ.Client/client/impl/ModelBase.cs
```diff
@@ -1365,14 +1340,14 @@ public async Task WaitForConfirmsOrDieAsync(CancellationToken token = default)
...
-                await Close(
+                await CloseAsync(
                     new ShutdownEventArgs(ShutdownInitiator.Application, Constants.ReplySuccess, "Nacks Received",
                         new IOException("nack received")),
                     false).ConfigureAwait(false);
             }
             catch (TaskCanceledException exception)
             {
-                await Close(new ShutdownEventArgs(ShutdownInitiator.Application,
+                await CloseAsync(new ShutdownEventArgs(ShutdownInitiator.Application,
```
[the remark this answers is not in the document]
bollhals: Doesn't show in the diff. (Should be on the next line)

PR #1016, opened 2021-02-11
Framing benchmarks and optimizations

stebet:
## Proposed Changes

Added benchmarks for frame generation and added optimizations which should benefit all commands sent from the clients.

## Types of Changes

## Checklist

## Further Comments

Main optimizations were done by reducing method calls, adding inlining and reducing code branches. There is definitely room for more improvements, but this is a pretty solid start :)

Benchmark results for serializing `BasicAck`, `BasicDeliver` and `ChannelClose` to their actual frames:
``` ini
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.21277
Intel Core i7-10700 CPU 2.90GHz, 1 CPU, 16 logical and 8 physical cores
.NET Core SDK=5.0.103
  [Host]     : .NET Core 3.1.12 (CoreCLR 4.700.21.6504, CoreFX 4.700.21.6905), X64 RyuJIT
  .NET 4.8 = .NET Framework 4.8 (4.8.4261.0), X64 RyuJIT
  Core 3.1 = .NET Core 3.1.12 (CoreCLR 4.700.21.6504, CoreFX 4.700.21.6905), X64 RyuJIT
```

**Before**
|                  Method |  Runtime |      Mean |    Error |   StdDev | Code Size |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------------ |--------- |----------:|---------:|---------:|----------:|-------:|------:|------:|----------:|
|       ChannelCloseWrite | .NET 4.8 |  82.41 ns | 0.699 ns | 0.653 ns |    1817 B | 0.0088 |     - |     - |      56 B |
|       ChannelCloseWrite | Core 3.1 |  31.48 ns | 0.350 ns | 0.327 ns |    1117 B | 0.0067 |     - |     - |      56 B |
|       BasicPublishWrite | .NET 4.8 | 225.96 ns | 0.905 ns | 0.755 ns |    1817 B | 0.0241 |     - |     - |     152 B |
|       BasicPublishWrite | Core 3.1 | 116.96 ns | 0.684 ns | 0.640 ns |    1117 B | 0.0181 |     - |     - |     152 B |
| BasicPublishMemoryWrite | .NET 4.8 | 162.78 ns | 1.383 ns | 1.226 ns |    1817 B | 0.0241 |     - |     - |     152 B |
| BasicPublishMemoryWrite | Core 3.1 |  71.53 ns | 0.239 ns | 0.200 ns |    1117 B | 0.0181 |     - |     - |     152 B |
|           BasicAckWrite | .NET 4.8 |  70.30 ns | 0.504 ns | 0.472 ns |    1817 B | 0.0088 |     - |     - |      56 B |
|           BasicAckWrite | Core 3.1 |  27.09 ns | 0.508 ns | 0.450 ns |    1117 B | 0.0067 |     - |     - |      56 B |

**After:**
|                  Method |  Runtime |      Mean |   Diff |    Error |   StdDev | Code Size |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------------ |--------- |----------:|-------:|---------:|---------:|----------:|-------:|------:|------:|----------:|
|       ChannelCloseWrite | .NET 4.8 |  69.72 ns | -15,4% | 0.440 ns | 0.411 ns |     759 B | 0.0088 |     - |     - |      56 B |
|       ChannelCloseWrite | Core 3.1 |  23.36 ns | -25,8% | 0.482 ns | 0.691 ns |    1195 B | 0.0067 |     - |     - |      56 B |
|       BasicPublishWrite | .NET 4.8 | 200.58 ns | -11,2% | 0.687 ns | 0.574 ns |    2814 B | 0.0241 |     - |     - |     152 B |
|       BasicPublishWrite | Core 3.1 |  98.16 ns | -16,1% | 0.381 ns | 0.356 ns |    2438 B | 0.0181 |     - |     - |     152 B |
| BasicPublishMemoryWrite | .NET 4.8 | 133.69 ns | -17,9% | 0.336 ns | 0.314 ns |    2942 B | 0.0241 |     - |     - |     152 B |
| BasicPublishMemoryWrite | Core 3.1 |  55.57 ns | -22,3% | 0.503 ns | 0.470 ns |    2438 B | 0.0181 |     - |     - |     152 B |
|           BasicAckWrite | .NET 4.8 |  58.54 ns | -16,7% | 0.245 ns | 0.229 ns |     759 B | 0.0089 |     - |     - |      56 B |
|           BasicAckWrite | Core 3.1 |  18.06 ns | -33.3% | 0.381 ns | 0.558 ns |    1195 B | 0.0067 |     - |     - |      56 B |

### DIFF
--- a/projects/RabbitMQ.Client/client/impl/Frame.cs
+++ b/projects/RabbitMQ.Client/client/impl/Frame.cs
@@ -35,6 +35,7 @@ using System.IO;
 using System.Net.Sockets;
 using System.Runtime.CompilerServices;
 using System.Runtime.ExceptionServices;
+
 using RabbitMQ.Client.Exceptions;
 using RabbitMQ.Util;
 
@@ -47,22 +48,11 @@ namespace RabbitMQ.Client.Impl
          * +------------+---------+----------------+---------+------------------+
          * | 1 byte     | 2 bytes | 4 bytes        | x bytes | 1 byte           |
          * +------------+---------+----------------+---------+------------------+ */
-        private const int BaseFrameSize = 1 + 2 + 4 + 1;
+        internal const int BaseFrameSize = 1 + 2 + 4 + 1;
+        internal const int StartFrameType = 0;
+        internal const int StartChannel = 1;
+        internal const int StartPayloadSize = 3;
         private const int StartPayload = 7;
-
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private static int WriteBaseFrame(Span<byte> span, FrameType type, ushort channel, int payloadLength)
-        {
-            const int StartFrameType = 0;
-            const int StartChannel = 1;
-            const int StartPayloadSize = 3;
-
-            span[StartFrameType] = (byte)type;
-            NetworkOrderSerializer.WriteUInt16(span.Slice(StartChannel), channel);
-            NetworkOrderSerializer.WriteUInt32(span.Slice(StartPayloadSize), (uint)payloadLength);
-            span[StartPayload + payloadLength] = Constants.FrameEnd;
-            return StartPayload + 1 + payloadLength;
-        }
 
         internal static class Method
         {
@@ -75,14 +65,17 @@ namespace RabbitMQ.Client.Impl
              * +----------+-----------+-----------+ */
             public const int FrameSize = BaseFrameSize + 2 + 2;
 
-            public static int WriteTo(Span<byte> span, ushort channel, MethodBase method)
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            public static int WriteTo<T>(Span<byte> span, ushort channel, T method) where T : MethodBase
             {
                 const int StartClassId = StartPayload;
-                const int StartMethodArguments = StartPayload + 4;
+                const int StartMethodArguments = StartClassId + 4;
 
+                int payloadLength = method.WriteArgumentsTo(span.Slice(StartMethodArguments)) + 4;
+                NetworkOrderSerializer.WriteUInt64(span, ((ulong)Constants.FrameMethod << 56) | ((ulong)channel << 40) | ((ulong)payloadLength << 8));
                 NetworkOrderSerializer.WriteUInt32(span.Slice(StartClassId), (uint)method.ProtocolCommandId);
-                int offset = method.WriteArgumentsTo(span.Slice(StartMethodArguments));
-                return WriteBaseFrame(span, FrameType.FrameMethod, channel, StartMethodArguments - StartPayload + offset);
+                span[payloadLength + StartPayload] = Constants.FrameEnd;
+                return payloadLength + BaseFrameSize;
             }
         }
 
@@ -95,18 +88,19 @@ namespace RabbitMQ.Client.Impl
              * +----------+----------+-------------------+-----------+ */
             public const int FrameSize = BaseFrameSize + 2 + 2 + 8;
 
-            public static int WriteTo(Span<byte> span, ushort channel, ContentHeaderBase header, int bodyLength)
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            public static int WriteTo<T>(Span<byte> span, ushort channel, T header, int bodyLength) where T : ContentHeaderBase
             {
                 const int StartClassId = StartPayload;
-                const int StartWeight = StartPayload + 2;
                 const int StartBodyLength = StartPayload + 4;
                 const int StartHeaderArguments = StartPayload + 12;
 
-                NetworkOrderSerializer.WriteUInt16(span.Slice(StartClassId), header.ProtocolClassId);
-                NetworkOrderSerializer.WriteUInt16(span.Slice(StartWeight), 0); // Weight - not used
+                int payloadLength = 12 + header.WritePropertiesTo(span.Slice(StartHeaderArguments));
+                NetworkOrderSerializer.WriteUInt64(span, ((ulong)Constants.FrameHeader << 56) | ((ulong)channel << 40) | ((ulong)payloadLength << 8));
+                NetworkOrderSerializer.WriteUInt32(span.Slice(StartClassId), (uint)header.ProtocolClassId << 16); // The last 16 bytes (Weight) aren't used
                 NetworkOrderSerializer.WriteUInt64(span.Slice(StartBodyLength), (ulong)bodyLength);
-                int offset = header.WritePropertiesTo(span.Slice(StartHeaderArguments));
-                return WriteBaseFrame(span, FrameType.FrameHeader, channel, StartHeaderArguments - StartPayload + offset);
+                span[payloadLength + StartPayload] = Constants.FrameEnd;
+                return payloadLength + BaseFrameSize;
             }
         }
 
@@ -119,12 +113,14 @@ namespace RabbitMQ.Client.Impl
              * +--------------+ */
             public const int FrameSize = BaseFrameSize;
 
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
             public static int WriteTo(Span<byte> span, ushort channel, ReadOnlySpan<byte> body)
             {
                 const int StartBodyArgument = StartPayload;
-
+                NetworkOrderSerializer.WriteUInt64(span, ((ulong)Constants.FrameBody << 56) | ((ulong)channel << 40) | ((ulong)body.Length << 8));
                 body.CopyTo(span.Slice(StartBodyArgument));
-                return WriteBaseFrame(span, FrameType.FrameBody, channel, StartBodyArgument - StartPayload + body.Length);
+                span[StartPayload + body.Length] = Constants.FrameEnd;
+                return body.Length + BaseFrameSize;
             }
         }
 
--- a/projects/RabbitMQ.Client/client/impl/ISession.cs
+++ b/projects/RabbitMQ.Client/client/impl/ISession.cs
@@ -31,6 +31,7 @@
 
 using System;
 using System.Collections.Generic;
+
 using RabbitMQ.Client.Framing.Impl;
 
 namespace RabbitMQ.Client.Impl
@@ -73,8 +74,7 @@ namespace RabbitMQ.Client.Impl
         void Close(ShutdownEventArgs reason, bool notify);
         bool HandleFrame(in InboundFrame frame);
         void Notify();
-        void Transmit(in OutgoingCommand cmd);
-        void Transmit(IList<OutgoingCommand> cmds);
-
+        void Transmit<T>(in T cmd) where T : struct, IOutgoingCommand;
+        void Transmit<T>(List<T> cmds) where T : struct, IOutgoingCommand;
     }
 }
--- a/projects/RabbitMQ.Client/client/impl/ModelBase.cs
+++ b/projects/RabbitMQ.Client/client/impl/ModelBase.cs
@@ -316,7 +316,7 @@ namespace RabbitMQ.Client.Impl
         {
             if (CloseReason is null)
             {
-                 _continuationQueue.Enqueue(k);
+                _continuationQueue.Enqueue(k);
             }
             else
             {
@@ -364,20 +364,13 @@ namespace RabbitMQ.Client.Impl
 
         protected void ModelSend(MethodBase method)
         {
-            ModelSend(method, null, ReadOnlyMemory<byte>.Empty);
+            Session.Transmit(new OutgoingCommand(method));
         }
 
         protected void ModelSend(MethodBase method, ContentHeaderBase header, ReadOnlyMemory<byte> body)
         {
-            if (method.HasContent)
-            {
-                _flowControlBlock.Wait();
-                Session.Transmit(new OutgoingCommand(method, header, body));
-            }
-            else
-            {
-                Session.Transmit(new OutgoingCommand(method, header, body));
-            }
+            _flowControlBlock.Wait();
+            Session.Transmit(new OutgoingContentCommand(method, header, body));
         }
 
         internal void OnCallbackException(CallbackExceptionEventArgs args)
@@ -1234,7 +1227,7 @@ namespace RabbitMQ.Client.Impl
             }
         }
 
-        internal void SendCommands(IList<OutgoingCommand> commands)
+        internal void SendCommands(List<OutgoingContentCommand> commands)
         {
             _flowControlBlock.Wait();
             if (NextPublishSeqNo > 0)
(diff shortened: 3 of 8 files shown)


projects/RabbitMQ.Client/client/impl/OutgoingCommand.cs
```diff
@@ -1,4 +1,4 @@
-// This source code is dual-licensed under the Apache License, version
+// This ource code is dual-licensed under the Apache License, version
```
danielmarbach, line 1: This should probably be reverted

projects/RabbitMQ.Client/client/impl/ModelBase.cs
```diff
@@ -367,20 +367,13 @@ public void HandleCommand(in IncomingCommand cmd)
...
-            if (method.HasContent)
-            {
-                _flowControlBlock.Wait();
-                Session.Transmit(new OutgoingCommand(method, header, body));
-            }
-            else
-            {
-                Session.Transmit(new OutgoingCommand(method, header, body));
-            }
+            _flowControlBlock.Wait();
```
danielmarbach, line 375: What is the impact of this change due to always waiting on the flow control?

projects/RabbitMQ.Client/client/impl/OutgoingCommand.cs
```diff
@@ -31,49 +31,75 @@
...
+
+        public OutgoingContentCommand(MethodBase method, ContentHeaderBase header, ReadOnlyMemory<byte> body)
+        {
+            _method = method;
             _header = header;
             _body = body;
         }
 
-        internal void Transmit(ushort channelNumber, Connection connection)
+        public ReadOnlyMemory<byte> Transmit(ushort channelNumber, uint frameMax)
```
danielmarbach, line 86: Is transmit still a good name since we don't really transmit but more Prepare or create the payload

projects/RabbitMQ.Client/client/impl/SessionBase.cs
```diff
@@ -32,6 +32,7 @@
 using System;
 using System.Collections.Generic;
 using System.Threading;
+
```
danielmarbach, line 35: Is this intentional?

projects/RabbitMQ.Client/client/impl/ISession.cs
```diff
@@ -73,8 +74,7 @@ internal interface ISession
         void Close(ShutdownEventArgs reason, bool notify);
         bool HandleFrame(in InboundFrame frame);
         void Notify();
-        void Transmit(in OutgoingCommand cmd);
-        void Transmit(IList<OutgoingCommand> cmds);
-
+        void Transmit<T>(in T cmd) where T : struct, IOutgoingCommand;
+        void Transmit<T>(IList<T> cmds) where T : struct, IOutgoingCommand;
```
danielmarbach, line 78: Given that we never modify the commands would it make sense to use a concrete read-only structure directly and avoid any interface dispatches?

projects/RabbitMQ.Client/client/impl/SessionBase.cs
```diff
@@ -135,14 +136,14 @@ public virtual void Transmit(in OutgoingCommand cmd)
...
+            Connection.Write(cmd.Transmit(ChannelNumber, Connection.FrameMax));
         }
 
-        public virtual void Transmit(IList<OutgoingCommand> cmds)
+        public virtual void Transmit<T>(IList<T> cmds) where T : struct, IOutgoingCommand
         {
             for (int i = 0; i < cmds.Count; i++)
             {
-                cmds[i].Transmit(ChannelNumber, Connection);
+                Connection.Write(cmds[i].Transmit(ChannelNumber, Connection.FrameMax));
```
danielmarbach, line 146: Probably not going to make a huge difference but technically we don't need to reacquire the FrameMax inside the loop all the time right?

projects/RabbitMQ.Client/client/impl/ModelBase.cs
stebet: It was always called, since this was only called with methods that had HasContent = true. The method above is for non-content methods :)

projects/RabbitMQ.Client/client/impl/OutgoingCommand.cs
stebet: Yep

projects/RabbitMQ.Client/client/impl/OutgoingCommand.cs
stebet: Yeah, something like `SerializeToFrames` would probably be more descriptive.

projects/RabbitMQ.Client/client/impl/SessionBase.cs
stebet: VS autoformatter added this. It's set to split usings between base-namespaces. I can remove if wanted.

projects/RabbitMQ.Client/client/impl/SessionBase.cs
stebet: Good catch.

projects/RabbitMQ.Client/client/impl/ISession.cs
stebet: The IList one is just used for batch publishes. So the list is populated first but we might be able to just hardcode it to use a List<T>?

michaelklishin: If this list is instantiated by the client, yes. If it is user-provided, I don't think so.

bollhals:
Nice improvements! Looks very much worth it.

Some of your improvements I also applied in #962, so we should look how we coordinate these two PRs to get the maximum out of it 🔥 🚀 

Out of curiosity, while changing it, what change did bring the most benefit? Is shifting & then write as ulong that much faster?

projects/RabbitMQ.Client/client/impl/Frame.cs
```diff
@@ -75,14 +65,17 @@ internal static class Method
...
-            public static int WriteTo(Span<byte> span, ushort channel, MethodBase method)
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            public static int WriteTo<T>(Span<byte> span, ushort channel, T method) where T : MethodBase
             {
                 const int StartClassId = StartPayload;
-                const int StartMethodArguments = StartPayload + 4;
+                const int StartMethodArguments = StartClassId + 4;
 
+                int payloadLength = method.WriteArgumentsTo(span.Slice(StartMethodArguments)) + 4;
+                NetworkOrderSerializer.WriteUInt64(span, ((ulong)Constants.FrameMethod << 56) | ((ulong)channel << 40) | ((ulong)payloadLength << 8));
```
bollhals, line 75: I assume this is repeated as it is not inlineable if it is in its own method?

projects/RabbitMQ.Client/client/impl/OutgoingCommand.cs
```diff
@@ -31,49 +31,75 @@
...
         {
-            Method = method;
+            int size = Method.GetRequiredBufferSize() + Framing.Method.FrameSize;
+
+            // Will be returned by SocketFrameWriter.WriteLoop
+            byte[] rentedArray = ArrayPool<byte>.Shared.Rent(size);
+            Span<byte> span = rentedArray.AsSpan(0, size);
+            Framing.Method.WriteTo(span, channelNumber, Method);
+#if DEBUG
+            if (offset != size)
```
bollhals, line 62: offset is not declared?

projects/RabbitMQ.Client/client/impl/OutgoingCommand.cs
```diff
@@ -83,34 +109,14 @@ internal void Transmit(ushort channelNumber, Connection connection)
                 }
             }
 
+#if DEBUG
```
bollhals, line 112: Maybe we'd want to use `Debug.Assert(offset != size, $"Serialized to wrong size, expect {size}, offset {offset}");` instead of `#if DEBUG`

projects/RabbitMQ.Client/client/impl/Frame.cs
```diff
@@ -119,12 +113,14 @@ internal static class BodySegment
              * +--------------+ */
             public const int FrameSize = BaseFrameSize;
 
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
             public static int WriteTo(Span<byte> span, ushort channel, ReadOnlySpan<byte> body)
             {
                 const int StartBodyArgument = StartPayload;
-
+                span[StartPayload + body.Length] = Constants.FrameEnd;
```
bollhals, line 120: If this is before the body.CopyTo, then we  could potentially overwrite it. I'd suggest moving it after, so also to be in line with the other WriteTo methods.

projects/RabbitMQ.Client/client/impl/Frame.cs
```diff
@@ -47,23 +48,12 @@ internal static class Framing
          * +------------+---------+----------------+---------+------------------+
          * | 1 byte     | 2 bytes | 4 bytes        | x bytes | 1 byte           |
          * +------------+---------+----------------+---------+------------------+ */
-        private const int BaseFrameSize = 1 + 2 + 4 + 1;
+        internal const int BaseFrameSize = 1 + 2 + 4 + 1;
+        const int StartFrameType = 0;
```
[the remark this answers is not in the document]
stebet: Will do

projects/RabbitMQ.Client/client/impl/OutgoingCommand.cs
stebet: Good catch

projects/RabbitMQ.Client/client/impl/OutgoingCommand.cs
stebet: Good idea

projects/RabbitMQ.Client/client/impl/Frame.cs
stebet: Yeah. This was an attempt to reduce bounds checks but should be unnecessary. I'll move it back.

stebet:
The UInt64 change helped get rid of method calls and calls to Span.Slice, which even if they are cheap, add up quickly on a hot-path.

The biggest factor was splitting up `OutgoingCommand` to reduce code branches (ifs) and extra method calls.

I have an idea for a few more optimizations which I'll try out tomorrow.

projects/RabbitMQ.Client/client/impl/OutgoingCommand.cs
```diff
@@ -31,49 +31,75 @@
...
+        public MethodBase Method => _method;
 
         public OutgoingCommand(MethodBase method)
-            : this(method, null, ReadOnlyMemory<byte>.Empty)
         {
+            _method = method;
         }
 
-        public OutgoingCommand(MethodBase method, ContentHeaderBase header, ReadOnlyMemory<byte> body)
+        public ReadOnlyMemory<byte> SerializeToFrames(ushort channelNumber, uint frameMax)
```
bollhals, line 53: now that I think of it, shouldn't this method be declared readonly? Otherwise calling it will have to create a defensive copy before the call.

bollhals: scratch that, the whole struct is readonly ... sorry

projects/RabbitMQ.Client/client/impl/OutgoingCommand.cs
```diff
@@ -31,49 +31,75 @@
...
         {
-            Method = method;
+            int size = Method.GetRequiredBufferSize() + Framing.Method.FrameSize;
+
+            // Will be returned by SocketFrameWriter.WriteLoop
+            byte[] rentedArray = ArrayPool<byte>.Shared.Rent(size);
+            Span<byte> span = rentedArray.AsSpan(0, size);
```
bollhals, line 59: technically we don't need the span, we can just create the span over the whole array, since we don't check the size at the end, we also don't need to limit it.

projects/RabbitMQ.Client/client/impl/ISession.cs
stebet: Changed to concrete type since it isn't user provided.

projects/RabbitMQ.Client/client/impl/Frame.cs
stebet: It's just a one-liner, so not worried about it.

projects/RabbitMQ.Client/client/impl/OutgoingCommand.cs
stebet: Fixed

stebet:
Rebased and squashed after reviews.

projects/RabbitMQ.Client/client/impl/OutgoingCommand.cs
```diff
@@ -31,49 +31,74 @@
...
-            int size = GetMaxSize(maxBodyPayloadBytes);
+            int maxBodyPayloadBytes = (int)(frameMax == 0 ? int.MaxValue : frameMax - Framing.BaseFrameSize);
+            int size = Method.GetRequiredBufferSize() +
+                _header.GetRequiredPayloadBufferSize() +
+                Framing.BodySegment.FrameSize * (maxBodyPayloadBytes == int.MaxValue ? 1 : (_body.Length + maxBodyPayloadBytes - 1) / maxBodyPayloadBytes) + _body.Length
+                + Framing.Method.FrameSize + Framing.Header.FrameSize;
 
             // Will be returned by SocketFrameWriter.WriteLoop
             byte[] rentedArray = ArrayPool<byte>.Shared.Rent(size);
             Span<byte> span = rentedArray.AsSpan(0, size);
```
bollhals, line 95: Also here, we can directly pass it into the method (or use rentedArray.AsSpan(offset)) => safes one local variable

stebet:
Updated with latest perf numbers

projects/RabbitMQ.Client/client/impl/OutgoingCommand.cs
stebet: Not sure I follow

stebet: Fixed

projects/RabbitMQ.Client/client/impl/OutgoingCommand.cs
bollhals: The #if Debug can be removed now, as the Debug.Assert is only compiled for Debug anways.

stebet:
Anyone want to do a final review of this? If not, I think this PR is ready :) @michaelklishin @danielmarbach @bollhals

bollhals:
Go for it. I'll update #962 afterwards to pickup these improvements and extend where possible.

danielmarbach (approved):

PR #1017, opened 2021-02-18
Code cleanups and performance optimizations.

stebet:
## Proposed Changes

Removing some redundant code and some performance optimizations.

## Types of Changes

## Checklist

### DIFF
--- a/projects/RabbitMQ.Client/client/impl/Frame.cs
+++ b/projects/RabbitMQ.Client/client/impl/Frame.cs
@@ -165,23 +165,22 @@ namespace RabbitMQ.Client.Impl
             _rentedArray = rentedArray;
         }
 
-        private static void ProcessProtocolHeader(Stream reader)
+        private static void ProcessProtocolHeader(Stream reader, ReadOnlySpan<byte> frameHeader)
         {
             try
             {
-                byte b1 = (byte)reader.ReadByte();
-                byte b2 = (byte)reader.ReadByte();
-                byte b3 = (byte)reader.ReadByte();
-                if (b1 != 'M' || b2 != 'Q' || b3 != 'P')
+                if (frameHeader[0] != 'M' || frameHeader[1] != 'Q' || frameHeader[2] != 'P')
                 {
                     throw new MalformedFrameException("Invalid AMQP protocol header from server");
                 }
 
-                int transportHigh = reader.ReadByte();
-                int transportLow = reader.ReadByte();
-                int serverMajor = reader.ReadByte();
                 int serverMinor = reader.ReadByte();
-                throw new PacketNotRecognizedException(transportHigh, transportLow, serverMajor, serverMinor);
+                if (serverMinor == -1)
+                {
+                    throw new EndOfStreamException();
+                }
+
+                throw new PacketNotRecognizedException(frameHeader[3], frameHeader[4], frameHeader[5], reader.ReadByte());
             }
             catch (EndOfStreamException)
             {
@@ -198,39 +197,37 @@ namespace RabbitMQ.Client.Impl
 
         internal static InboundFrame ReadFrom(Stream reader, byte[] frameHeaderBuffer)
         {
-            int type = default;
             try
             {
-                type = reader.ReadByte();
+                if (reader.Read(frameHeaderBuffer, 0, 7) == 0)
+                {
+                    throw new EndOfStreamException("Reached the end of the stream. Possible authentication failure.");
+                }
             }
             catch (IOException ioe)
             {
                 // If it's a WSAETIMEDOUT SocketException, unwrap it.
                 // This might happen when the limit of half-open connections is
                 // reached.
-                if (ioe.InnerException is null ||
-                    !(ioe.InnerException is SocketException exception) ||
-                    exception.SocketErrorCode != SocketError.TimedOut)
+                if (ioe?.InnerException is SocketException exception && exception.SocketErrorCode == SocketError.TimedOut)
+                {
+                    ExceptionDispatchInfo.Capture(exception).Throw();
+                }
+                else
                 {
                     throw;
                 }
-
-                ExceptionDispatchInfo.Capture(ioe.InnerException).Throw();
             }
 
-            switch (type)
+            if (frameHeaderBuffer[0] == 'A')
             {
-                case -1:
-                    throw new EndOfStreamException("Reached the end of the stream. Possible authentication failure.");
-                case 'A':
-                    // Probably an AMQP protocol header, otherwise meaningless
-                    ProcessProtocolHeader(reader);
-                    break;
+                // Probably an AMQP protocol header, otherwise meaningless
+                ProcessProtocolHeader(reader, frameHeaderBuffer.AsSpan(1, 6));
             }
 
-            reader.Read(frameHeaderBuffer, 0, frameHeaderBuffer.Length);
-            int channel = NetworkOrderDeserializer.ReadUInt16(new ReadOnlySpan<byte>(frameHeaderBuffer));
-            int payloadSize = NetworkOrderDeserializer.ReadInt32(new ReadOnlySpan<byte>(frameHeaderBuffer, 2, 4)); // FIXME - throw exn on unreasonable value
+            FrameType type = (FrameType)frameHeaderBuffer[0];
+            int channel = NetworkOrderDeserializer.ReadUInt16(new ReadOnlySpan<byte>(frameHeaderBuffer, 1, 2));
+            int payloadSize = NetworkOrderDeserializer.ReadInt32(new ReadOnlySpan<byte>(frameHeaderBuffer, 3, 4)); // FIXME - throw exn on unreasonable value
 
             const int EndMarkerLength = 1;
             // Is returned by InboundFrame.ReturnPayload in Connection.MainLoopIteration
@@ -257,7 +254,7 @@ namespace RabbitMQ.Client.Impl
                 throw new MalformedFrameException($"Bad frame end marker: {payloadBytes[payloadSize]}");
             }
 
-            return new InboundFrame((FrameType)type, channel, new Memory<byte>(payloadBytes, 0, payloadSize), payloadBytes);
+            return new InboundFrame(type, channel, new Memory<byte>(payloadBytes, 0, payloadSize), payloadBytes);
         }
 
         public byte[] TakeoverPayload()
--- a/projects/RabbitMQ.Client/client/impl/Frame.cs
+++ b/projects/RabbitMQ.Client/client/impl/Frame.cs
@@ -180,7 +180,7 @@ namespace RabbitMQ.Client.Impl
                     throw new EndOfStreamException();
                 }
 
-                throw new PacketNotRecognizedException(frameHeader[3], frameHeader[4], frameHeader[5], reader.ReadByte());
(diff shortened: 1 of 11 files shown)


stebet:
@bollhals might want to take a look as well.

projects/RabbitMQ.Client/client/impl/Frame.cs
```diff
@@ -169,23 +170,22 @@ private InboundFrame(FrameType type, int channel, ReadOnlyMemory<byte> payload,
...
-                int transportLow = reader.ReadByte();
-                int serverMajor = reader.ReadByte();
                 int serverMinor = reader.ReadByte();
-                throw new PacketNotRecognizedException(transportHigh, transportLow, serverMajor, serverMinor);
+                if (serverMinor == -1)
+                {
+                    throw new EndOfStreamException();
+                }
+
+                throw new PacketNotRecognizedException(frameHeader[3], frameHeader[4], frameHeader[5], reader.ReadByte());
```
bollhals, line 188: shouldn't that be serverMinor?

projects/RabbitMQ.Client/client/impl/MainSession.cs
```diff
@@ -44,50 +46,42 @@ namespace RabbitMQ.Client.Impl
...
-            lock (_closingLock)
-            {
-                if (!_closing)
-                {
-                    return base.HandleFrame(in frame);
-                }
-            }
-
-            if (!_closeServerInitiated && frame.Type == FrameType.FrameMethod)
+            if (_closing)
```
bollhals, line 60: can we switch the common case first?

projects/RabbitMQ.Client/client/impl/MainSession.cs
```diff
@@ -44,50 +46,42 @@ namespace RabbitMQ.Client.Impl
     internal sealed class MainSession : Session
     {
         private readonly object _closingLock = new object();
```
bollhals, line 48: Could this be optimized out?

projects/RabbitMQ.Client/client/impl/SessionBase.cs
```diff
@@ -126,13 +127,11 @@ public void Notify()
 
         public virtual void Transmit(in OutgoingCommand cmd)
         {
-            if (CloseReason != null)
+            if (cmd.Method.ProtocolCommandId != client.framing.ProtocolCommandId.ChannelCloseOk && !IsOpen)
```
bollhals, line 130: Why swap the commandId first? The common case is that they're not equal and thus have to check !IsOpen anyway.

stebet:
> Looking good to me. Question is though, what will remain in Protocol class? It seems to slowly die out.

Yes. The Protocol class seems redundant to me. I can see what the point of it was but extending the client with a new protocol other than AMQP 0.9.8 is a much bigger task than just adding a new protocol.

Is there a reason to keep the protocol abstraction around @michaelklishin? Seems like it'd be easier to write a new client for a new protocol than to support more than one in the current client.

projects/RabbitMQ.Client/client/impl/SessionBase.cs
stebet: Makes sense. I was going for the cheaper case first (instead of the more common) which doesn't make sense since it'll almost always be hit anyway.

bollhals:
> > Looking good to me. Question is though, what will remain in Protocol class? It seems to slowly die out.
> 
> 
> 
> Yes. The Protocol class seems redundant to me. I can see what the point of it was but extending the client with a new protocol other than AMQP 0.9.8 is a much bigger task than just adding a new protocol.
> 
> 
> 
> Is there a reason to keep the protocol abstraction around @michaelklishin? Seems like it'd be easier to write a new client for a new protocol than to support more than one in the current client.

Sounds reasonable to me.

projects/RabbitMQ.Client/client/impl/Frame.cs
```diff
@@ -202,39 +202,37 @@ private static void ProcessProtocolHeader(Stream reader)
...
-                    ProcessProtocolHeader(reader);
-                    break;
+                // Probably an AMQP protocol header, otherwise meaningless
+                ProcessProtocolHeader(reader, frameHeaderBuffer.AsSpan(1, 6));
             }
 
-            reader.Read(frameHeaderBuffer, 0, frameHeaderBuffer.Length);
-            int channel = NetworkOrderDeserializer.ReadUInt16(new ReadOnlySpan<byte>(frameHeaderBuffer));
-            int payloadSize = NetworkOrderDeserializer.ReadInt32(new ReadOnlySpan<byte>(frameHeaderBuffer, 2, 4)); // FIXME - throw exn on unreasonable value
+            FrameType type = (FrameType)frameHeaderBuffer[0];
```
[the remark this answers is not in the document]
stebet: Made it simpler :)

projects/RabbitMQ.Client/client/impl/Frame.cs
```diff
@@ -202,39 +202,37 @@ private static void ProcessProtocolHeader(Stream reader)
 
         internal static InboundFrame ReadFrom(Stream reader, byte[] frameHeaderBuffer)
         {
-            int type = default;
             try
             {
-                type = reader.ReadByte();
+                if (reader.Read(frameHeaderBuffer, 0, 7) == 0)
```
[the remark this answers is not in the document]
stebet: Done

projects/RabbitMQ.Client/client/impl/Frame.cs
stebet: Yes. Fixed.

projects/RabbitMQ.Client/client/impl/MainSession.cs
stebet: That'll involve doing a lot more checks really. This just does the first check on `if _closing` which is almost always false and then just continues to the transmit. This usually compiles to no code jumping unless the if matches which is usually faster.

projects/RabbitMQ.Client/client/impl/MainSession.cs
stebet: It could perhaps be replace with an Interlocked.Exchange, but this is called so rarely that I don't think it's necessary. I did add another if check before the lock though which should save us from taking the lock most of the time.

The lock is only taken when we are trying to close the Session so it's rare anywas.

projects/RabbitMQ.Client/client/impl/MainSession.cs
michaelklishin: I expect this to be cheaper than acquiring a lock either way ;)

michaelklishin:
This looks reasonable, thank you. Any benchmarks that demonstrate the improvement for some key workloads?

projects/RabbitMQ.Client/client/impl/MainSession.cs
bollhals: This for sure =) 
A quick "is null" check might be as fast as a enum compare 🤔

projects/RabbitMQ.Client/client/impl/MainSession.cs
bollhals: We could also lock on the session object or anything else, so we do not require to allocate an additional object for this rare case.

stebet: Yeah, it's an internal object so we can make sure ourselves noone else is taking that lock. I'll do that instead.

projects/RabbitMQ.Client/client/impl/MainSession.cs
stebet: > This looks reasonable, thank you. Any benchmarks that demonstrate the improvement for some key workloads?

I'll see if I can do some benchmarks. They are a bit harder to do here since they require actual usages of Connection/Model objects.

stebet:
> This looks reasonable, thank you. Any benchmarks that demonstrate the improvement for some key workloads?

I'll see if I can do some benchmarks. They are a bit harder to do here since they require actual usages of Connection/Model objects.

projects/RabbitMQ.Client/client/impl/MainSession.cs
```diff
@@ -98,35 +91,31 @@ public override bool HandleFrame(in InboundFrame frame)
         ///</remarks>
         public void SetSessionClosing(bool closeServerInitiated)
         {
-            lock (_closingLock)
+            if (!_closing)
             {
-                if (!_closing)
+                lock (this)
```
danielmarbach, line 96: In general it is not a good practice to lock on this since this can lead to deadlocks because your are locking the whole instance on which someone external could also lock on. Either bring the lock instance object back or use a compare exchange spin approach?

bollhals: Sessions are internal objects not exposed in any way to the public. While I agree with the general statement, I think one does not always have to follow it if the lock from outside is not possible.

stebet: Yeah, I know locking on the instance is not recommended but in this case we can be sure that no one external will lock on it, unless through reflection gymnastics but all bets are off at that point anyway.

danielmarbach: I'm a fan of least surprises. Even for internal usage. It doesn't necessarily have to be user code. Sometimes over time things slip in. Having a dedicated lock is the safer choice. Yet I'm not hung up on it. Feel free to ignore

stebet:
Rudimentary benchmarks for unthrottled publishing with 8 connections:
**Before:**
Sent 18663500 messages in 60044 ms. 311058 messages per sec.

Calls to ManualResetEventSlim.Wait():
image

**After:**
Sent 18683500 messages in 60036 ms. 311391 messages per sec.

Calls to ManualResetEventSlim.Wait():
image

projects/RabbitMQ.Client/client/impl/MainSession.cs
michaelklishin: We can add a new field just for locking. One more object per connection is not going to make much of a difference.

michaelklishin: Should we do this?

michaelklishin:
Given that the gain is 0,1% or so, do we want to proceed with this solely for code simplification (if any) or just close it? We obviously appreciate the time @stebet and the reviewers have put into this PR. But if an optimization PR doesn't make any difference to benchmarks in the end, is it worth adopting?

stebet:
> Given that the gain is 0,1% or so, do we want to proceed with this solely for code simplification (if any) or just close it? We obviously appreciate the time @stebet and the reviewers have put into this PR. But if an optimization PR doesn't make any difference to benchmarks in the end, is it worth adopting?

I'm biased, this being my PR and all, but from a code cleaning perspective I'd say yes. The performance optimization was purely a bonus and not the main point of the PR, although less locking is always better too. Although throughput gains were minimal, CPU usage also dropped a bit as a result.

michaelklishin:
> The Protocol class seems redundant to me

At this point, we can just drop it and leave only 0-9-1 as supported. 0-8 and 0-9 share a lot with 0-9-1 (unlike 1.0, which is a completely different protocol), so years ago it made sense to support them all.

0-9-1 is available in RabbitMQ 2.0+, which has been out for more than a decade. So we can remove all 0-8 and 0-9 bits (I don't expect there to be many) in a separate PR.

michaelklishin (approved):

PR #1018, opened 2021-02-20
remove socket.poll

bollhals:
## Proposed Changes

Removes the call to [Socket.Poll](https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.socket.poll?view=net-5.0) as to my knowledge this shouldn't be needed, especially because we don't check for the result...

## Types of Changes

## Checklist

### DIFF
--- a/projects/RabbitMQ.Client/client/impl/SocketFrameHandler.cs
+++ b/projects/RabbitMQ.Client/client/impl/SocketFrameHandler.cs
@@ -35,7 +35,6 @@ using System.IO;
 using System.Net;
 using System.Net.Sockets;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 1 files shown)


michaelklishin:
The poll was introduced in https://github.com/rabbitmq/rabbitmq-dotnet-client/commit/d74b821f917230613e45cd39771c76312678e401. @stebet do you agree to remove it?

stebet:
It wasn't introduced there. Simply moved from the code elsewhere. Since we only have one thread writing to the socket, I don't think the poll is needed.

michaelklishin (approved):

michaelklishin:
Great, thank you both for walking me through this.

PR #1026, opened 2021-03-09
Workaround: Fixing analyzer build errors on VS2019 v16.9

stebet:
## Proposed Changes

VS 2019 automatically enables nullable checks and some analyzers for projects. To avoid builds failing when developing locally I disabled the .NET analyzers for the time being until they can be reviewed and fixed properly.

Would be nice if someone can confirm this behavior before and after: @bollhals @danielmarbach 

## Types of Changes

## Checklist

### DIFF
--- a/projects/RabbitMQ.Client/RabbitMQ.Client.csproj
+++ b/projects/RabbitMQ.Client/RabbitMQ.Client.csproj
@@ -33,7 +33,6 @@
 
   <PropertyGroup>
     <AnalysisMode>AllEnabledByDefault</AnalysisMode>
-    <RunAnalyzersDuringBuild>false</RunAnalyzersDuringBuild>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(CONCOURSE_CI_BUILD)' == 'true'">
--- a/projects/RabbitMQ.Client/RabbitMQ.Client.csproj
+++ b/projects/RabbitMQ.Client/RabbitMQ.Client.csproj
@@ -31,10 +31,6 @@
     <ReleaseVersion>8.0</ReleaseVersion>
   </PropertyGroup>
 
-  <PropertyGroup>
-    <AnalysisMode>AllEnabledByDefault</AnalysisMode>
-  </PropertyGroup>
-
   <PropertyGroup Condition="'$(CONCOURSE_CI_BUILD)' == 'true'">
     <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
     <Deterministic>true</Deterministic>
@@ -67,6 +63,7 @@
     <PackageReference Include="MinVer" Version="2.4.0" PrivateAssets="All" />
     <PackageReference Include="System.Memory" Version="4.5.4" />
     <PackageReference Include="System.Threading.Channels" Version="5.0.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="5.0.3" />
   </ItemGroup>
 
 </Project>


danielmarbach:
@bording has much more context in this area. I defer to him

stebet:
B.t.w, I'm not saying we shouldn't fix the issues. Just workaround it for now and create PRs later with proper fixes for them. These analysers are also automatically enabled on net5.0 projects, so it'd be a good preparation to fix them properly.

projects/RabbitMQ.Client/RabbitMQ.Client.csproj
```diff
@@ -33,6 +33,7 @@
 
   <PropertyGroup>
     <AnalysisMode>AllEnabledByDefault</AnalysisMode>
```
bording, line 35: This is the most extreme mode that the analyzers can be set to. Only a handful of rules are enabled by default, so this is opting in to all of them.

projects/RabbitMQ.Client/RabbitMQ.Client.csproj
```diff
@@ -33,6 +33,7 @@
...
+    <EnableNETAnalyzers>false</EnableNETAnalyzers>
```
bording, line 36: This would a way to globally disable the analyzers, though since this project isn't targeting `net5.0`, they are already disabled by default. Seems like this is redundant at the moment.

projects/RabbitMQ.Client/RabbitMQ.Client.csproj
stebet: Thanks for this. I'll take a closer look tomorrow and see if I can get it back to the previous behaviour.

projects/RabbitMQ.Client/RabbitMQ.Client.csproj
stebet: I removed this line so they aren't all disabled now.

projects/RabbitMQ.Client/RabbitMQ.Client.csproj
```diff
@@ -67,7 +63,7 @@
     <PackageReference Include="MinVer" Version="2.4.0" PrivateAssets="All" />
     <PackageReference Include="System.Memory" Version="4.5.4" />
     <PackageReference Include="System.Threading.Channels" Version="5.0.0" />
-    <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="5.0.1" />
+    <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="5.0.3" />
```
bording, line 66: Do we know what version of the .NET SDK we're using to build against in CI? If we can just update to a .NET 5 SDK, I'd prefer to do that over explicitly adding a package reference to the analyzer.

Then we just need to have `<EnableNETAnalyzers>true</EnableNETAnalyzers>` in the project file to enable the default set of analyzers to run against this project.

michaelklishin: RabbitMQ's Concourse uses

```
dotnet-sdk-3.1
dotnet-runtime-2.1
```

I could not get .NET 5 build 3.1 apps without surprising failures (I no longer remember the details). I'd be all for moving to use .NET 5 (but target .NET Core 3.1+ for now) but perhaps it should be a separate PR.

michaelklishin:
@stebet @bording @bollhals I'm OK with merging this and perhaps investigating how/why we should switch to using .NET 5 for development and CI next. It was not straightforward to use .NET 5 and target 3.1 on a Mac last time I tried it but it easily could have been my lack of experience with .NET.

stebet:
> @stebet @bording @bollhals I'm OK with merging this and perhaps investigating how/why we should switch to using .NET 5 for development and CI next. It was not straightforward to use .NET 5 and target 3.1 on a Mac last time I tried it but it easily could have been my lack of experience with .NET.

Sounds good

stebet:
@michaelklishin rebase done :)

michaelklishin:
Thank you!

PR #1040, opened 2021-04-22
Resolve #1039 Confirms Failure should use ShutdownInitiator.Library

### LINKED ISSUE #1039: ModelBase.WaitForConfirmsOrDieAsync uses ShutdownInitiator.Application rather than Library
I am using logic similar to that mentioned in #159 to recover connections, and it has worked great except for a recent case. The logic is:
```
model.ModelShutdown += (sender, args) =>
{
    if (args.Initiator == ShutdownInitiator.Application)
        return; //We closed it, do not attempt to recover.
    // Library/Peer closed it so recover. [Code snipped]
}; 
```
However, the code in `ModelBase.WaitForConfirmsOrDieAsync` does this:
https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/66dc8ca4b18b3fcd2c5b28623a9109c4e177c468/projects/RabbitMQ.Client/client/impl/ModelBase.cs#L1098-L1121

Since it is using `ShutdownInitiator.Application`, my autorecovery code does not fire. I would expect the RabbitMq Client to use `ShutdownInitiator.Library` in this case. Is this a bug, or should I handle things differently?

StevenBonePgh:
## Proposed Changes

Rather than using `ShutdownInitiator.Application` in `ModelBase.WaitForConfirmsOrDieAsync`, use `ShutdownInitiator.Library` so consumers of the library can react differently based on the source of the Close.  See #1039.

## Types of Changes

## Checklist

## Further Comments

Unfortunately, I cannot determine a mechanism by which I can add unit tests to verify this behavior. While this change can be considered a breaking change, it is binary compatible (no public interface changes), and the change aligns with the documented behavioral expectation of Models that were Closed by the library itself. I do not believe that any user of the library would have relied on the previous behavior.

### DIFF
--- a/projects/RabbitMQ.Client/client/impl/ModelBase.cs
+++ b/projects/RabbitMQ.Client/client/impl/ModelBase.cs
@@ -1106,13 +1106,13 @@ namespace RabbitMQ.Client.Impl
                 }
 
                 await CloseAsync(
-                    new ShutdownEventArgs(ShutdownInitiator.Application, Constants.ReplySuccess, "Nacks Received",
+                    new ShutdownEventArgs(ShutdownInitiator.Library, Constants.ReplySuccess, "Nacks Received",
                         new IOException("nack received")),
                     false).ConfigureAwait(false);
             }
             catch (TaskCanceledException exception)
             {
-                await CloseAsync(new ShutdownEventArgs(ShutdownInitiator.Application,
+                await CloseAsync(new ShutdownEventArgs(ShutdownInitiator.Library,
                         Constants.ReplySuccess,
                         "Timed out waiting for acks",
                         exception),


StevenBonePgh:
I've run the tests locally on Windows with this change, and the `TestWaitForConfirmsWithEvents` is successful consistently with this change. In the debugger, I am seeing Assert.True(c > 20) where c is 88, 96, 72, etc - always well past the threshold for a test failure.  Is there a way to rerun the tests?

michaelklishin (approved):

michaelklishin:
I cannot reproduce this failure locally either but such tests are very timing-sensitive.

michaelklishin:
Backported to `6.x`.

PR #1064, opened 2021-08-10
Change recovery process to allow configuration recovery of closed channels

Inlustris:
## Proposed Changes

I figured out that a channel that is closed at the point where it is tying to recover is giving problems while recovering. The issue occurring "Not able to recover binding" is blocking the completion of the full recovery. Stopping before getting to recover the Consumers. All Connections and Channels are recovered at this point. (this results in a repeat every 5 seconds and creating new connection and channels every time)

While looking into it i thought how can it be this class(old channel) still exists? I used it within a constructor. It should be disposed. Trying to resolve this i created an using statement around this code for the channel. Which results in a correctly disposed channel. This results in the continuation of the recovery in case of the consumers. (And not repeating the recover process every 5 seconds)

Which brings me to the following exception within the recovery process. Where it is trying to recover bindings from the IModel that now is disposed. Because this is an unknown error for the recovery process it ignores this state and continues. This makes the recovery complete. 
```
it's not a known problem with connectivty, ignoring it
```

This still results in an uncomplete recovery because the binding was not able to recover. Reasons:
- Channel = Closed
- Channel = Disposed

Both of them result in the effect that the binding is not recovered. 
- Where as Closed results in failing of recovery at all with 5 secondsrepeat
- Where as Disposed results in full recovery except the bindings

```
Informational: Performing automatic recovery
Informational: Connection recovery completed
Error: Topology recovery exception
Error: Exception: RabbitMQ.Client.Exceptions.TopologyRecoveryException
Caught an exception while recovering binding between HTMLPDFDirect and HTMLPDFQueue: Cannot access a disposed object.
Object name: 'RabbitMQ.Client.Impl.AutorecoveringModel'.

InnerException:
System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'RabbitMQ.Client.Impl.AutorecoveringModel'.
   at RabbitMQ.Client.Impl.AutorecoveringModel.get_Delegate()
   at RabbitMQ.Client.Impl.RecordedEntity.get_ModelDelegate()
   at RabbitMQ.Client.Impl.RecordedQueueBinding.Recover()
   at RabbitMQ.Client.Framing.Impl.AutorecoveringConnection.RecoverBindings()
Informational: Will not retry recovery because of System.ObjectDisposedException: it's not a known problem with connectivty, ignoring it
Informational: Connection recovery completed
```

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #1061)

## Checklist

## Further Comments

I chose a simple solution to kick of a discussion. This would fix the problem on hand and would not create any issues for existing code so far as i can imagine. But this is probably not the correct way to do make this solution. This will trigger some kind of refactoring if this is correct

### DIFF
--- a/projects/RabbitMQ.Client/client/impl/AutorecoveringConnection.Recovery.cs
+++ b/projects/RabbitMQ.Client/client/impl/AutorecoveringConnection.Recovery.cs
@@ -146,7 +146,7 @@ namespace RabbitMQ.Client.Framing.Impl
                             RecoverExchanges(model);
                             RecoverQueues(model);
                             RecoverBindings(model);
-                            RecoverConsumers(model);
+                            RecoverConsumers();
                         }
                     }
 
@@ -278,7 +278,7 @@ namespace RabbitMQ.Client.Framing.Impl
             }
         }
 
-        private void RecoverConsumers(IModel model)
+        private void RecoverConsumers()
         {
             Dictionary<string, RecordedConsumer> recordedConsumersCopy;
             lock (_recordedEntitiesLock)
@@ -293,7 +293,7 @@ namespace RabbitMQ.Client.Framing.Impl
 
                 try
                 {
-                    cons.Recover(model);
+                    cons.Recover();
                     string newTag = cons.ConsumerTag;
                     UpdateConsumer(oldTag, newTag, cons);
 
--- a/projects/RabbitMQ.Client/client/impl/AutorecoveringConnection.Recovery.cs
+++ b/projects/RabbitMQ.Client/client/impl/AutorecoveringConnection.Recovery.cs
@@ -135,7 +135,7 @@ namespace RabbitMQ.Client.Framing.Impl
                     RecoverModels();
                     if (_factory.TopologyRecoveryEnabled)
                     {
-                        using (var model = _innerConnection.CreateModel())
+                        using (IModel model = _innerConnection.CreateModel())
                         {
                             // The recovery sequence is the following:
                             //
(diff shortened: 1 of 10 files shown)


michaelklishin (changes requested):
This does not seem to be a safe thing to do for consumers. If you recover a consumer on a one-off channel, and that channel is then closed, auto-delete queues may get deleted since all consumers will be gone (and there were some).

Inlustris:
Just found out the way to unit test correctly. I came to the same conclusion. and changed it for the consumer part. But i don't think you like the way i changed it. Can you give me something to go on to fix it correctly?

Now i created RecordedConsumerEntity to have different set functionality. but it is not clean.

Inlustris:
@michaelklishin Just updated my code with a cleaner solution to the whole problem.
- Created a new File for RecordedConsumerEntity
- Stripped RecordedEntity of its channel because it was not used anymore,
- Updated all files wich used this Constructor

michaelklishin:
@Inlustris I don't have a better idea than what you have in the second iteration,, thank you. Having a test is also promising.

 I'll try to QA this in the next couple of days.

Inlustris:
@michaelklishin When running the unit-tests i get the same problem i have with my own program. A restore loop which keeps restoring. 

This happens only when running all unit-tests at once. It loops on this test TestBasicModelRecovery. It won't get through this unit-test. (Won't fail and won't succeed) 

It does not happen when i run all unit-tests except the ones in TestConnectionRecovery. And run those in a new test run after the original.

All unit-test succeed. 

Maybe something to look at within the testing, could it be a race condition or left over data within the start of the recovery tests?

Inlustris:
@michaelklishin What is the current situation in completing this fix?

Inlustris:
@michaelklishin Can you confirm this is implemented by you guys already @ the moment

Inlustris:
@bollhals Am i right and is your fix the fix i proposed? And is this already fixed and i can close this PR?

bollhals:
> @bollhals Am i right and is your fix the fix i proposed? And is this already fixed and i can close this PR?

I'd say so yes.

Inlustris:
@bollhals I indeed see the same fix. I'm happy to see it fixed. Has there been a release for this already? I'm not sure how to follow the changes to a release.

PR #1081, opened 2021-09-02
fix topology recovery

### LINKED ISSUE #1076: ConnectionRecovery recovers models before consumers
I have a reproduced state in our application, where the connection recovery leads to an unrecoverable error.

Situation:
- 1 AutorecoverConnection + Channel
- The connection dropped (plugged out the cable)
- The recovery procedure starts its loop to recover
- Cable is plugged in again
- Connection can be established again, and the recover starts.

Now the fun begins...
While the recovery is [within these lines here](https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/6.x/projects/RabbitMQ.Client/client/impl/AutorecoveringConnection.cs#L420-L424) a RPC message gets tried to published. 

Since the models have already been recovered (Line 420), the publish actually succeeds, but the server will respond with `PRECONDITION_FAILED - fast reply consumer does not exist` because the consumer for the replies has not yet been recovered. (Line 424)

The outcome of this is that the channel will close and rendered unusable, while the connection recovery succeeds.

So the question is, should the AutorecoveringModel already be usable again prior to the recovery being finished?

It kind of is related to #1061 as it will touch similar area.
If a temporary model would restore the topology, then at least the time between the recovery of the models and the recovery of the consumers would be shortened. (Still possible tough)

bollhals:
## Proposed Changes

Backport of #1077 to fix #1076 on 6.x

## Types of Changes

## Checklist

## Further Comments

I've tried to reduce it to the minimum necessary to fix the issue.


bollhals:
Could this trigger a new 6.x  bugfix release?

PR #1096, opened 2021-10-07
introduce struct BasicProperties for basicPublish

bollhals:
## Proposed Changes

Introduces a struct BasicProperties and opens up the header implementation to a public.

Changes in detail:
- Replaced IBasicProperties in the receiving area with ReadOnlyBasicProperties
- Replaced BasicProperties class with a struct version
- Modified BasicPublish to be generic (TProperties : IBasicProperties, IAmqpHeader)
- Removed CreateBasicProperties method, replacement is 'new BasicProperties()'
- Introduce public IAmqpHeader interface. (Can be implemented by anyone and used as header properties)
- Removed IStreamProperties, IContentHeader (Not needed for anything)
- Aligned the name for getting the buffer size: GetRequiredPayloadBufferSize -> GetRequiredBufferSize
- Aligned the name for writing to the buffer: WritePropertiesTo -> WriteTo

## Types of Changes

- [ ] Bug fix (non-breaking change which fixes issue #NNNN)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)

## Checklist

- [x] I have read the `CONTRIBUTING.md` document
- [x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [x] All tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [x] Any dependent changes have been merged and published in related repositories

## Further Comments

I'm going to post a few benchmark numbers as soon as I get around to run them. (soon)

PR #1098, opened 2021-10-14
use ref instead of in for generic T + interface

bollhals:
## Proposed Changes

Figured this out while #1096, as the performance was not what I wanted it to be. While investigating I figured out that when we use `method<T>(in T parameter) where T: struct, Interface` and then call a method on it, the compiler is unable to figure out that they might be readonly. The effect of it is that it does a defensive copy. ([Link to dotnet issue](https://github.com/dotnet/csharplang/issues/3055))

While the structs itself are usually quite small, it still quite noticable. 

The workaround for this is to use ref instead of  in, as this allows calling methods on it without defensive copies.

## Types of Changes

## Checklist

## Further Comments

After
| Method |        When |       Runtime |       Mean |    Error |   StdDev | Code Size |
-------------------------- |----------- |-------------- |----------:|---------:|---------:|----------:|
| BasicPublishWriteNonEmpty | Before |      .NET 4.8 |  375.61 ns | 2.775 ns | 2.596 ns |    6935 B |
|  BasicPublishWriteNonEmpty | After |      .NET 4.8 |  370.54 ns | 1.648 ns | 1.542 ns |    6831 B |
|         BasicPublishWrite |Before |      .NET 4.8 | 206.93 ns | 0.450 ns | 0.399 ns |    6807 B | 
|          BasicPublishWrite | After |      .NET 4.8 |  203.22 ns | 0.407 ns | 0.361 ns |    6703 B | 
|   BasicPublishMemoryWrite | Before |      .NET 4.8 | 145.62 ns | 0.574 ns | 0.537 ns |    7208 B |
|    BasicPublishMemoryWrite | After |      .NET 4.8 |  138.49 ns | 0.663 ns | 0.620 ns |    7044 B |
| BasicPublishWriteNonEmpty | Before | .NET Core 3.1 |  263.46 ns | 1.574 ns | 1.473 ns |    5776 B |
|  BasicPublishWriteNonEmpty | After | .NET Core 3.1 |  250.61 ns | 1.237 ns | 1.157 ns |    5638 B |
|         BasicPublishWrite | Before | .NET Core 3.1 |  115.68 ns | 0.390 ns | 0.364 ns |    5776 B | 
|          BasicPublishWrite | After | .NET Core 3.1 | 106.93 ns | 0.436 ns | 0.408 ns |    5638 B |
|   BasicPublishMemoryWrite |Before | .NET Core 3.1 |  71.74 ns | 0.274 ns | 0.243 ns |    6318 B | 
|    BasicPublishMemoryWrite | After | .NET Core 3.1 |  60.68 ns | 0.356 ns | 0.333 ns |    6127 B |

### DIFF
--- a/projects/Benchmarks/WireFormatting/MethodFraming.cs
+++ b/projects/Benchmarks/WireFormatting/MethodFraming.cs
@@ -14,13 +14,13 @@ namespace RabbitMQ.Benchmarks
     [BenchmarkCategory("Framing")]
     public class MethodFramingBasicAck
     {
-        private readonly BasicAck _basicAck = new BasicAck(ulong.MaxValue, true);
+        private BasicAck _basicAck = new BasicAck(ulong.MaxValue, true);
 
         [Params(0)]
         public ushort Channel { get; set; }
 
         [Benchmark]
-        public ReadOnlyMemory<byte> BasicAckWrite() => Framing.SerializeToFrames(_basicAck, Channel);
+        public ReadOnlyMemory<byte> BasicAckWrite() => Framing.SerializeToFrames(ref _basicAck, Channel);
     }
 
     [Config(typeof(Config))]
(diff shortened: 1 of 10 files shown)


bollhals:
> This makes perfect sense since `in` basically is a message to the compiler saying: "This should be a read-only reference" so in forces a stack copy to be made. Good catch!
> 

Yes, but the strange thing is even if you have a generic method, it still doesn't recognizes the readyonly of the implementig method, it only seems to look at the interface, for which you can't put readonly on methods...

PR #1124, opened 2021-12-14
Clean-up the library's project file.

teo-tsirpanis:
## Proposed Changes

There are a couple of problems with `RabbitMQ.Client.csproj` that this PR fixes:

* The library's license was specified in a file. Since the license is a combination of known open-source licenses, it is now specified as an SPDX license expression, helping readability for both humans and machines.
* The project was publishing its debug symbols both in the main `.nupkg` file, and in a separate `.snupkg` file. There is no reason to do both; the symbols are now included only in the `.snupkg`, saving some tens of kilobytes from the `.nupkg`'s size. If there are compatibility concerns with older IDEs I think we should discontinue the `.snupkg` instead for the sake of reducing duplication.
* The project was publicly referencing two redundant packages; the one contains analyzers that do not need to be public, and the other contains the `Span` type and its friends for frameworks that don't already have it; .NET Core does not need it.

## Types of Changes

- [X] Bug fix (non-breaking change which fixes issue N/A)

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works (not applicable)
- [ ] I have added necessary documentation (if appropriate) (not applicable)
- [ ] Any dependent changes have been merged and published in related repositories (not applicable)

## Further Comments

N/A

### DIFF
--- a/projects/RabbitMQ.Client/RabbitMQ.Client.csproj
+++ b/projects/RabbitMQ.Client/RabbitMQ.Client.csproj
@@ -12,14 +12,13 @@
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageIcon>icon.png</PackageIcon>
-    <PackageLicenseFile>LICENSE</PackageLicenseFile>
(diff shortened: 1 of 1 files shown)


michaelklishin (approved):

michaelklishin:
Thank you!

michaelklishin:
Backported to `6.x` as well.

PR #1205, opened 2022-05-13
Fix incorrect unbind queue after recovery connection

MIVerTFT:
## Proposed Changes

It seems to me that there is an error in the connection recovery code. The error is that when restoring a binding between the exchange and the server-named queue, there is a change in the "RecordedQueueBinding.Destination" that affects the hash, which makes it impossible to delete records from the _recordedBindings dictionary, since the dictionary does not re-calculate the hash value. This situation eventually leads to the fact that all subsequent connection shutdown will not be correctly restored.

The written test demonstrates this problem.

## Types of Changes

## Checklist

## Further Comments

considered, etc.

### DIFF
--- a/projects/RabbitMQ.Client/client/impl/AutorecoveringConnection.cs
+++ b/projects/RabbitMQ.Client/client/impl/AutorecoveringConnection.cs
@@ -946,12 +946,26 @@ namespace RabbitMQ.Client.Framing.Impl
         {
             lock (_recordedEntitiesLock)
             {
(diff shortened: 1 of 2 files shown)


michaelklishin:
@MIVerTFT thank you. Can you please submit the same change against `main`?

MIVerTFT:
> @MIVerTFT thank you. Can you please submit the same change against main?

@michaelklishin Yes, no problem.

michaelklishin:
Yup, I've just found out just as much. Thank you for confirming @MIVerTFT.

PR #1233, opened 2022-06-25
Do not create string objects from consumerTag, exchange and routingKey, or get them from a string cache

zgabi:
## Proposed Changes

See issue #1231 

## Types of Changes

## Checklist


bollhals:
I'd prefer to split off at least the confirmTag into it's own PR. (as it contains a lot of special code to handle things, simpler to review track errors)

The other two are so similar that they can remain in this one.

As mentioned in the issue, I think we should look into also providing new API that take ROM<byte> for the exchangeName/RoutingKey

michaelklishin (dismissed):

michaelklishin:
I do not have a strong opinion on whether this should be two separate PRs. However,
> As mentioned in the issue, I think we should look into also providing new API that take ROM for the exchangeName/RoutingKey

would be nice and can be included into this PR or a separate one.

lukebakken (dismissed):
This seems like a reasonable way to reduce memory use.

Gsantomaggio:
I improved similarly to the .NET stream client [here](https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/pull/180), avoiding tons of allocations.

We need to test the PR deeply, but it makes sense to me. 

The problem here is the breaking changes.

zgabi:
@Gsantomaggio We are using this PR in our application since almost a year ago without any problem... Today I only updated it to the latest code. (nothing else changed in this PR)

lukebakken:
@zgabi thank you! We appreciate you keeping this PR up-to-date. As you can see, we're making progress on version 7.0.

zgabi:
@lukebakken Could you please run the build again, there was a problem in the APIApproval file

lukebakken:
Well, it's going to take some time to resolve conflicts with the current state of `main`.

@zgabi - if you have time to start resolving conflicts, I'd appreciate it.

lukebakken:
@zgabi - I've resolved a lot of the conflicts, but a couple compilation errors remain. I will return to this tomorrow.

lukebakken:
@zgabi @stebet I'm assuming that this code defeats the entire purpose of this PR:

https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/1233/files#diff-1d0ab4919eef61c6b1cc4efcc8849c4fc04b84cf71adcf4f757855de8ec2bfa3R120-R123

But, I'm wondering what we can do to still support this information in tracing?

zgabi:
You can write the following:

            string routingKey = Encoding.UTF8.GetString(deliverEventArgs.RoutingKey.Span);
            string exchange = Encoding.UTF8.GetString(deliverEventArgs.Exchange.Span);

So only string allocation is needed, no array.

And put the exchange line to line 136, so create only when needed in the if statement.

I'm not very familiar with the latest code.. is this Deliver executed every time when you deliver a message or just in some special case, e.g. when some tracing is enabled?

lukebakken:
Thank you @zgabi!!!!

lukebakken:
@zgabi thanks for all of your work on this. I'm planning to merge it by the end of the day tomorrow (16 May) unless there are further changes needed.

lukebakken:
Please see this comment: https://github.com/rabbitmq/rabbitmq-dotnet-client/issues/1231#issuecomment-2115566382

lukebakken:
@zgabi @bollhals @michaelklishin take a look at this commit:

https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/1233/commits/d4e88ca738219fbcdac09324e10b8ead38966c73

I would like to replace the usage of `ReadOnlyMemory<byte>` and `string`, when used for exchange and queue names, and routing keys, with the dedicated types in that commit. Thoughts?

zgabi:
Hi,

So the parameters of the of the HandleBasicDeliver will be alsoe `ExchangeKey` and `RoutingKey`?
For me it is ok.. I had a different branch/commit for this change (However they are using CachedStrings, but not so hard to separate them to the 2 new types):
https://github.com/zgabi/rabbitmq-dotnet-client/commits/routingkeyandexchange

This branch was an alternative to the "stringallocations" branch.

lukebakken:
@zgabi I realize this PR now contains an enormous number of changes, but the vast majority of them are the following:

* Replace `string` with `RoutingKey`, `QueueName`, `ExchangeName` or `ConsumerTag`, as appropriate.
* Update APIs with the above

Here are the "big" changes -
* Consumer dispatchers now pass `RentedMemory` for the method and body part of a delivery, which prevents string allocations unless the handler explicitly triggers them in the `AmqpString` instance - https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/1233/files#diff-74d6657893b833f7aefa9d2a615ae0d7dc1c4cfefa6615461c1b1638ff4b17abR53-R59
* [This](https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/1233/files#diff-1d0ab4919eef61c6b1cc4efcc8849c4fc04b84cf71adcf4f757855de8ec2bfa3R157-R162) is one such place where casting an instance of an `AmqpString` subclass will result in a string allocation.

@zgabi if you wouldn't mind re-running your allocation tests that would be great. Or, let me know how you did them and I will run them. Thanks!

bollhals:
While this change and the effort from you @lukebakken are honerable, I think it is currently way too big to properly analyze and judge the impact of the change. And also it contains many unrelated changes, making it even harder to judge.

My proposal would be to split off any unrelated changes, to pass them individually.

Regarding the main topic of the PR:
- With the introduction of these types, it's much clearer what you're suppose to pass as a parameter. But this comes at a tradeoff as well:
  - You always have to go through an indirection (the type) if you want / need it or not
  - You need to allocate this intermediate type, which slows you down if you do not plan on reusing it.
  
As it is right now, I think it is further away from what it tried to acheive than where the main is right now ... :/ 

IMHO, having classes wrapping string / bytes is only useful if 
- You should be reusing them / they're not supposed to be allocated just for this one call
- The library has a cache to avoid the allocation of it (e.g. Avoid allocating the exchange string + ExchangeName object for the delivery of messages, as they're often the same over and over again.)

projects/RabbitMQ.Client/client/impl/Connection.cs
```diff
@@ -484,7 +484,7 @@ public void Dispose()
             {
                 if (IsOpen)
                 {
-                    this.AbortAsync().GetAwaiter().GetResult();
+                    throw new InvalidOperationException("CloseAsync/AbortAsync must be called prior to Dispose");
```
bollhals, line 487: Throwing exceptions on dispose is generally not expected behaviour. I'd challange this change.

lukebakken: What is the best option, then?

I have run into bizarre CI failures that can only be traced to the use of `.GetAwaiter().GetResult()` within `Dispose()`

projects/Test/Integration/TestExchangeDeclare.cs
```diff
@@ -61,18 +61,21 @@ public async Task TestConcurrentExchangeDeclareAndBind()
             {
                 async Task f()
                 {
-                    try
+                    using (IChannel ch = await _conn.CreateChannelAsync())
```
bollhals, line 64: Wasn't the idea of these tests to verify that they work when called in one channel? So splitting them to many channels defeats the purpose of those tests no?

lukebakken: See my comment [here](https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/1233#discussion_r1608906517).

`IModel` / `IChannel` has never been guaranteed to be thread safe, so I removed

lukebakken:
Hi @bollhals thanks for your input.

> The library has a cache to avoid the allocation of it (e.g. Avoid allocating the exchange string + ExchangeName object for the delivery of messages, as they're often the same over and over again.)

Well, that's exactly what this PR accomplishes with the `AmqpString`/`ExchangeName` types, so there's that at least 😄.

I will see about moving other changes to a different PR.

projects/RabbitMQ.Client/client/impl/ConsumerDispatching/AsyncConsumerDispatcher.cs
```diff
@@ -26,22 +27,56 @@ protected override async Task ProcessChannelAsync(CancellationToken token)
...
                                 {
-                                    WorkType.Deliver => work.AsyncConsumer.HandleBasicDeliver(
-                                        work.ConsumerTag, work.DeliveryTag, work.Redelivered,
-                                        work.Exchange, work.RoutingKey, work.BasicProperties, work.Body.Memory),
+                                    case WorkType.Deliver:
+                                        {
+                                            IAsyncBasicConsumer? deliverConsumer = work.AsyncConsumer ??
+                                                throw new InvalidOperationException("[CRITICAL] should never see this");
+                                            ConsumerTag? deliverConsumerTag = work.ConsumerTag;
+                                            var method = new BasicDeliver(work.Method.Memory);
```
bollhals, line 38: I do not see where the rented method memory is released again? Have I missed it somehwere?

bollhals: (Applies to everywhere, not specifically to this call here)

lukebakken: https://github.com/zgabi/rabbitmq-dotnet-client/blob/stringallocations/projects/RabbitMQ.Client/client/impl/ConsumerDispatching/AsyncConsumerDispatcher.cs#L26-L86

The `WorkStruct` is disposed at the end of this `using` block:

https://github.com/zgabi/rabbitmq-dotnet-client/blob/stringallocations/projects/RabbitMQ.Client/client/impl/ConsumerDispatching/AsyncConsumerDispatcher.cs#L26

projects/Test/Integration/TestExchangeDeclare.cs
lukebakken: Now that `Dispose` does not use `GetAwaiter().GetResult()`, I can modify these tests to test both cases.

projects/RabbitMQ.Client/client/impl/ConsumerDispatching/AsyncConsumerDispatcher.cs
lukebakken: Don't get me wrong, I am not thrilled with how convoluted the code is in this library, but it is what it is at this time.

projects/RabbitMQ.Client/client/impl/Connection.cs
lukebakken: This is one example, from running the entire testsuite on my local Windows 11 workstation. I have no idea how the operations [here](https://github.com/zgabi/rabbitmq-dotnet-client/blob/stringallocations/projects/Test/Integration/TestAsyncConsumer.cs#L598-L620) could have lead to the following error. Notice that, somehow, the test's `DisposeAsync` method (which calls `CloseAsync`) occurred at _the same time_ as a `BasicCancelAsync` operation 🤔 Each operation would have had to acquire the same `SemaphoreSlim` (since the test was using the shared `_channel` instance when the error happened). This is also why I modified the tests to use dedicated `IChannel` instances per-Task:

[rabbitmq-dotnet-client-1233-stacktrace.txt](https://github.com/rabbitmq/rabbitmq-dotnet-client/files/15395296/rabbitmq-dotnet-client-1233-stacktrace.txt)

projects/RabbitMQ.Client/client/impl/ConsumerDispatching/AsyncConsumerDispatcher.cs
bollhals: Ah thanks, missed that using, as it was already like that before the change, so it didn't show up 👍

lukebakken: No problem! It is not intuitive **at all**

lukebakken:
> IMHO, having classes wrapping string / bytes is only useful if

@bollhals there is another benefit - I have had RabbitMQ support cases that boil down to the wrong string passed as the wrong parameter to a library call. Having dedicated types makes that kind of mistake much less likely.

projects/RabbitMQ.Client/client/impl/Connection.cs
lukebakken: I could look into bringing back a synchronous `Abort` to be used by `Dispose` in another PR.

bollhals: Generally you either do sync or async, which includes on Dispose you call sync methods not asyncs. Otherwise you'll implement IAsyncDisposable and do it there.

On the other hand, this.AbortAsync().GetAwaiter().GetResult(); Should be the way to go. It will block the current thread as a consequence, which might be problematic if
- The threadpool is maxed out, all threads are blocked (See Threadpool starvation)
- the thread has a synchronization context set and it deadlocks itself (Not entirely sure about the details here, rarely happens wihtout an UI AFAIK)


So I don't know exactly based on this why it happens without looking deep into it.

Sidenote: this would be a change I rather see separate from this PR, as I see the intent of enfocing the close as a different breaking change rather than  it being burried in a large PR with a different goal in mind.

bollhals:
> > IMHO, having classes wrapping string / bytes is only useful if
> 
> @bollhals there is another benefit - I have had RabbitMQ support cases that boil down to the wrong string passed as the wrong parameter to a library call. Having dedicated types makes that kind of mistake much less likely.

Sure, that's what I ment by stating to make it much clearer what to pass.
But the tradeoff is the allocation of the objects and the indirection through another object.  

It depends in the end what the goal is:
- Increasing performance? => I have a feeling that these additional objects & indirection made it slower for most real world usecases. (Perf tests needed to confirm)
- Increasing usability? => Fine, this surely helps, but then the PR Title should be changed, as the goal of not allocating string objects becomes a side thing.

lukebakken:
I'll run some performance tests. I don't _think_ I have access to memory profiling tools to see how this PR affects allocations ... @zgabi ?

lukebakken:
@zgabi @bollhals I have force-reset the `stringallocations` branch to 2320480c286d460d71d997661413594117c01c29, which is just before I went crazy with this PR. I saved my work on another branch.

projects/RabbitMQ.Client/client/impl/Connection.cs
paulomorgado: `GetAwaiter().GetResult()` per-se should not throw exceptions. Exceptions come from task being observed.

The exception in question seems to be coming from [here](https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/929047d5eef58f238e2162ac6a2fddb55d2b985b/projects/RabbitMQ.Client/client/impl/RpcContinuationQueue.cs#L79)

If you are thinking of providing synchronous APIs, they should not be really sync and not sync over async. If it's just for the sake of implementing `IDisposable`, implement just `IAsyncDisposable`.

paulomorgado:
> * Increasing performance? => I have a feeling that these additional objects & indirection made it slower for most real world usecases. (Perf tests needed to confirm)

Sometimes, minimal local extra work/complexity avoids extra processing, allocations, GC work. It really needs to be measured.

lukebakken:
Here is what I see when comparing allocations between the `stringallocations` branch and the current `main` (as of 1600PDT 2024-05-22), using the `MassPublish` program to send 64 batches of 1024 messages per batch.

It seems like `stringallocations` is a bit faster (due to fewer string allocations? ... I will re-test that). Memory usage appears to be about the same.

@zgabi @bollhals @paulomorgado thoughts?

## `stringallocations` branch
image
## `main` branch
image

lukebakken:
Hi @paulomorgado 

> If the stringallocations branch has 98,479 strings and the main branch has 265,205 strings, where did those strings go? What are those strings? Can those strings be pooled/cached?

That's the point of this PR, to reduce string allocations when this library handles `basic.deliver` from the AMQP server. Exchange name and routing key used to be converted to strings, and this PR keeps them as `ReadOnlyMemory<byte>`

> I find very worrying the existence of System.ArrayEnumerator instances in the heap. This is an indication that they are being boxed. Coupling this with System.Linq.Enumerable (a known performance killer), I'd say LONQ to Objects is being used and shouldn't.

That's easy to fix. Visual Studio "helpfully" suggested adding a `.Cast<>` in a couple `foreach` loops. I will open a new PR.

> What is System.Collections.Generic.LinkedList<System.UInt64> being used for?

```
C:\Users\lbakken\development\rabbitmq\rabbitmq-dotnet-client_main [main ≡]
> git grep LinkedList
projects/RabbitMQ.Client/client/impl/ChannelBase.cs:63:        private readonly LinkedList<ulong> _pendingDeliveryTags = new LinkedList<ulong>();
```

paulomorgado:
Hi @lukebakken

> > If the stringallocations branch has 98,479 strings and the main branch has 265,205 strings, where did those strings go? What are those strings? Can those strings be pooled/cached?
> 
> That's the point of this PR, to reduce string allocations when this library handles `basic.deliver` from the AMQP server. Exchange name and routing key used to be converted to strings, and this PR keeps them as `ReadOnlyMemory<byte>`

As you pointed out, the only change is the reduced number of `string` instances and not overall memory or CPU.

So, what's the benefit here?

Is it the tests that are not taking advantage of this?

> ```
> C:\Users\lbakken\development\rabbitmq\rabbitmq-dotnet-client_main [main ≡]
> > git grep LinkedList
> projects/RabbitMQ.Client/client/impl/ChannelBase.cs:63:        private readonly LinkedList<ulong> _pendingDeliveryTags = new LinkedList<ulong>();
> ```

So, it's kind of a queue, but you need to remove items from both ends and, sometimes, anywhere. Is that it?

lukebakken:
@zgabi - if you have time, please read @paulomorgado's observations:

> As you pointed out, the only change is the reduced number of string instances and not overall memory or CPU. So, what's the benefit here? Is it the tests that are not taking advantage of this?

I am running the `MassPublish` program using the ".NET Object Allocation Tracking" performance monitor, and, as @paulomorgado notes, there does not seem to be a reduction in memory use.

At this point I'm getting the feeling like I'm testing this incorrectly, perhaps?

zgabi:
Reducing the allocations reduces the GC overhead only.
You wrote that the running time and memory is almost the same... but is the CPU usage also the same? I think it should be bigger (because of multiple threads) when there are a lot of allocations.

paulomorgado:
I'm assuming the same test conditions, otherwise the tests are not comparable.

lukebakken:
Hi everyone -

In the interest of shipping v7 of this library this month, I am leaning towards moving this PR to version 8. I am going to release RC1 of version 7 by the end-of-day tomorrow, June 4th, 2024, so if someone has a _very_ strong opinion about why this PR should be in version 7, comment here.

paulomorgado:
I haven't had the time to look into it properly.

If this still allows the user to use `string` and can be improved in a 7.x version, I'd keep it in 7.0.0.

lukebakken:
> If this still allows the user to use string and can be improved in a 7.x version, I'd keep it in 7.0.0.

There would still have to be work in the API to keep it consistent. Given the time constraint and what I've seen in testing it doesn't seem like the benefit is worth further delay of v7. I still have quite a few issues to get through in the milestone (https://github.com/rabbitmq/rabbitmq-dotnet-client/milestone/53)

paulomorgado:
Shiping is also a feature, right?

If there are doubts, it should be postponed.

If it is just addictive for the API point of view, it doesn't need to be a new major version.

danielmarbach:
Should this simply be closed?

Issue #662, opened 2019-10-21
(un)signed shorts in table serialisation


visander:
We would like to use an official release for our RabbitMQ library and in the latest stable version 5.1.2 the (un)signed shorts are missing in ReadFieldValue in WireFormatting.cs

Is this planned for a future release?

michaelklishin:
Before you draw any conclusions about what table field types are supported or not, see the [Protocol Errata](https://www.rabbitmq.com/amqp-0-9-1-errata.html#section_3) section.

This client is over a decade old, so we can confidently tell that some field types are basically irrelevant, possibly because there is no meaningful representation of those types in languages such as Python or JavaScript.

We would consider a PR for anything that's missing.

visander:
Thanks for the fast reply. 

The errate you mention states that Rabbit supports these two types:
        U       s            Signed 16-bit      (A1)
        u       u            Unsigned 16-bit

From this draw the conclusion that it is missing in the dotnet implementation. Or did I miss something?

michaelklishin:
`s` [is present](https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/master/projects/client/RabbitMQ.Client/src/client/impl/WireFormatting.cs#L159) in `ReadFieldValue` :

``` csharp
case 's':
                    value = reader.ReadInt16();
                    break;
```

`u` is not.

visander:
My apologizes, I missed that one.
Will the unsigned short also be implemented then?

michaelklishin:
@visander as I said this field has been missing for a decade and this is the first time in about seven years that I see it brought up (at least for this client). However, this is open source software. You are welcome to implement it.

PR #1270, opened 2022-10-17
fix: add unsigned short support in table deserialisation

gytsen:
While it seems that this wasn't needed in the lifetime of this library so far, it seems weird not to support the full protocol as stated on the RabbitMQ website (including errata).

While I've been unable to reproduce it locally with a local install of RabbitMQ server, I have encountered at least one instance where a server sent a numeric header value that fit into an unsigned short over the wire. My extremely unscientific hypothesis is that a strict client pushed this message onto the queue which the server passed on directly, but I could be wrong here.

Bottom line is: I have a crashing `RabbitMQ.Client` that says it doesn't understand the field type `'u'` while the docs clearly say it's part of the protocol.

This commit adds parsing support for unsigned shorts, which should ensure the library conforms fully to the protocol in those few edge cases where it is still needed.

I added tests in the `TestFieldTableFormatting` and `TestFieldTableFormattingGeneric` files since that seemed the most logical place to put them, but I could have missed a more appropriate test.

Inspired by: #662

## Proposed Changes

As described in #662, `RabbitMQ.Client` does not support unsigned short reading out of the box. Since this is a trivial addition with relatively low-impact by my estimates, I'd figure I'd add it to ensure the library is not at fault when facing stricter or older servers.

~~As for the nature of the change, it would be and observable change in existing systems in that the library doesn't crash anymore when RMQ servers send unsigned shorts over the wire. I'm not sure if this is considered breaking or not.~~

I had to change the writer for the roundtrip tests as well, but that poses a conundrum: where previously unsigned shorts were
explicitly rejected, it is now accepted and serialized. That is a pretty big change to make, so I'm pretty sure that's a breaking change. Especially since it seems that some libraries (I checked [node-amqp](https://github.com/amqp-node/amqplib/blob/b3647780265dd0f84451b0a29776f6c2dbfa9d3d/lib/codec.js#L192) real quick) just do not support these types. Then again, there's a [pull request already open](https://github.com/amqp-node/amqplib/pull/620) to fix that.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #662)

## Checklist

### DIFF
--- a/projects/RabbitMQ.Client/client/impl/WireFormatting.Read.cs
+++ b/projects/RabbitMQ.Client/client/impl/WireFormatting.Read.cs
@@ -130,6 +130,9 @@ namespace RabbitMQ.Client.Impl
                     case 's':
                         bytesRead = 3;
                         return NetworkOrderDeserializer.ReadInt16(slice);
+                    case 'u':
+                        bytesRead = 3;
+                        return NetworkOrderDeserializer.ReadUInt16(slice);
                     case 'T':
                         bytesRead = 1 + ReadTimestamp(slice, out var timestamp);
                         return timestamp;
--- a/projects/RabbitMQ.Client/client/impl/WireFormatting.Write.cs
+++ b/projects/RabbitMQ.Client/client/impl/WireFormatting.Write.cs
@@ -200,6 +200,10 @@ namespace RabbitMQ.Client.Impl
                     case BinaryTableValue val:
                         destination = (byte)'x';
                         return 1 + WriteLongstr(ref fieldValue, val.Bytes);
+                    case ushort val:
+                        destination = (byte)'u';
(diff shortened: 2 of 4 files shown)


michaelklishin:
Most clients do not support unsigned types because they are either absent or extremely rarely used in their respective programming languages/communities.

This should not be a breaking change for systems that still do not use unsigned shorts.

michaelklishin (approved):

michaelklishin:
Thank you!

michaelklishin:
I think this is safe to backport.

gytsen:
I could take a stab at a backport, which versions would need it?

michaelklishin:
Just cherry-picking this to `6.x` would be enough. Thank you.

gytsen:
Alright, I'll try to get a PR ready for it later this evening for you

PR #1619, opened 2024-06-28
Avoid boxing in IEndpointResolverExtensions.cs

lechu445:
## Proposed Changes

Technically T can be a value type. `Object.Equals` has an overload that takes `object` therefore any struct would be boxed. To avoid it, `EqualityComparer<T>.Default.Equals` should be used.

## Types of Changes

## Checklist


lukebakken (approved):
Thanks for noticing. This isn't a hot path in the code, so there won't be a perf improvement, but it's good to know about this option at any rate.

PR #1628, opened 2024-07-05
make IncomingCommand a class and simplify code around it

bollhals:
## Proposed Changes

Similar to the frame one, there's also just one Command at a time. So it can be allocated once and updated throughout the lifetime. 
This way it can be changed to a class instead of struct, and thus avoids all the complexity of the struct handling especially with the async code.

Slight performance increase as well.

| Method              | Mean     | Error   | StdDev  | Ratio | Gen0      | Allocated | Alloc Ratio |
|-------------------- |---------:|--------:|--------:|------:|----------:|----------:|------------:|
| Pr_Publish_Hello_World | 275.2 ms | 5.50 ms | 8.57 ms |  1.00 | 1000.0000 |   5.85 MB |        1.00 |
| Main_Publish_Hello_World | 279.8 ms | 5.55 ms | 11.47 ms |  1.00 | 1000.0000 |   5.84 MB |        1.00 |

## Types of Changes

- [x] Performance / simplification

## Checklist

## Further Comments

I've changed a few things along the way

### DIFF
--- a/projects/RabbitMQ.Client/client/RentedMemory.cs
+++ b/projects/RabbitMQ.Client/client/RentedMemory.cs
@@ -36,8 +36,6 @@ namespace RabbitMQ.Client
 {
     internal struct RentedMemory : IDisposable
     {
-        private bool _disposedValue;
-
         internal RentedMemory(byte[] rentedArray)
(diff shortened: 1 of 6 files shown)


lukebakken (approved):
Thank you!

PR #1690, opened 2024-09-27
Fix exception when refreshing oauth2 token

BreakingBread0:
## Proposed Changes

This fixes a small bug that happens when refreshing the OAuth2 access token from the OAuth backchannel provider. The dictionary key "scope" in BuildRefreshParameters is already set in BuildRequestParameters and therefore this throws an exception.

## Types of Changes

- [x] Bug fix (has no issue associated with it)

## Checklist

## Further Comments

Before being added to the dictionary, _scope undergoes two different checks: Nullability and string.IsNullOrEmpty in the two aforementioned methods. I do not know if this is intentional or not.

### DIFF
--- a/projects/RabbitMQ.Client.OAuth2/OAuth2Client.cs
+++ b/projects/RabbitMQ.Client.OAuth2/OAuth2Client.cs
@@ -309,11 +309,6 @@ namespace RabbitMQ.Client.OAuth2
             Dictionary<string, string> dict = BuildRequestParameters();
             dict[GRANT_TYPE] = REFRESH_TOKEN;
 
-            if (_scope != null)
-            {
-                dict.Add(SCOPE, _scope);
-            }
-
             if (token.RefreshToken != null)
             {
                 dict.Add(REFRESH_TOKEN, token.RefreshToken);


projects/RabbitMQ.Client.OAuth2/OAuth2Client.cs
```diff
@@ -309,7 +309,7 @@ private Dictionary<string, string> BuildRefreshParameters(IToken token)
             Dictionary<string, string> dict = BuildRequestParameters();
             dict[GRANT_TYPE] = REFRESH_TOKEN;
 
-            if (_scope != null)
+            if (_scope != null && !dict.ContainsKey(SCOPE))
```
danielmarbach, line 312: Wouldn't in general the safer code be to replace the key?

BreakingBread0: The whole check can probably be omitted entirely since the SCOPE dictionary key is already being set in BuildRequestParameters.

BreakingBread0:
> Thank you for this PR @BreakingBread0. Would you mind sharing which version of this library you found this issue in?

The versions I currently use (and in which I found the bug) are:
* RabbitMQ.Client 6.8.1 and
* RabbitMQ.Client.OAuth2 1.0.0
Both were installed from Nuget.

lukebakken (approved):
Thank you @BreakingBread0

PR #1843, opened 2025-06-06
Improve AsyncManualResetEvent implementation to address races

danielmarbach:
## Proposed Changes

I used the implementation further in some very high concurrent scenarios and ran into token problems with the manual reset implementation under the cover. I have concluded the original version I tweaked had some races that we collectively missed. 

The race condition occurs in WaitAsync() where the IsSet check and valueTaskSource.Version capture happen at different moments without synchronization, allowing Set() or Reset() to execute between these operations and change the state. If Reset() is called after the IsSet check passes but before the ValueTask is created, the version becomes stale and the awaited task will never complete because it references the old version while the ManualResetValueTaskSourceCore has been reset. Additionally, the gap between checking IsSet and updating state in both Set() and Reset() creates windows where multiple threads can pass the initial checks simultaneously, leading to operations being performed on inconsistent state.

This implementation passed my concurrency tests, but it doesn't hurt if the original involved reviewers give this another review @lukebakken @paulomorgado @bollhals 

## Types of Changes

## Checklist

## Further Comments

considered, etc.

### DIFF
--- a/projects/RabbitMQ.Client/Impl/AsyncManualResetEvent.cs
+++ b/projects/RabbitMQ.Client/Impl/AsyncManualResetEvent.cs
@@ -29,127 +29,44 @@
 //  Copyright (c) 2007-2025 Broadcom. All Rights Reserved.
 //---------------------------------------------------------------------------
 
-using System;
-using System.Diagnostics.CodeAnalysis;
 using System.Threading;
 using System.Threading.Tasks;
-using System.Threading.Tasks.Sources;
 
 namespace RabbitMQ.Client.Impl
 {
-    sealed class AsyncManualResetEvent : IValueTaskSource
+    sealed class AsyncManualResetEvent
     {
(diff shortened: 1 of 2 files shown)


danielmarbach:
I pushed an implementation based on https://raw.githubusercontent.com/dotnet/runtime/refs/heads/main/src/libraries/System.Net.Quic/src/System/Net/Quic/Internal/ValueTaskSource.cs which should properly address the races and keep the allocations on par with the current version

danielmarbach:
I reverted the value task source based implementation. The TCS based version reliably works while the value task source based one I have already tried to implement several "fixes" even with assistance I could never get it to be race free under load

danielmarbach:
If someone wants to go down that level of optimization, be my guest but I have depleted my available experimenting and fiddling around budget I'm willing to spend on this

lukebakken (approved):
Thank you!

PR #1901, opened 2026-02-06
Fix unconditional semaphore release in BasicPublishAsync when Cancell…

### LINKED ISSUE #1900: Publisher confirms: _confirmSemaphore broken by cancelled CancellationToken allows concurrent access, corrupts _nextPublishSeqNo
### Describe the bug

When `CancellationToken` is cancelled during `BasicPublishAsync`, `_confirmSemaphore.WaitAsync(ct)` in `MaybeStartPublisherConfirmationTrackingAsync` throws `OperationCanceledException` **before acquiring** the semaphore. The `finally` block always calls `MaybeEndPublisherConfirmationTrackingAsync` which does `_confirmSemaphore.Release()` unconditionally — releasing a semaphore that was never acquired.

After many cancelled tokens trigger this path, the semaphore breaks and allows multiple threads into the critical section simultaneously. This causes two issues:

1. **TryAdd collision (Stack Trace 1):** Two threads read the same `_nextPublishSeqNo = K`, both call `TryAdd(K)` — one fails with "already exists". The winner's `++` then `--` (on `ModelSendAsync` failure) brings counter back to K → counter never advances → "already exists for K" repeats on every publish.

2. **SemaphoreFullException (Stack Trace 2):** If the over-released semaphore count reaches max, `Release()` throws `SemaphoreFullException` → wrapped as "BUG FOUND".
3. **Lost decrement:** When both threads are in the critical section and both fail (`ModelSendAsync` throws), both call `_nextPublishSeqNo--` but the decrement is not thread-safe (read-modify-write) — both read K and write K-1. Counter decreases by 1 instead of 2, leaving it higher than expected with an orphaned entry in `_confirmsTaskCompletionSources`.

**Stack Trace 1** — Stuck sequence number (repeats the same number):

```
System.InvalidOperationException: Failed to track the publisher confirmation for sequence number '777' because it already exists.
   at RabbitMQ.Client.Impl.Channel.MaybeStartPublisherConfirmationTrackingAsync(CancellationToken cancellationToken)
   at RabbitMQ.Client.Impl.Channel.BasicPublishAsync[TProperties](String exchange, String routingKey, Boolean mandatory, TProperties basicProperties, ReadOnlyMemory`1 body, CancellationToken cancellationToken)
   at RabbitMQ.Client.Impl.Channel.BasicPublishAsync[TProperties](String exchange, String routingKey, Boolean mandatory, TProperties basicProperties, ReadOnlyMemory`1 body, CancellationToken cancellationToken)
   at MassTransit.RabbitMqTransport.RabbitMqChannelContext.<>c__DisplayClass12_1.<<BasicPublishAsync>g__WaitAck|1>d.MoveNext()
   at MassTransit.RabbitMqTransport.ScopeChannelContext.BasicPublishAsync(...)
   at MassTransit.RabbitMqTransport.SharedChannelContext.BasicPublishAsync(...)
   at MassTransit.RabbitMqTransport.RabbitMqSendTransportContext.Send[T](...)
   at MassTransit.Transports.SendTransport`1.SendPipe`1.Send(TContext context)
```

**Stack Trace 2** — SemaphoreFullException (BUG FOUND):

```
System.InvalidOperationException: BUG FOUND - please report this exception (with stacktrace) here: https://github.com/rabbitmq/rabbitmq-dotnet-client/issues
   at RabbitMQ.Client.Impl.Channel.MaybeEndPublisherConfirmationTrackingAsync(PublisherConfirmationInfo publisherConfirmationInfo, CancellationToken cancellationToken)
   at RabbitMQ.Client.Impl.Channel.BasicPublishAsync[TProperties](...)
   at MassTransit.RabbitMqTransport.RabbitMqChannelContext.<>c__DisplayClass12_1.<<BasicPublishAsync>g__WaitAck|1>d.MoveNext()
   at MassTransit.RabbitMqTransport.ScopeChannelContext.BasicPublishAsync(...)
   at MassTransit.RabbitMqTransport.SharedChannelContext.BasicPublishAsync(...)
   at MassTransit.Internals.TaskExtensions.<>c__DisplayClass1_0.<<OrCanceled>g__WaitAsync|0>d.MoveNext()
   at MassTransit.RabbitMqTransport.RabbitMqSendTransportContext.Send[T](...)
   at MassTransit.Transports.SendTransport`1.SendPipe`1.Send(TContext context)

Inner: System.Threading.SemaphoreFullException: Adding the specified count to the semaphore would cause it to exceed its maximum count.
   at System.Threading.SemaphoreSlim.Release(Int32 releaseCount)
   at RabbitMQ.Client.Impl.Channel.MaybeEndPublisherConfirmationTrackingAsync(...)
```

### Reproduction steps

1. Use `BasicPublishAsync` with publisher confirms enabled and confirmation tracking enabled
2. Cancel `CancellationToken` while multiple threads are calling `BasicPublishAsync` (e.g. connection drop under load, MassTransit cancellation)
3. The cancelled token causes `WaitAsync` to throw before acquiring `_confirmSemaphore`
4. `finally` block calls `Release()` without acquiring → semaphore broken
5. After enough cancellations, multiple threads enter the critical section concurrently
6. Concurrent reads of `_nextPublishSeqNo` cause → "already exists" stuck forever, or `SemaphoreFullException` → "BUG FOUND"

### Expected behavior

`_confirmSemaphore.Release()` should only be called when the semaphore was actually acquired. When `WaitAsync` throws due to cancellation, the `finally` block should skip the `Release()`.

## Environment

- **RabbitMQ.Client**: 7.2.0
- **MassTransit**: 8.5.5 (RabbitMQ transport)
- **.NET**: 10.0

EvheniyHlushko:
When `CancellationToken` is already cancelled during `BasicPublishAsync`, `_confirmSemaphore.WaitAsync()` fails but the `finally` block unconditionally releases the semaphore. This breaks the semaphore and allows concurrent access, corrupting `_nextPublishSeqNo`.

The fix makes the semaphore release conditional on successful acquisition.

Fixes #1900

### DIFF
--- a/projects/RabbitMQ.Client/Impl/Channel.BasicPublish.cs
+++ b/projects/RabbitMQ.Client/Impl/Channel.BasicPublish.cs
@@ -51,7 +51,8 @@ namespace RabbitMQ.Client.Impl
         {
             PublisherConfirmationInfo? publisherConfirmationInfo = null;
             RateLimitLease? lease =
-                await MaybeAcquirePublisherConfirmationLockAsync(cancellationToken).ConfigureAwait(false);
+                await MaybeAcquirePublisherConfirmationLockAsync(cancellationToken)
+                    .ConfigureAwait(false);
             try
             {
                 publisherConfirmationInfo = MaybeStartPublisherConfirmationTracking();
@@ -107,7 +108,8 @@ namespace RabbitMQ.Client.Impl
         {
             PublisherConfirmationInfo? publisherConfirmationInfo = null;
             RateLimitLease? lease =
-                await MaybeAcquirePublisherConfirmationLockAsync(cancellationToken).ConfigureAwait(false);
+                await MaybeAcquirePublisherConfirmationLockAsync(cancellationToken)
+                    .ConfigureAwait(false);
             try
             {
                 publisherConfirmationInfo = MaybeStartPublisherConfirmationTracking();
--- a/projects/RabbitMQ.Client/Impl/Channel.PublisherConfirms.cs
+++ b/projects/RabbitMQ.Client/Impl/Channel.PublisherConfirms.cs
@@ -288,8 +288,9 @@ namespace RabbitMQ.Client.Impl
                 RateLimitLease? lease = null;
                 if (_publisherConfirmationTrackingEnabled && _outstandingPublisherConfirmationsRateLimiter is not null)
                 {
-                    lease = await _outstandingPublisherConfirmationsRateLimiter.AcquireAsync(cancellationToken: cancellationToken).ConfigureAwait(false);
-                    if (!lease.IsAcquired)
+                    lease = await _outstandingPublisherConfirmationsRateLimiter.AcquireAsync(cancellationToken: cancellationToken)
+                        .ConfigureAwait(false);
+                    if (false == lease.IsAcquired)
                     {
                         throw new InvalidOperationException("Could not acquire a lease from the rate limiter.");
                     }
@@ -297,7 +298,8 @@ namespace RabbitMQ.Client.Impl
 
                 try
                 {
-                    await _confirmSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+                    await _confirmSemaphore.WaitAsync(cancellationToken)
+                        .ConfigureAwait(false);
                 }
                 catch (OperationCanceledException)
                 {


danielmarbach:
@EvheniyHlushko I tried to pull the change apart because I had a hard time seeing the bug fix

danielmarbach (dismissed):

PR #1906, opened 2026-02-24
Connection recovery: remove recorded bindings of auto-delete entities

### LINKED ISSUE #1905: Auto-recovery attempts to restore bindings to deleted queues causing recovery failures
### Describe the bug

When a channel with an exclusive atodelete queue is closed (for example, during or after a connection loss), the queue is properly removed from the recovery tracking, but the queue's bindings remain tracked. This causes the auto-recovery mechanism to repeatedly attempt restoring bindings to a queue that no longer exists, resulting in recovery errors and repeated shutdown events.

and there might be design reasons I'm not aware of, but this behavior seems inconsistent. When explicitly deleting a queue via QueueDeleteAsync(string, bool, bool, bool, CancellationToken), the method DoDeleteRecordedQueueAsync correctly removes both the queue and its bindings. However, when a queue is implicitly removed through DeleteAutoDeleteQueue(string) (called when closing a channel with consumers), only the queue is removed from tracking while bindings remain.
This creates an unexpected situation where closing a channel appears to clean up the queue but leaves orphaned bindings that cause issues during connection recovery.

Root cause location:
AutorecoveringConnection.Recording.cs:414, method DeleteAutoDeleteQueue(string) (line ~414)
•	Currently removes the queue but not the bindings
•	Compare with DoDeleteRecordedQueueAsync which properly removes both

I am not that experienced with the library and I apologize if I'm misunderstanding the intended design, but I wanted to report this behavior in case it's unintended.

### Reproduction steps

I've attached a minimal console application that reproduces this issue (RabbitmqOrphanedBindingAutorecoveryIssue project).
Steps:
1.	Run the application (connects to localhost RabbitMQ)
2.	Application creates an exclusive queue with a binding
3.	Stop RabbitMQ server (simulate connection loss)
4.	Press enter in the console so code advances and closes the channel
5.	Restart RabbitMQ server
6.	Observe repeated recovery failures attempting to bind to the non-existent queue

[RabbitmqOrphanedBindingAutorecoveryIssue.zip](https://github.com/user-attachments/files/25522157/RabbitmqOrphanedBindingAutorecoveryIssue.zip)

### Expected behavior

Expected behavior:
•	One shutdown event when connection is lost
•	Clean recovery after RabbitMQ restart
•	No errors about missing queues

Actual behavior:
•	Multiple shutdown events (2, 3, 4...)
•	Recovery errors: "no queue 'test.exclusive.queue' in vhost '/'"
•	Continuous recovery failures until application is stopped

### Additional context

_No response_

michaelklishin:
The same way Java client does it [1][2][3].

Fixes #1905.

1. https://github.com/rabbitmq/rabbitmq-java-client/blob/main/src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringChannel.java#L913-L919
2. https://github.com/rabbitmq/rabbitmq-java-client/blob/main/src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java#L1118-L1131
3. https://github.com/rabbitmq/rabbitmq-java-client/blob/main/src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java#L1063-L1069

### DIFF
--- a/projects/RabbitMQ.Client/Impl/AutorecoveringConnection.Recording.cs
+++ b/projects/RabbitMQ.Client/Impl/AutorecoveringConnection.Recording.cs
@@ -265,6 +265,8 @@ namespace RabbitMQ.Client.Framing
 
         }
 
+        internal int RecordedBindingsCount => _recordedBindings.Count;
+
         internal async ValueTask RecordBindingAsync(RecordedBinding binding,
             bool recordedEntitiesSemaphoreHeld)
         {
@@ -419,6 +421,15 @@ namespace RabbitMQ.Client.Framing
                 if (!AnyConsumersOnQueue(queue))
                 {
                     _recordedQueues.Remove(queue);
+                    // remove bindings targeting this queue; also cascade to auto-delete exchanges
+                    foreach (RecordedBinding binding in _recordedBindings.ToArray())
+                    {
+                        if (binding.Destination == queue)
+                        {
+                            DoDeleteRecordedBinding(binding);
+                            DoDeleteAutoDeleteExchange(binding.Source);
+                        }
+                    }
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


lukebakken (approved):
Thanks!

PR #1912, opened 2026-03-02
Support for zero-copying and alloc free publish

PauloHMattos:
## Proposed Changes

This PR introduces an architectural refactoring of the outgoing framing pipeline to support zero-copy payload serialization, significantly reducing memory allocations, GC pressure, and Large Object Heap (LOH) fragmentation for high-throughput publishers.

Currently, the client uses a background task to dispatch frames to the socket. To safely cross this asynchronous boundary, message payloads must be copied into rented `ArrayPool<byte>` buffers. While amortized, copying large messages (>85KB) forces the ArrayPool to rent from the Large Object Heap, leading to severe LOH fragmentation and expensive Gen 2 garbage collections under heavy load.

To solve this while maintaining strict backward compatibility, this PR:

**Refactors Framing Lifecycle:** Replaces `RentedMemory` with a custom `OutgoingFrameMemory` struct, deferring payload buffer allocations until the exact moment the thread boundary is crossed.

**Decouples Dispatch Models:** Extracts the existing queue logic into `BackgroundSocketFrameHandler` and introduces a new `InlineSocketFrameHandler`.

**Implements Zero-Copy:** The new `InlineSocketFrameHandler` uses `IBufferWriter<byte>.GetSpan()` to stream the caller's original `ReadOnlyMemory<byte>` payload directly to the socket pipeline. This completely bypasses the ArrayPool rent + copy for the message body.

**Exposes Configuration:** Adds `UseBackgroundFrameWriter` (defaults to true) to `ConnectionFactory`. By defaulting to true, existing applications see no behavioral changes. Users can set this to false to opt-in to inline, zero-copy publishing (trading immediate task return for lower CPU/GC overhead).

## Types of Changes

- [ ] Bug fix (non-breaking change which fixes issue #NNNN)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)

## Checklist

- [x] I have read the `CONTRIBUTING.md` document
- [x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq) (waiting on the email to sign it)
- [x] All tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [x] Any dependent changes have been merged and published in related repositories

## Further Comments

[...]

PR #1913, opened 2026-03-04
Add BasicPublishAsync overloads that accept externally owned body and a disposable

PauloHMattos:
## Proposed Changes

Alternative to #1912. If the mantainers think this is the best approach I will improve this PR body.

## Types of Changes

## Checklist

### Benchmark

<details>
   <summary>Official version</summary>
Body size       : 1 MB
Iterations      : 500
Tasks           : 16
Non-copying     : False
Startup memory  : 16 MB

---  Start  ---
Memory usage: 17 MB
Memory usage: 19 MB
Memory usage: 19 MB
Memory usage: 21 MB
Memory usage: 21 MB
Memory usage: 2230 MB
Memory usage: 2236 MB
Memory usage: 2248 MB
Memory usage: 2249 MB
Memory usage: 2249 MB
Memory usage: 2249 MB
Memory usage: 2250 MB
Memory usage: 2249 MB
Memory usage: 2249 MB

--- Results ---
Avg time        : 10948 ms
Min time        : 8910 ms
Max time        : 11633 ms
Memory          : 2252 MB
Queue length    : 8000 / 8000
Valid messages  : 100 / 100 (first 100 of 8000)
</details>

<details>
   <summary>PR + ReadOnlyMemory</summary>
Body size       : 1 MB
Iterations      : 500
Tasks           : 16
Non-copying     : False
Startup memory  : 16 MB

---  Start  ---
Memory usage: 17 MB
Memory usage: 19 MB
Memory usage: 19 MB
Memory usage: 22 MB
Memory usage: 22 MB
Memory usage: 2119 MB
Memory usage: 2129 MB
Memory usage: 2133 MB
Memory usage: 2134 MB
Memory usage: 2135 MB
Memory usage: 2135 MB
Memory usage: 1782 MB

--- Results ---
Avg time        : 8835 ms
Min time        : 8615 ms
Max time        : 9216 ms
Memory          : 1782 MB
Queue length    : 8000 / 8000
Valid messages  : 100 / 100 (first 100 of 8000)
</details>

<details>
   <summary>PR + IMemoryOwner</summary>
Forked version
Body size       : 1 MB
Iterations      : 500
Tasks           : 16
Non-copying     : True
Startup memory  : 16 MB

---  Start  ---
Memory usage: 18 MB
Memory usage: 19 MB
Memory usage: 20 MB
Memory usage: 22 MB
Memory usage: 23 MB
Memory usage: 35 MB
Memory usage: 37 MB
Memory usage: 37 MB
Memory usage: 37 MB
Memory usage: 38 MB
Memory usage: 40 MB
Memory usage: 41 MB

--- Results ---
Avg time        : 8940 ms
Min time        : 8563 ms
Max time        : 9373 ms
Memory          : 41 MB
Queue length    : 8000 / 8000
Valid messages  : 100 / 100 (first 100 of 8000)
</details>


paulomorgado:
> > A `ReadOnlySequence<T>` might still be holding pooled memory. it would be nice to have a way to be notified when it's been consumed.
> 
> The IMemoryOwner overload gives you that. You can have a wrapper that is aware of the memory you pooled and then you get notified

`IMemoryOwner<T>` holds 1 `Memory<T>`.
`ReadOnlySequence<T>` holds many `ReadOnlyMemory<T>`. But all those may come from a `IMemoryOwner<T>`.

Because there's no out-of-the-box slicing of `IMemoryOwner<T>` and no `IReadOnlyMemoryOwner<T>`, when I want to keep the APIs with strict read-only semantics, I use a `ReadOnlyMemory<T>` and an additional `IDisposable?` memory owner to dispose of that.

Since there's no out-of-the-box implementation of `ReadOnlySequenceSegment<T>`, maybe the consuming code could check if the segment implements `IDisposable` and dispose of it if it does.

danielmarbach:
@PauloHMattos Something like this https://github.com/rabbitmq/rabbitmq-dotnet-client/commit/863508c885f6355e1b3dd9c983a8de62f6e2db78 ( probably needs some more polishing and thinking). I'm quite time-constrained, but it shows your current OutgoingFrame design would be extendable.

danielmarbach:
> In most APIs I build, I opted by sending a `ReadOnlyMemory<byte>` and a `IDisposable?`. It's more versatile.

My understanding is that memory owner is the canonical way of transferring owner ship, see rule 7 and 8 in https://github.com/dotnet/docs/blob/main/docs/standard/memory-and-spans/memory-t-usage-guidelines.md#rule-7-if-you-have-an-imemoryownert-reference-you-must-at-some-point-dispose-of-it-or-transfer-its-ownership-but-not-both but it could also be that I'm misreading the guidance.

paulomorgado (dismissed):

lukebakken (dismissed):
The 🧞 and I did a thorough review of this PR. Here's a summary of what was found and addressed:

**API shape decision**

There was discussion about whether `IMemoryOwner<byte> + int bodyLength` is the right public API shape vs. `ReadOnlyMemory<byte> + IDisposable?`. After consulting the official .NET `Memory<T>` and `Span<T>` usage guidelines (Rules 7 and 8), I've decided that `IMemoryOwner<byte> + int bodyLength` is the approach that will be used. `IMemoryOwner<T>` is the canonical .NET mechanism for transferring ownership of a buffer, and that is exactly the semantics this API needs. Thank you @danielmarbach for finding the official guidance.

**Bugs fixed**

Three commits were pushed on top of the existing branch:

- a4df6655c — `SessionBase.TransmitAsync` was throwing `AlreadyClosedException` without first disposing the caller-transferred `IMemoryOwner<byte>`, leaking the memory owner. Fixed by disposing before throwing. Thanks @paulomorgado 

- 469252de0 — `OutgoingFrame.Dispose` had a redundant `_methodAndHeader = default` assignment after already setting it to `null`, used `!= null` instead of `is not null`, and used `null` instead of `default` for `_body`. Also fixed inconsistent remarks ordering and missing periods in `IChannelExtensions` XML docs. Thanks again @paulomorgado 

- 2ff424930 — The `ReadOnlyMemory<byte>` overload of `SerializeToFrames` was regressed from one allocation to two by copying the body into a `MemoryPool` buffer and delegating to the `IMemoryOwner` overload. Restored the original single-allocation approach. Also improved variable naming (`frameSize` → `payloadSize`, split `remainingBodyBytes` into `bodyLength` + `remainingBodyBytes`) and added explanatory comments to both overloads to make the copy-vs-zero-copy distinction explicit. This is something that the genie noticed and seemed to be worthwhile to revert, even though it means two slightly-different-and-confusing methods.

I'm going to approve this now even though I suspect there will be more discussion.

PauloHMattos:
> I'll leave it up to you all to have a cage match about this 🥊
> 
> My preference is to adopt whatever would be easiest for end-users to use while also following the official guidance. If someone were to use an internet search or AI to figure out how to use this library, which API would be a better fit?
> 
> I'm assuming the maintenance burden of one vs the other is negligible.

@lukebakken haha, hopefully we can avoid the cage match! 🥊

I reached out to five colleagues to get the perspective of developers who aren't as deep into this specific PR as we are (but they are used to working with low level .NET, so may not be an good sample of the average .NET developer). All of them were able to grasp the concepts behind the three options:

- `IMemoryOwner + Length`: The consensus was that this is the most explicit way to convey ownership transfer, but the writable nature of the handle felt "weird" for a payload.

- `ReadOnlyMemory + IDisposable`: This was fully grasped by four of them just by looking at the API shape. The fifth required a quick explanation that we can easily provide via XML documentation.

- Custom wrapper: This was mostly disliked; the general feeling was that while such a type should exist in the BCL, a external library shouldn't be the one to introduce a non-standard wrapper.


I also tested the API with various AI tools to see how they interpreted the semantics. I used the prompt: 
```
Explain in a single and very small paragraph the semantics of this method:
Task Publish(ReadOnlyMemory<byte> body, IDisposable? dispose);
```

- Gemini 3 Flash: 
> This method asynchronously publishes a message payload to a system, accepting the data as a ReadOnlyMemory<byte> for efficient, allocation-free memory handling. The IDisposable parameter allows the method to take ownership of a resource (like a memory-managed buffer) and ensure it is properly cleaned up after the publication task completes or fails.

- Gemini 3 Pro: 
> This method asynchronously transmits a binary payload (body) with minimal allocation overhead, returning a Task to track the operation's completion. The optional dispose parameter provides a mechanism for the method to safely clean up underlying resources—such as a rented memory pool buffer from which the body was sliced—exactly when the publish operation completes and the memory is no longer needed.

- Claude Haiku 4.5:
> This method publishes a message by accepting a read-only byte buffer (`body`) containing the message payload and an optional `IDisposable` resource (`dispose`) that will be cleaned up after publication completes. It returns a `Task` representing the asynchronous publish operation.

- Claude Sonet 4.6:
> `Publish` asynchronously sends a binary message represented as a read-only byte buffer (`body`) to some underlying channel or broker, and accepts an optional `dispose` handle that, if provided, will be released (disposed) once the publish operation completes or is no longer needed — allowing the caller to tie the lifetime of a resource (e.g., a pooled buffer owner) to the completion of the send.

- Claude Opus 4.6:
> This method publishes a message whose body is a read-only chunk of bytes, optionally taking ownership of a disposable resource (via `dispose`) that should be freed once the message has been fully processed or sent. It returns a `Task`, meaning the publish operation is asynchronous and the caller can await its completion.


Since both humans and AIs seem to navigate this pattern well, and it provides the best ergonomics for the caller (baked-in slicing and read-only intent), I'd like to move forward with the ReadOnlyMemory + IDisposable shape. I'll include clear XML documentation to ensure there's no ambiguity regarding the ownership transfer.

Does that sound good to everyone? Can I go ahead and change the PR to the `ReadOnlyMemory + IDisposable` form?
@lukebakken @danielmarbach @paulomorgado

danielmarbach:
I also concluded yesterday it is the best of all options but only because @paulomorgado said it at least three times 🤪

Joking aside all the custom wrapper approaches led me down a deep rabbit hole (pun intended) and are not worth it

projects/RabbitMQ.Client/OutgoingFrame.cs
```diff
@@ -110,7 +109,7 @@ public void Dispose()
             if (memoryOwner is not null)
             {
                 memoryOwner.Dispose();
-                _body?.Dispose();
+                _bodyOwner?.Dispose();
```
danielmarbach: Would it be better to have a dedicated if for the body disposable?

danielmarbach (dismissed):

PauloHMattos:
@lukebakken, do you want any more changes, or would you like me to clean up the PR in some way? The commit history is a bit messy, but it shows our thought process.

michaelklishin:
@PauloHMattos thank you for this work!

Team RabbitMQ is not particularly fussy about commit history. It can be squashed right before or during merging. So don't worry about that part.

projects/RabbitMQ.Client/OutgoingFrame.cs
danielmarbach: I see this has been addressed by @lukebakken

lukebakken:
@PauloHMattos I pushed a couple commits that you should review. It appears that you used `IDisposable bodyOwner` when (I think) you intended to use `IDisposable? bodyOwner`. I changed that and did a couple other small things.

PauloHMattos:
@lukebakken I was in doubt whether it should be nullable or not. But I think you are right and nullable is the way to go since the whole implementation already handles it.
In the future we could also remove the overload with just the ReadOnlyMemory from the interface (move it to be an extension) and keep the interface more lean.

About the other things, everything looks good to me

lukebakken (dismissed):

lukebakken:
@danielmarbach @paulomorgado anything from either of you?

projects/RabbitMQ.Client/Impl/Channel.BasicPublish.cs
```diff
@@ -117,27 +98,21 @@ await MaybeAcquirePublisherConfirmationLockAsync(cancellationToken)
...
                     : default;
 
-                ulong publishSequenceNumber = 0;
-                if (publisherConfirmationInfo is not null)
-                {
-                    publishSequenceNumber = publisherConfirmationInfo.PublishSequenceNumber;
-                }
+                ulong publishSequenceNumber = publisherConfirmationInfo?.PublishSequenceNumber ?? 0;
 
                 BasicProperties? props = PopulateBasicPropertiesHeaders(basicProperties, sendActivity, publishSequenceNumber);
```
paulomorgado: If any of this throws and `bodyOwner` is not `null`, it won't be disposed.

projects/RabbitMQ.Client/Impl/SocketFrameHandler.cs
paulomorgado, line 278: When `FlushAsync` throws (network error), the write loop exits. Any `OutgoingFrame` instances already queued in `_channelReader` are never dequeued or disposed — their `_bodyOwner` and rented
method+header memory leak permanently.

Fix: Add while `(_channelReader.TryRead(out var leftover)) leftover.Dispose();` in a finally block around the write loop.

paulomorgado, line 254: `_channelWriter.WriteAsync(frames, cancellationToken)` — if the bounded channel (128 slots) is full and the token is cancelled, the `OutgoingFrame` copy held by the BCL channel's internal async state
machine is abandoned without `Dispose()` being called.

Fix: Wrap in try/catch to dispose frames on cancellation/failure.

projects/RabbitMQ.Client/Impl/IFrameHandler.cs
```diff
@@ -68,6 +68,6 @@ internal interface IFrameHandler
 
         ValueTask SendProtocolHeaderAsync(CancellationToken cancellationToken);
 
-        ValueTask WriteAsync(RentedMemory frames, CancellationToken cancellationToken);
+        ValueTask WriteAsync(OutgoingFrame frames, CancellationToken cancellationToken);
```
paulomorgado, line 71: Takes the 48-byte struct by value. When _closed == true, it disposes the copy, not the caller's original. This is mitigated because `Connection.WriteAsync`
doesn't use the original after the call, but it's both semantically incorrect and costs an unnecessary copy.

Fix: Consider `in OutgoingFrame` for correctness and efficiency.

paulomorgado (commented):
Test Coverage Gaps (all models noted)

 - No test for `bodyOwner` disposal on failed publish (channel closed, nack, cancellation)
 - No test for `bodyOwner` disposal when the connection drops mid-publish
 - The existing `TestMemoryOwnerBody` only covers the happy path

PauloHMattos:
I will look at @paulomorgado feedback and make the fixes next weekend. Thanks

danielmarbach:
@PauloHMattos I tried to address those. We probably also need to resolve the conflicts on the API changes. I might be able to do this later

danielmarbach:
@PauloHMattos I force pushed after a rebase

projects/RabbitMQ.Client/Impl/SessionBase.cs
```diff
@@ -128,21 +129,22 @@ public virtual ValueTask TransmitAsync<T>(in T cmd, CancellationToken cancellati
...
             {
+                bodyOwner?.Dispose();
                 ThrowAlreadyClosedException();
             }
 
-            RentedMemory bytes = Framing.SerializeToFrames(ref Unsafe.AsRef(in cmd), ref Unsafe.AsRef(in header), body, ChannelNumber, Connection.MaxPayloadSize);
+            OutgoingFrame bytes = Framing.SerializeToFrames(ref Unsafe.AsRef(in cmd), ref Unsafe.AsRef(in header), body, bodyOwner, ChannelNumber, Connection.MaxPayloadSize);
             RabbitMQActivitySource.PopulateMessageEnvelopeSize(Activity.Current, bytes.Size);
             return Connection.WriteAsync(bytes, cancellationToken);
         }
```
paulomorgado, line 175: Isn't there the chance for leaking the `bodyOwner` on error?

```csharp
 public ValueTask TransmitAsync<TMethod, THeader>(..., IDisposable? bodyOwner, ...)
 {
     if (!IsOpen && cmd.ProtocolCommandId != ProtocolCommandId.ChannelCloseOk)
     {
         bodyOwner?.Dispose();
         ThrowAlreadyClosedException();
     }
 
     OutgoingFrame bytes = default;
     try
     {
         bytes = Framing.SerializeToFrames(..., bodyOwner, ...);
         RabbitMQActivitySource.PopulateMessageEnvelopeSize(Activity.Current, bytes.Size);
         return Connection.WriteAsync(bytes, cancellationToken);
     }
     catch
     {
         // If SerializeToFrames failed: bytes is default → Dispose() is no-op,
         // but bodyOwner was never captured → dispose it directly.
         // If SerializeToFrames succeeded: bytes holds bodyOwner → Dispose() covers both.
         bytes.Dispose();
         if (bytes.Size == 0)
             bodyOwner?.Dispose();
         throw;
     }
 }
```

danielmarbach: @paulomorgado That would only capture synchronous exceptions. Returning a task or value task and expecting a try / catch to fire is a recipe for disaster. Let me have a closer look

danielmarbach: I don't think we have a problem here. The real guard against `bodyOwner` leaking from the `BasicPublishCoreAsync` to `TransmitAsync` path is already the `bodyOwnerTransferred` fix we put in `BasicPublishCoreAsync`. 

For `TransmitAsync` itself: once `SerializeToFrames` succeeds, `OutgoingFrame` owns `bodyOwner`, and `SocketFrameHandler` is responsible for disposing of it in all paths (already fixed). The only true gap was `PopulateMessageEnvelopeSize` throwing after `SerializeToFrames` but that's a `Activity` tagging call with no realistic throw surface, and patching it with an incorrect try/catch or unnecessary introducing awaits here seems worse than leaving it

paulomorgado: I'm starting to think I analyzed this wrong.

Maybe it should be as simple as:

```csharp
public ValueTask TransmitAsync<TMethod, THeader>(..., IDisposable? bodyOwner, ...)
{
    try
    {
        // ...
    }
    finally
    {
        bodyOwner?.Dispose();
    }
}
```

But that would require the method to be async.

Or not, with some extra work and creating an async local method if `bodyOwner` is not `null` and a sync one if it is.

paulomorgado: Asked Copilot and this seems to be the recommendation:

### Problem

`TransmitAsync` has no protection around the synchronous code between the `!IsOpen` guard and `Connection.WriteAsync`. If `SerializeToFrames` or `PopulateMessageEnvelopeSize` throws, `bodyOwner` (and possibly the rented method+header memory) leaks:

```csharp
// Current code — no try/catch around these calls
OutgoingFrame bytes = Framing.SerializeToFrames(..., body, bodyOwner, ...);  // can throw (OOM)
RabbitMQActivitySource.PopulateMessageEnvelopeSize(Activity.Current, bytes.Size);  // can throw
return Connection.WriteAsync(bytes, cancellationToken);  // async faults handled by write pipeline ✅
```

Meanwhile, `BasicPublishCoreAsync` has already set `bodyOwnerTransferred = true` before calling `ModelSendAsync`, so its outer `finally` won't clean up either.

### Failure analysis

| Failure point | Who disposes? |
|---|---|
| `!IsOpen` early exit | `bodyOwner?.Dispose()` before throw ✅ |
| `SerializeToFrames` throws (e.g., OOM) | **Nobody** ❌ — frame never created, `bodyOwner` not captured |
| `PopulateMessageEnvelopeSize` throws | **Nobody** ❌ — entire `OutgoingFrame` (including `bodyOwner`) leaks |
| `Connection.WriteAsync` throws synchronously | **Nobody** ❌ — same as above |
| `Connection.WriteAsync` returns faulted `ValueTask` | `WriteAsyncCore` catch ✅ |
| `WriteLoopAsync` crashes | `WriteLoopAsync` finally drain ✅ |

### Recommended fix

```csharp
public ValueTask TransmitAsync<TMethod, THeader>(
    in TMethod cmd, in THeader header,
    ReadOnlyMemory<byte> body, IDisposable? bodyOwner,
    CancellationToken cancellationToken = default)
    where TMethod : struct, IOutgoingAmqpMethod
    where THeader : IAmqpHeader
{
    if (!IsOpen && cmd.ProtocolCommandId != ProtocolCommandId.ChannelCloseOk)
    {
        bodyOwner?.Dispose();
        ThrowAlreadyClosedException();
    }

    OutgoingFrame bytes = default;
    try
    {
        bytes = Framing.SerializeToFrames(
            ref Unsafe.AsRef(in cmd), ref Unsafe.AsRef(in header),
            body, bodyOwner, ChannelNumber, Connection.MaxPayloadSize);
        RabbitMQActivitySource.PopulateMessageEnvelopeSize(Activity.Current, bytes.Size);
        return Connection.WriteAsync(bytes, cancellationToken);
    }
    catch
    {
        // If SerializeToFrames succeeded (Size > 0), bytes owns both
        // the rented method+header memory and bodyOwner — Dispose covers both.
        // If SerializeToFrames failed, bytes is default — Dispose is a no-op,
        // so we must dispose bodyOwner directly.
        bytes.Dispose();
        if (bytes.Size == 0)
        {
            bodyOwner?.Dispose();
        }

        throw;
    }
}
```

### Why this works

| Failure | What happens |
|---|---|
| `!IsOpen` | `bodyOwner?.Dispose()` before throw ✅ |
| `SerializeToFrames` throws | `bytes` is `default` → `Dispose()` is no-op → `bodyOwner?.Dispose()` ✅ |
| `PopulateMessageEnvelopeSize` throws | `bytes.Dispose()` frees frame + bodyOwner ✅ |
| `Connection.WriteAsync` throws synchronously | Same as above ✅ |
| `Connection.WriteAsync` returns faulted `ValueTask` | `WriteAsyncCore` catch disposes frame ✅ (try/catch not involved) |
| Happy path | Frame flows through write pipeline → `WriteLoopAsync` disposes ✅ |

### Practical severity

Low-to-medium. `SerializeToFrames` is arithmetic + `MemoryPool.Rent` + buffer writes — only OOM or catastrophic failure triggers this. `PopulateMessageEnvelopeSize` is `Activity?.SetTag` — essentially cannot throw. But this is the last gap in an otherwise complete ownership chain.

danielmarbach: Like I mentioned above my assessment was it is not worth it because when those edge cases happen you are in catastrophic territory and returning buffers is the least of your worry since you need to restart anyway.

That being said I might be missing something and I'm happy to be convinced otherwise

lukebakken: I'm fine adding the changes suggested by the 🧞 - @PauloHMattos ?

danielmarbach: I don't think this is a good change.

`TransmitAsync` is a very central method that every session interaction flows through, so we should be careful about adding code and control-flow complexity there unless the payoff is very clear.

In this case the proposed `try/catch` only covers the synchronous portion up to the point where `WriteAsync(...)` returns its `ValueTask`. It does not protect the async lifetime of the write operation itself, so the practical benefit is narrow.

On top of that, I think the pattern is misleading. A `try/catch` wrapped around `return Connection.WriteAsync(...)` is not a shape people usually associate with robust async error handling. For anyone familiar with `async`/`await`, it creates the impression that the write path is being guarded more comprehensively than it really is, when in reality it only catches synchronous pre-return failures.

So from my point of view this change makes a hot path bigger, introduces a somewhat surprising pattern, and still only addresses a very small synchronous gap. That does not feel like a good tradeoff to me.

But again I might be missing an important detail and I'm also not a 🧞

lukebakken:
I should have time to wrap this up and the [other `7.3.0` release items](https://github.com/rabbitmq/rabbitmq-dotnet-client/milestone/76) starting in May. I haven't forgotten.

PauloHMattos:
Sorry for being MIA, I had some rough weeks at work, but I will catch up with the latest changes here and help where I can to wrap this PR

lukebakken:
## Review update

The 🧞‍♂️ (genie) and I spent some time reviewing this PR in depth today. I pushed four commits while going through it. Here's the genie's summary of the work:

----

**bf8b43aec Revert unintentional toxiproxy-netcore submodule bump**

The initial commit of this branch moved the `projects/toxiproxy-netcore` submodule from 9aeec29a9 to 13e7757bc. That change is unrelated to the `BasicPublishAsync` work and wasn't justified in the PR body, so I reverted the submodule pointer to match main.

**d893417eb Fix stale `IChannelExtensions` XML doc cross-references**

Five of the new `bodyOwner`-accepting extension overloads had `<c>IChannel.BasicPublishAsync{TProperties}(..., IMemoryOwner{byte}, int, CancellationToken)</c>` cross-references left over from the earlier API shape. I updated them to `<see cref="...">` with the current `ReadOnlyMemory{byte}, IDisposable, CancellationToken` signatures, which also gets them validated by the XML doc compiler.

**3bd79613a Close `bodyOwner` leak window in `SessionBase.TransmitAsync`**

This is the only substantive fix in the set. `BasicPublishCoreAsync` sets `bodyOwnerTransferred = true` before awaiting `ModelSendAsync`, so the outer `finally` no longer disposes `bodyOwner`. The actual ownership transfer happens inside `TransmitAsync` when `SerializeToFrames` captures `bodyOwner` in the returned `OutgoingFrame`. If anything throws between those two points (most realistically `MemoryPool.Rent` under OOM inside `SerializeToFrames`), `bodyOwner` leaked.

I went with @paulomorgado's suggested try/catch rather than plumbing a transferred-signal back up through `ModelSendAsync`. The catch discriminates by `bytes.Size == 0`: on a default `OutgoingFrame` the owner was never captured, so dispose it directly; otherwise the frame owns it and `bytes.Dispose()` releases both. @danielmarbach, I think your objection about hot-path cost and surprising shape is real but solvable with a comment. The scope of the try/catch is precisely the synchronous window that isn't already covered by `SocketFrameHandler.WriteAsyncCore`'s own catch on the async side; I put an explanatory comment in place so that intent is clear.

I also considered @paulomorgado's suggestion to change `IFrameHandler.WriteAsync` to take `in OutgoingFrame`, but declined it. The change bumps against CS1988 (async methods can't have `in` parameters), so `WriteAsyncCore` must still take the frame by value, which makes the savings one struct copy at the non-async hop and adds an `Unsafe.AsRef` dance in the `_closed` path (because `OutgoingFrame.Dispose` mutates fields). Marginal benefit, real readability cost.

Other items from @paulomorgado's earlier comments (leaks in `WriteLoopAsync` / `WriteAsyncCore`, the pre-close disposal, style fixes in `OutgoingFrame.Dispose`) are already resolved in the branch; I verified each against head before moving on.

**00e4d715f Use RabbitMQ 4.2 until we can make tests 4.3 compatible**

Unrelated to this PR but needed to unblock CI. When I ran the integration tests they failed across ~20 cases with `INTERNAL_ERROR - Feature transient_nonexcl_queues is deprecated`. RabbitMQ 4.3 moved a handful of deprecated features from `permitted_by_default` to `denied_by_default`, and a number of our tests (plus one code path in `AutorecoveringChannel.BasicQosAsync(..., global: true, ...)` and the public `Headers.XQueueMode` constant) run into them. I pinned CI to `rabbitmq:4.2-management` on Ubuntu and `4.2.6` on Windows so this PR can get finished.

Full 4.3 compatibility work is tracked in #1931, targeted at 7.3.0.

With these changes and the earlier ones, from my side I think this is ready. I'd like to keep this PR open briefly for everyone to sanity-check the `TransmitAsync` try/catch before I mark it approved.

danielmarbach:
@lukebakken I added your Genie sentence as reference for funsies 

> I think your objection about hot-path cost ... is real but solvable with a comment

How does adding a comment solve something on the hot path? Does the genie have a magic dotnet feature that turns comments into fairy dust?

Joking aside my point was the try catch has costs and we are comparing the costs there on the hot path against an extreme edge case that is quite likely catastrophic and then returning pooled objects is the least of your problems. So for me the real thing would have been laying out the edge case and the mitigation scenarios vs the cost for the througput in the hot path due to the larger call stack. A benchmark would have shown that. So the trade-off of introducing the catch here is that we are trying our best best to always return buffers and favour maintainability and extensibility here over the potential costs which we still have no baseline. I'm ok with that but I want to highlight the genie comment doesn't do justice in that regard

lukebakken:
I'll run benchmarks to compare that change.

lukebakken:
@danielmarbach here are some numbers!

The genie and I wrote a small publish-throughput harness ([GH-1913-Benchmark](https://github.com/rabbitmq/rabbitmq-dotnet-client/tree/gh-1913-benchmark/projects/Applications/GH-1913-Benchmark)) and ran it in an EC2 test environment. The benchmark program ran using SDK 8 on a `t4g.medium` instance, while RabbitMQ (on the `main` code) ran on three `m7g.large` instances in a cluster. Quorum queue, `mandatory: true`, publisher confirms enabled but tracking OFF (so each publish returns as soon as the frame is enqueued, not after the broker confirms - this should maximize the time spent in the client-side hot path). 5 iterations of 30 seconds each, 5 second warmup, 256-byte body.

Three branches, identical except for the code under test:

- [gh-1913-benchmark](https://github.com/rabbitmq/rabbitmq-dotnet-client/tree/gh-1913-benchmark) - `main`, no `PR #1913` changes. Baseline.
- [bench-withoutcatch](https://github.com/rabbitmq/rabbitmq-dotnet-client/tree/bench-withoutcatch) - `PR #1913` at d893417eb (the commit just before the try/catch was added). Has the `IDisposable? bodyOwner` machinery but no try/catch in `TransmitAsync`.
- [bench-withcatch](https://github.com/rabbitmq/rabbitmq-dotnet-client/tree/bench-withcatch) - `PR #1913` at 00e4d715f (branch head, includes 3bd79613a). Has the try/catch.

### Results

| Metric | main | withoutcatch | withcatch |
|---|---|---|---|
| Median msg/s | 25022 | 24951 | 24670 |
| Median B/msg | 109 | 228 | 226 |
| Median gen0/iter | ~2 | 4 | 4 |

### Try/catch cost (withcatch vs withoutcatch)

- **Throughput:** 24670 vs 24951 = -1.1%. Both builds have ~5% run-to-run spread. Not statistically significant.
- **Allocations per message:** 226 vs 228 B/msg. Effectively identical.
- **GC:** identical (3-4 gen0 per 30s iteration on both).

The try/catch doesn't seem to have measurable impact on either throughput or allocations.

### PR #1913 overall cost (either PR branch vs main)

- **Throughput:** ~25000 msg/s on all three. No measurable difference.
- **Allocations:** 109 -> ~227 B/msg (+108%). This is the `OutgoingFrame` / `IDisposable? bodyOwner` ownership machinery, present on both PR branches equally.
- **GC:** gen0 roughly doubled (2 -> 4 per 30s).

Do we care about the per-message allocation increase from the PR?

danielmarbach:
These are really good insights. We pay more for the abstraction/ unification of the code than for the additional try / catch machinery. Thanks for doing this extra step 😍 I guess now we can make informed trade-offs and it shows where the majority of the costs are

I'll look at the Benchmarks later

danielmarbach:
I have looked at the benchmark and at the cause of the additional allocations on the regular path. These are quite likely caused by `MemoryPool<byte>.Shared` whose `Rent() `method allocates a new `IMemoryOwner` object every call. We could remove this allocation by going back to `ArrayPool.Rent `on the non-body owner path which would require us to create an additional constructor overload on OutgoingFrame and deal with the fact that depending on the "path" taken, another pool would be used. I don't really like this approach because if we decide to ever allow the pool to be controlled from the outside via a configuration setting, this dualism in the code path would make this extensibility story a nightmare.

Alternatively, we could create a small MemoryPool<byte> implementation that internally uses a ConcurrentStack of MemoryOwners to amortize those costs which would unify the code path and allow future extensibility. There would be "warmup" overhead too from that rental and on massive parallel publishes the ConcurrentStack might cause other overheads too.

Thoughts?

paulomorgado:
> I have looked at the benchmark and at the cause of the additional allocations on the regular path. These are quite likely caused by `MemoryPool<byte>.Shared` whose `Rent() `method allocates a new `IMemoryOwner` object every call. We could remove this allocation by going back to `ArrayPool.Rent `on the non-body owner path which would require us to create an additional constructor overload on OutgoingFrame and deal with the fact that depending on the "path" taken, another pool would be used. Thoughts?

That sounds good. If the code is in full control of the rented "memory", there's no need for a memory owner.

danielmarbach:
@paulomorgado I updated my comment with new revelations

danielmarbach:
I actually think now we can get rid of MemoryPool. Let me try something

danielmarbach (approved):

danielmarbach:
@PauloHMattos @paulomorgado @lukebakken Pushed a small tweak that should get rid of those allocations 🤞

lukebakken:
@danielmarbach nice! I'll do a quick re-run of those benchmarks.

lukebakken:
Here are the results for 9d49d4a9b:

### Results

| Metric | main (baseline) | bench-prhead (with 9d49d4a9b) | bench-withcatch (yesterday, before fix) |
|---|---|---|---|
| Median msg/s | 24672 | 24473 (-0.8%) | 24670 |
| Median B/msg | 121 | 136 (+12%) | 226 (+108%) |
| Median gen0/iter | 2 | 2 | 4 |

Same setup as before: quorum queue, mandatory true, publisher confirms on but tracking off, 256-byte body, 5 iterations of 30s each, 5s warmup.

The allocation fix did what we hoped. The PR's per-message allocation overhead dropped from +108% to +12%. GC is back to parity with main (2 gen0 per 30-second iteration vs 4 yesterday). Throughput is unchanged within noise.

The residual +15 B/msg vs main is likely the `IDisposable? bodyOwner` field on `OutgoingFrame`, the `bodyOwnerTransferred` tracking in `BasicPublishCoreAsync`, and the async state machine differences. Small enough that further optimization probably isn't worth the complexity.

LGTM on this from a perf standpoint.

lukebakken (approved):

lukebakken:
@PauloHMattos thank you for this contribution, and thank you @danielmarbach and @paulomorgado for all of your work.

PauloHMattos:
I've just sat down to take a lot at this PR. Great to se that was merged.
Thank you all for the help getting this done.

Hopefully I will be able to contribute more in the future (I have my 👀 on the support for ReadOnlySequence<byte> now )

danielmarbach:
Folks I must say this was an outstanding collaboration. A lot of fun and despite different opinions being held we all worked together to get common understanding, listened to each other's point and found good ways to move forward. This keeps me motivated to lurk around and make random comments, be warned 🤪😂

PR #1922, opened 2026-03-28
Update `ExchangeType` to include more core and tier 1 plugin-provided types

michaelklishin:
`x-modulus-hash` is an built-in exchange type as of RabbitMQ `4.3.0`.

### DIFF
--- a/projects/RabbitMQ.Client/ExchangeType.cs
+++ b/projects/RabbitMQ.Client/ExchangeType.cs
@@ -64,7 +64,39 @@ namespace RabbitMQ.Client
         /// </summary>
         public const string Topic = "topic";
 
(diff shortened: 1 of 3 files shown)


michaelklishin:
@lukebakken what's the right way to submit a backport PR for `7.x` these days?

lukebakken:
`main` is 7.x so you're all good here.

lukebakken (approved):

michaelklishin:
> `main` is 7.x so you're all good here.

Well, still. Do you want me to set up Mergify? (for those less familiar with RabbitMQ development: this is what RabbitMQ server repo has been using for a few years with generally a very positive experience).

lukebakken:
I think we're good until we actually need it 😸

PR #1932, opened 2026-04-25
Close write-loop channel on crash to unblock publishers

### LINKED ISSUE #1930: BasicPublishAsync hangs after RabbitMQ server restart
### Discussed in https://github.com/rabbitmq/rabbitmq-dotnet-client/discussions/1929

<sup>Originally posted by **MarcoBarbieri90** April 22, 2026</sup>
### **Environment**

- Client: RabbitMQ .NET Client 7.2.1
- Server (tested on): RabbitMQ 3.13.4 + Erlang 26.2.5.1 / RabbitMQ 4.0.25 + Erlang 27.3.4.10
- Server configuration: Default
- OS: Windows (RabbitMQ running as a Windows Service)

### **Description**

We are experiencing a hang in BasicPublishAsync after restarting the RabbitMQ Windows service while publishing at high throughput (~50,000 messages every 7 seconds) with AutomaticRecoveryEnabled = true. The severity of the issue depends on whether publisher confirms are enabled or not.

When publisher confirms are disabled (publisherConfirmationsEnabled: false, publisherConfirmationTrackingEnabled: false), BasicPublishAsync silently hangs for approximately 3 minutes before automatically resuming. During this time, IConnection events (ConnectionShutdownAsync, RecoverySucceededAsync, ConnectionRecoveryErrorAsync) are not fired immediately — they are delayed by the same ~3 minutes, making them effectively useless for any timely recovery logic.

When publisher confirms are enabled (publisherConfirmationsEnabled: true, publisherConfirmationTrackingEnabled: false), the situation is worse. The IConnection events are still delayed by ~3 minutes and the connection does eventually recover, but BasicPublishAsync hangs indefinitely — even after full recovery, the publisher never resumes and no exception is ever thrown.

### **Root Cause**
Both scenarios share the same underlying root cause, observed via the Visual Studio Task Debugger.
During the hang, the Task window reveals the following deadlock-like situation:

**In the publisher without confirms**, the Task window reveals the following call chain blocked at shutdown:
Image

HandleConnectionCloseAsync(cmd, cancellationToken) inside Channel.cs is blocked attempting to write a frame to the SocketFrameHandler.
The internal Channel (bounded capacity: 128, FullMode: Wait) is full.
The reader task (WriteLoopAsync()) that drains this channel exits immediately when the RabbitMQ Windows service is stopped, leaving the channel permanently full and all writers blocked.

This means that when the server goes down, the write loop exits but the channel is never drained, causing any subsequent write attempt to wait indefinitely until automatic recovery eventually reconnects and unblocks the pipeline (~3 minutes later).

**In the publisher confirms scenario**, an additional problem compounds the above. 
The Task window reveals the following call chain blocked at shutdown:
Image

MaybeHandlePublisherConfirmationTcsOnChannelShutdownAsync
  └─> SemaphoreSlim.WaitUntilCountOrTimeoutAsync   ← blocked indefinitely

OnChannelShutdownAsync triggers MaybeHandlePublisherConfirmationTcsOnChannelShutdownAsync, which tries to acquire a SemaphoreSlim to complete or cancel all pending publisher confirm TaskCompletionSources.
This SemaphoreSlim is never released during shutdown (likely held by a write operation that is itself blocked — same root cause as issue #XXXX).
As a result, all pending BasicPublishAsync calls remain awaiting a confirm that will never arrive, even after the connection and channel are recovered.

### Reproduction steps

1. Start a RabbitMQ server (3.13.x or 4.0.x) with default configuration.
2. Run the attached publisher targeting a quorum queue at ~50,000 msg/7s throughput, with AutomaticRecoveryEnabled = true.
3. While publishing, restart the RabbitMQ Windows service.

**Scenario A** — _confirms disabled (publisherConfirmationsEnabled: false, publisherConfirmationTrackingEnabled: false):_

[Publisher - No Confirms.txt](https://github.com/user-attachments/files/26964519/Publisher.-.No.Confirms.txt)

Observe:

After ~3 minutes, IConnection events fire and the connection recovers.
BasicPublishAsync hangs silently for ~3 minutes.
After ~3 minutes, publishing resumes automatically.

**Scenario B** — confirms enabled (publisherConfirmationsEnabled: true, publisherConfirmationTrackingEnabled: false):

[Publisher - Confirms.txt](https://github.com/user-attachments/files/26964524/Publisher.-.Confirms.txt)

Observe:

BasicPublishAsync hangs.
After ~3 minutes, IConnection events fire and the connection recovers.
BasicPublishAsync remains blocked indefinitely — the publisher never resumes.
No exception is ever thrown.

### Expected behavior

1. IConnection events should fire promptly when the server goes down, not after a multi-minute delay. 
2. BasicPublishAsync should either throw an exception immediately upon disconnection — allowing the caller to handle the error and retry — or recover quickly without hanging.

### Additional context

No workaround has been found so far.
_Are there any recommended workarounds while this is being investigated?_</div>

lukebakken:
Fixes #1930.

When the background write loop in `SocketFrameHandler` crashes on a socket failure, the bounded `Channel<RentedMemory>` was left open, leaving any subsequent producer blocked indefinitely on a full channel whose reader had died. With publisher confirms enabled this deadlocked the shutdown handler, because the holder of the publisher-confirm semaphore was a blocked publisher.

On write-loop failure, `TryComplete` the channel with the exception. Pending and future `WriteAsync` callers see a
`ChannelClosedException` whose `InnerException` is the socket fault, and release the publisher-confirm semaphore promptly, letting shutdown and recovery proceed. Any frames still buffered in the channel when the loop exits are drained and disposed so their pooled buffers return to `ArrayPool`.

Also dispose the caller's `RentedMemory` when
`_channelWriter.WriteAsync` faults (post-completion). Previously the struct was lost along with its pooled array. The hot path stays allocation-free.

### DIFF
--- a/projects/RabbitMQ.Client/Impl/SocketFrameHandler.cs
+++ b/projects/RabbitMQ.Client/Impl/SocketFrameHandler.cs
@@ -175,7 +175,13 @@ namespace RabbitMQ.Client.Impl
                     .ConfigureAwait(false);
                 try
                 {
-                    _channelWriter.Complete();
+                    // TryComplete rather than Complete: WriteLoopAsync may have
+                    // already completed the writer with an exception (see
+                    // issue #1930). Complete() would throw InvalidOperationException
+                    // in that race, the outer catch would swallow it, and
+                    // `await _writerTask` below would be skipped, leaving the
+                    // write-loop exception unobserved.
+                    _channelWriter.TryComplete();
                     if (_writerTask != null)
                     {
                         await _writerTask.ConfigureAwait(false);
--- a/projects/RabbitMQ.Client/Impl/SocketFrameHandler.cs
+++ b/projects/RabbitMQ.Client/Impl/SocketFrameHandler.cs
@@ -313,15 +313,16 @@ namespace RabbitMQ.Client.Impl
                 // TryComplete rather than Complete: CloseAsync may have raced
                 // us and already completed the channel.
                 _channelWriter.TryComplete(ex);
-
+                throw;
+            }
+            finally
+            {
                 // Drain any leftover frames so their pooled buffers return to
                 // the array pool rather than waiting for GC.
                 while (_channelReader.TryRead(out OutgoingFrame leftover))
                 {
                     leftover.Dispose();
                 }
-
-                throw;
             }
         }
     }
--- a/projects/RabbitMQ.Client/Impl/SocketFrameHandler.cs
+++ b/projects/RabbitMQ.Client/Impl/SocketFrameHandler.cs
@@ -240,16 +240,31 @@ namespace RabbitMQ.Client.Impl
             return WriteAsyncCore(frames, cancellationToken);
         }
 
-        private async ValueTask WriteAsyncCore(OutgoingFrame frames, CancellationToken cancellationToken)
+        private ValueTask WriteAsyncCore(OutgoingFrame frames, CancellationToken cancellationToken)
         {
-            try
+            ValueTask writeTask = _channelWriter.WriteAsync(frames, cancellationToken);
+            if (writeTask.IsCompletedSuccessfully)
             {
-                await _channelWriter.WriteAsync(frames, cancellationToken).ConfigureAwait(false);
+                return default;
             }
-            catch
+
+            // The channel accepted the frame asynchronously, or it rejected the
+            // write (e.g. because WriteLoopAsync completed the writer with an
+            // exception). In the rejection case `frames` never reached the
+            // reader, so this method owns the dispose. See issue #1930.
+            return AwaitAndDisposeOnFault(writeTask, frames);
+
+            static async ValueTask AwaitAndDisposeOnFault(ValueTask task, OutgoingFrame frames)
             {
-                frames.Dispose();
-                throw;
+                try
+                {
+                    await task.ConfigureAwait(false);
+                }
+                catch
+                {
+                    frames.Dispose();
+                    throw;
+                }
             }
         }
 
@@ -281,14 +296,26 @@ namespace RabbitMQ.Client.Impl
             catch (Exception ex)
             {
                 ESLog.Error("Background socket write loop has crashed", ex);
-                throw;
-            }
-            finally
-            {
+
+                // Fail both already-blocked and subsequent WriteAsync callers
+                // with ChannelClosedException(innerException: ex). Without
+                // this, producers block indefinitely on a full channel whose
+                // reader has died, and any holder of a publisher-confirm
+                // semaphore (Channel.BasicPublish.cs) blocks the shutdown
+                // handler in turn. See issue #1930.
+                //
+                // TryComplete rather than Complete: CloseAsync may have raced
+                // us and already completed the channel.
+                _channelWriter.TryComplete(ex);
+
+                // Drain any leftover frames so their pooled buffers return to
+                // the array pool rather than waiting for GC.
                 while (_channelReader.TryRead(out OutgoingFrame leftover))
                 {
                     leftover.Dispose();
                 }
+
+                throw;
             }
         }
     }
--- a/projects/RabbitMQ.Client/Impl/SocketFrameHandler.cs
+++ b/projects/RabbitMQ.Client/Impl/SocketFrameHandler.cs
@@ -252,9 +252,9 @@ namespace RabbitMQ.Client.Impl
             // write (e.g. because WriteLoopAsync completed the writer with an
             // exception). In the rejection case `frames` never reached the
             // reader, so this method owns the dispose. See issue #1930.
-            return AwaitAndDisposeOnFault(writeTask, frames);
+            return AwaitAndDisposeOnException(writeTask, frames);
 
-            static async ValueTask AwaitAndDisposeOnFault(ValueTask task, OutgoingFrame frames)
+            static async ValueTask AwaitAndDisposeOnException(ValueTask task, OutgoingFrame frames)
             {
                 try
                 {
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.claude/
 *.orig
 *.log
 _site/


projects/RabbitMQ.Client/Impl/SocketFrameHandler.cs
```diff
@@ -281,6 +296,25 @@ await _pipeWriter.FlushAsync()
...
+                // TryComplete rather than Complete: CloseAsync may have raced
+                // us and already completed the channel.
+                _channelWriter.TryComplete(ex);
+
+                // Drain any leftover frames so their pooled buffers return to
+                // the array pool rather than waiting for GC.
+                while (_channelReader.TryRead(out OutgoingFrame leftover))
+                {
+                    leftover.Dispose();
+                }
```
PauloHMattos, line 325: Now that #1913 is merged and the finally is disposing of the leftover this is redundant

lukebakken: Ah, thanks for noticing.

projects/RabbitMQ.Client/Impl/SocketFrameHandler.cs
```diff
@@ -281,14 +296,26 @@ await _pipeWriter.FlushAsync()
...
+                // Fail both already-blocked and subsequent WriteAsync callers
+                // with ChannelClosedException(innerException: ex). Without
+                // this, producers block indefinitely on a full channel whose
+                // reader has died, and any holder of a publisher-confirm
+                // semaphore (Channel.BasicPublish.cs) blocks the shutdown
+                // handler in turn. See issue #1930.
+                //
+                // TryComplete rather than Complete: CloseAsync may have raced
+                // us and already completed the channel.
+                _channelWriter.TryComplete(ex);
```
PauloHMattos, line 315: Since we are completing the channel here, should `CloseAsync`also be updated to use `_channelWriter.TryComplete()` instead of `_channelWriter.Complete()`?

I'm not sure if this case can happen, but if the write loop crashes before `CloseAsync` runs, the  `_channelWriter.Complete()` in `CloseAsync` will throw an `InvalidOperationException` (channel already completed). The inner `catch { // ignore }` swallows it, but that means the `await _writerTask` is never reached. 
This will leave the `_writerTask` exception unobserved.

The exception is already logged by `ESLog.Error` in `WriteLoopAsync`, so it will be reported, but in this case a `TaskScheduler.UnobservedTaskException` would still fire when the task is GC'd and may cause some confusion.

projects/RabbitMQ.Client/Impl/SocketFrameHandler.cs
```diff
@@ -240,16 +240,31 @@ public ValueTask WriteAsync(OutgoingFrame frames, CancellationToken cancellation
...
             }
-            catch
+
+            // The channel accepted the frame asynchronously, or it rejected the
+            // write (e.g. because WriteLoopAsync completed the writer with an
+            // exception). In the rejection case `frames` never reached the
+            // reader, so this method owns the dispose. See issue #1930.
+            return AwaitAndDisposeOnFault(writeTask, frames);
+
+            static async ValueTask AwaitAndDisposeOnFault(ValueTask task, OutgoingFrame frames)
```
PauloHMattos: I was wondering what `Fault` means in this context. I think would be better if the name was `AwaitAndDisposeOnException`

PauloHMattos (approved):
Just some observations, but looks good to me!

Also, can we convert the repro code into a regression test?
(I can do it over the weekend if you like)

lukebakken:
@PauloHMattos please, feel free. Open a PR for this PR 😸

projects/RabbitMQ.Client/Impl/SocketFrameHandler.cs
lukebakken: Going back on this after a closer look — I think we should keep the drain. Tracing the paths:

- The inner per-frame `finally { frames.Dispose(); }` disposes only the frame *currently being processed* (the one assigned to `frames`). Frames still sitting in `_channelReader`'s buffer when `FlushAsync` throws aren't covered by it.
- `WriteAsyncCore`'s `catch` disposes only the frame whose `WriteAsync` call faulted for the producer — not frames already accepted into the channel.
- `CloseAsync` doesn't drain (`Complete()` + `await _writerTask`; on a normal close the write loop consumes the channel to empty, on a crash the `catch` path here is what drains).
- `OutgoingFrame` is a struct with no finalizer, so orphaned frames leak their pooled `byte[]` (and any `bodyOwner`) until the channel itself is GC'd.

`TryComplete(ex)` alone doesn't dispose buffered items — it only closes the writer side. So without this drain, on a write-loop crash any of the up-to-128 queued frames would leak their `ArrayPool` buffers until GC. Happy to chat if you see a path I missed.

projects/RabbitMQ.Client/Impl/SocketFrameHandler.cs
danielmarbach, line 240: This might be a pure optimization and not necessarily required: Should we consider whether the `_closed` flag should be set in the `WriteLoopAsync's` crash handler to give subsequent callers a faster exit path?

projects/RabbitMQ.Client/Impl/SocketFrameHandler.cs
```diff
@@ -281,14 +302,26 @@ await _pipeWriter.FlushAsync()
             catch (Exception ex)
```
danielmarbach, line 302: I would strongly prefer this version

```csharp
catch (Exception ex)
{
    ESLog.Error("Background socket write loop has crashed", ex);
    // Fail both already-blocked and subsequent WriteAsync callers
    // with ChannelClosedException(innerException: ex). Without
    // this, producers block indefinitely on a full channel whose
    // reader has died, and any holder of a publisher-confirm
    // semaphore (Channel.BasicPublish.cs) blocks the shutdown
    // handler in turn. See issue #1930.
    //
    // TryComplete rather than Complete: CloseAsync may have raced
    // us and already completed the channel.
    _channelWriter.TryComplete(ex);
    throw;
}
finally
{
    // Drain any leftover frames so their pooled buffers return to
    // the array pool rather than waiting for GC.
    while (_channelReader.TryRead(out OutgoingFrame leftover))
    {
        leftover.Dispose();
    }
}
```

This separates the two concerns clearly: fault signaling in catch, resource cleanup in finally. Each block does one job, and the finally guarantees cleanup regardless of how the method exits, which is undoubtedly what finally is for.

More details about my perspective

1. "Always drain on exit" is easier to reason about
The drain loop doesn't depend on why WriteLoopAsync is exiting,  it depends on whether there are leftover frames. The invariant is: pooled buffers must not leak, regardless of the exit path. Putting the drain in finally expresses that invariant directly: "whenever this loop exits, drain anything remaining." That's simpler to read and verify than "drain only when this specific exception fires."

2. It matches the original code structure
The current master already has drain-in-finally. This PR moves it into catch, which makes the diff look like the drain logic changed semantically when it really only changed locationally. Keeping it in finally reduces cognitive diff and the only truly new thing in the catch/finally region is `TryComplete(ex)`.

3. The normal-exit path already has an empty channel, so finally is a no-op there. When `CloseAsync` completes the writer gracefully, WaitToReadAsync returns false only after all buffered items have been consumed by the inner `TryRead` loop. So the finally drain hits an empty channel `TryRead` returns false immediately, the loop body never executes. No wasted work, no behavioral difference.

3. It's more resilient to future changes. If someone later adds another exit path (e.g., a new exception type, a cancellation path, or an `OperationCanceledException` from the `CancellationToken` passed to CreateAsync), having the drain in finally means it just works for any exit. With drain-in-catch, a new exit path would need to remember to drain.

danielmarbach (changes requested):

projects/RabbitMQ.Client/Impl/SocketFrameHandler.cs
danielmarbach: Happy to contribute in case you agree

lukebakken: Please feel free to modify this branch, thanks!

lukebakken: I just got it!

projects/RabbitMQ.Client/Impl/SocketFrameHandler.cs
lukebakken: Thanks for the suggestion. After tracing through the callers the genie and I think this one is better left out, for two reasons.

The `_closed` fast-exit path in `WriteAsync` silently disposes the frame and returns `default` - no exception. That's correct for a graceful close, but on a crash path it's wrong for both caller patterns. For fire-and-forget callers (`BasicAckAsync`, `BasicNackAsync`, `BasicRejectAsync`, etc.), the frame is silently dropped and the caller thinks it succeeded. For RPC-style callers (`QueueDeclare`, `ExchangeDeclare`, `BasicConsume`, `ChannelOpen`, and a dozen more), it's actively worse: `ModelSendAsync` completes successfully, the code proceeds to `await k`, the server never receives the frame so it never responds, and the caller hangs until `ContinuationTimeout` fires. That's a multi-second delay before learning anything went wrong, instead of an immediate `ChannelClosedException`. The existing `TryComplete(ex)` path already gives every caller an immediate failure with the right exception.

There's also a thread-safety issue: `_closed` is a plain `bool` with no `volatile` or `Interlocked` fence. Today it's only written in `CloseAsync`'s `finally` and read by callers of `WriteAsync` - a usage pattern that avoids the problem in practice. Writing it from the writer-task thread in `WriteLoopAsync` introduces a genuine data race against any concurrent `WriteAsync` reader.

The performance argument is less compelling than it looks. `TryComplete(ex)` already makes the channel immediately done, so the next `WriteAsync` -> `WriteAsyncCore` -> `_channelWriter.WriteAsync` call gets a synchronously-faulted `ValueTask` back. It goes through `AwaitAndDisposeOnException` but never suspends. The extra overhead is one virtual dispatch and a no-suspend async state machine - negligible on a crash path.

So: seems like it's not needed.

*(Written by Claude with @lukebakken's review and approval.)*

danielmarbach (approved):

projects/RabbitMQ.Client/Impl/SocketFrameHandler.cs
danielmarbach: Nicely analyzed. Agree with the conclusion: setting `_closed` in the crash handler is a bad idea, but for one reason only: the `_closed` fast path returns default (success), which silently drops the frame. That's the wrong outcome on a crash path where the caller should get an exception.

The data race argument is a bit of a stretch, though. There's already a data race on `_closed` today between CloseAsync's write and `WriteAsync's` read (no volatile, no `Volatile.Read/Write`). If the semantics were correct, adding volatile would be a one-liner fix. The real reason not to do this is that winning the race produces the wrong behavior, which is just the return-semantics argument again in disguise.

PauloHMattos (approved):
LGTM!

About the test, I put copilot to work on it, but is not passing and I haven't had time to look at it. I think is not worth to wait for the test to merge this change.
Once I have time I will adjust the test code and target the PR to the main branch.

PR #1934, opened 2026-05-01
Add regression tests for #1930

PauloHMattos:
## Proposed Changes

Regression tests for #1930, plus the fixes needed to make them pass reliably.

**Issue summary:** when the write loop crashes (socket error or flush timeout) without calling `_channelWriter.TryComplete(ex)`, publishers blocked on a full `Channel<OutgoingFrame>(128)` stall indefinitely. With publisher confirms enabled the confirm semaphore is never released, which deadlocks the connection shutdown handler. With confirms disabled publishers block directly on `WriteAsync` with no escape path. Both scenarios were reported in #1930.

**What this PR adds:**

`SocketFlushTimeout` (internal) on `ConnectionFactory` - sets a cancellation deadline on `PipeWriter.FlushAsync` inside the write loop. When the deadline fires the write loop crashes via `OperationCanceledException`. This is a test knob, not a public API; the default (zero) disables it. It is used by the Toxiproxy tests to crash the write loop deterministically on all platforms without relying on TCP RST propagation, which is unreliable through Windows port-forwarding.

`ToxiproxyManager` changes - added a `ProxyHost` property and support for two new environment variables: `RABBITMQ_TOXIPROXY_HOST` (overrides both the management API host and the address clients connect to, useful when running tests from WSL2 using the Docker bridge IP) and `RABBITMQ_TOXIPROXY_UPSTREAM_HOST` (overrides the upstream RabbitMQ address seen by Toxiproxy, needed when both services run in Docker and `127.0.0.1` would resolve to the Toxiproxy container's own loopback). Connection construction was moved after all environment variable processing so the resolved host is used consistently.

Two regression tests using `BandwidthToxic(Rate=1, UpStream)` to throttle the write path to 1 KB/s - this fills the OS TCP send buffer, stalls `FlushAsync`, and fills the 128-slot outgoing channel, creating the exact pre-condition from the issue without killing the TCP connection:

- `TestPublishDoesNotHangAfterConnectionLoss_NoConfirms_GH1930` - scenario A from the issue: publishers block directly on `WriteAsync` when the channel is full and the write loop has stalled.
- `TestPublishDoesNotHangAfterConnectionLoss_WithConfirms_GH1930` - scenario B: one publisher holds `_confirmSemaphore(1,1)` while blocked on `WriteAsync`; without the fix the shutdown handler deadlocks waiting for the same semaphore.

**Fixes included in this PR (review feedback):**

- `Connection.CloseAsync`: moved `SetSessionClosingAsync` inside the try block and added `catch (OperationCanceledException)` alongside the existing `catch (ChannelClosedException)`. When the local 5-second abort timeout fires before `TransmitAsync` completes, `_channelWriter.WriteAsync` checks the cancelled token first and throws `OperationCanceledException` rather than `ChannelClosedException`. The missing catch caused the exception to escape through `DisposeAsync` to the test runner. See commit bbcd19a4c.
- `SocketFlushTimeout` made internal and removed from `IFrameHandler`. See commit cfc8d3f98.
- `FlushPipeAsync`: on NET, reuse a single `CancellationTokenSource` via `TryReset()` + `CancelAfter()` instead of allocating a new one per flush. See commit f0287a219.
- `blockedCount` guard lowered from `>= PublishCount/2` to `> 128`. The channel holds exactly 128 frames so any count above 128 proves the channel is full and the write loop has stopped draining, independent of OS TCP send-buffer size. See commit 55ca36b8f.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #1930)

## Checklist

### DIFF
--- a/projects/RabbitMQ.Client/ConnectionFactory.cs
+++ b/projects/RabbitMQ.Client/ConnectionFactory.cs
@@ -243,6 +243,8 @@ namespace RabbitMQ.Client
         /// </summary>
         public TimeSpan SocketWriteTimeout { get; set; } = DefaultConnectionTimeout;
 
+        internal TimeSpan SocketFlushTimeout { get; set; } = TimeSpan.Zero;
+
         /// <summary>
         /// TLS options setting.
         /// </summary>
@@ -625,6 +627,11 @@ namespace RabbitMQ.Client
             if (SocketWriteTimeout > RequestedHeartbeat)
             {
                 fh.WriteTimeout = SocketWriteTimeout;
+            }
+
+            if (SocketFlushTimeout != default && fh is SocketFrameHandler sfh)
+            {
+                sfh.FlushTimeout = SocketFlushTimeout;
             }
 
             return fh;
--- a/projects/RabbitMQ.Client/Impl/Connection.cs
+++ b/projects/RabbitMQ.Client/Impl/Connection.cs
@@ -353,17 +353,24 @@ namespace RabbitMQ.Client.Impl
                 await OnShutdownAsync(reason)
                     .ConfigureAwait(false);
 
-                await _session0.SetSessionClosingAsync(false, cts.Token)
-                    .ConfigureAwait(false);
-
                 try
                 {
+                    await _session0.SetSessionClosingAsync(false, cts.Token)
+                        .ConfigureAwait(false);
+
                     // Try to send connection.close wait for CloseOk in the MainLoop
                     if (false == _closed)
                     {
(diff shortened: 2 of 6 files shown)


PauloHMattos:
I'm having trouble running the toxiproxy tests, so I'm opening to main first where hopefully the new tests will fail. Then I will switch to the fix branch and hopefully they will pass.

lukebakken:
I'm reviewing this PR now! Some changes will be on the way.

lukebakken:
@PauloHMattos no rush, but let me know what you think. Thanks.

lukebakken (approved):

PR #1939, opened 2026-05-13
Change HandledProtocolCommandIds to use ReadOnlySpan for improved performance

danielmarbach:
## Proposed Changes

Replace `Queue<ProtocolCommandId[]>` with a lock-free `readonly struct` reinterpreted as `long` via `Unsafe.As` for atomic `Interlocked.Exchange`, eliminating heap allocations and fixing a thread-safety bug. Also change `HandledProtocolCommandIds` from `ProtocolCommandId[]` to `ReadOnlySpan<ProtocolCommandId>` to support zero-copy propagation.

### Thread-safety fix

The previous `Queue<ProtocolCommandId[]>` implementation had no thread-safety guarantees — `Queue<T>` is not thread-safe, and the check-then-act pattern (`Count > 0` then `Dequeue()`) was a TOCTOU race. The initial struct replacement also had an ordering vulnerability: `_lastTimedOutCommandIdsCount` was written before the command ID fields, so `ShouldIgnoreCommand` could observe a non-zero count with still-zeroed IDs.

The new implementation uses a single `long` field with `Interlocked.Exchange` for both the write (in `RpcCanceled`) and the read-and-reset (in `ShouldIgnoreCommand`). This provides:
- **Atomicity on all platforms** (including 32-bit) — readers see either the old state (0 = empty) or the complete new state, never a torn partial write
- **Full memory barrier** — proper visibility across threads without needing `volatile` or `lock`

### RpcContinuationQueue

- Replace `Queue<ProtocolCommandId[]>` with a `readonly struct LastTimedOutCommandIds(first, second = 0)` reinterpreted as `long` via `Unsafe.As` for lock-free atomic read-modify-write via `Interlocked.Exchange`
- Since `ProtocolCommandId : uint` has no zero-valued member, `default` (0) serves as the "empty" sentinel, eliminating the need for a separate count field
- `RpcCanceled` now accepts `ReadOnlySpan<ProtocolCommandId>` and constructs the struct, then writes it atomically — zero heap allocation
- `ShouldIgnoreCommand` uses `Interlocked.Exchange` to atomically read-and-reset, then compares directly against the struct fields — no LINQ, no Collection
- Early return on `raw == 0L` avoids any comparison work on the hot path (no timeout pending)
- `Debug.Assert` enforces `protocolCommandIds.Length is > 0 and <= 2`
- Removes `System.Collections.Generic`, `System.Linq` dependencies; adds `System.Runtime.CompilerServices` for `Unsafe.As`

### AsyncRpcContinuations

- Change `HandledProtocolCommandIds` from `ProtocolCommandId[]` to `ReadOnlySpan<ProtocolCommandId>`

### Tests

- Added 8 unit tests for `ShouldIgnoreCommand` and `RpcCanceled`: empty state, single/two command matching, non-matching, state consumption, and `responseReceived=true` guard

### Benchmark results

image

Although one could argue the comparison is "unfair" since this implementation is now properly handling the race. But still very good gains

### DIFF
--- a/projects/RabbitMQ.Client/Impl/AsyncRpcContinuations.cs
+++ b/projects/RabbitMQ.Client/Impl/AsyncRpcContinuations.cs
@@ -98,7 +98,7 @@ namespace RabbitMQ.Client.Impl
             return _tcsConfiguredTaskAwaitable.GetAwaiter();
         }
 
-        public abstract ProtocolCommandId[] HandledProtocolCommandIds { get; }
+        public abstract ReadOnlySpan<ProtocolCommandId> HandledProtocolCommandIds { get; }
 
         public async Task HandleCommandAsync(IncomingCommand cmd)
         {
@@ -208,7 +208,7 @@ namespace RabbitMQ.Client.Impl
         {
         }
 
-        public override ProtocolCommandId[] HandledProtocolCommandIds
+        public override ReadOnlySpan<ProtocolCommandId> HandledProtocolCommandIds
             => [ProtocolCommandId.ConnectionSecure, ProtocolCommandId.ConnectionTune];
 
         protected override Task DoHandleCommandAsync(IncomingCommand cmd)
@@ -248,8 +248,8 @@ namespace RabbitMQ.Client.Impl
             _expectedCommandId = expectedCommandId;
         }
 
-        public override ProtocolCommandId[] HandledProtocolCommandIds
-            => [_expectedCommandId];
+        public override ReadOnlySpan<ProtocolCommandId> HandledProtocolCommandIds
+            => new[] { _expectedCommandId };
 
         protected override Task DoHandleCommandAsync(IncomingCommand cmd)
         {
@@ -318,7 +318,7 @@ namespace RabbitMQ.Client.Impl
             _consumerDispatcher = consumerDispatcher;
         }
 
-        public override ProtocolCommandId[] HandledProtocolCommandIds
+        public override ReadOnlySpan<ProtocolCommandId> HandledProtocolCommandIds
             => [ProtocolCommandId.BasicConsumeOk];
 
         protected override async Task DoHandleCommandAsync(IncomingCommand cmd)
@@ -350,7 +350,7 @@ namespace RabbitMQ.Client.Impl
             _adjustDeliveryTag = adjustDeliveryTag;
         }
 
-        public override ProtocolCommandId[] HandledProtocolCommandIds
+        public override ReadOnlySpan<ProtocolCommandId> HandledProtocolCommandIds
             => [ProtocolCommandId.BasicGetOk, ProtocolCommandId.BasicGetEmpty];
 
         internal DateTime StartTime { get; } = DateTime.UtcNow;
@@ -468,7 +468,7 @@ namespace RabbitMQ.Client.Impl
         {
         }
 
-        public override ProtocolCommandId[] HandledProtocolCommandIds
+        public override ReadOnlySpan<ProtocolCommandId> HandledProtocolCommandIds
             => [ProtocolCommandId.QueueDeclareOk];
 
         protected override Task DoHandleCommandAsync(IncomingCommand cmd)
@@ -511,7 +511,7 @@ namespace RabbitMQ.Client.Impl
         {
         }
 
-        public override ProtocolCommandId[] HandledProtocolCommandIds
+        public override ReadOnlySpan<ProtocolCommandId> HandledProtocolCommandIds
             => [ProtocolCommandId.QueueDeleteOk];
 
         protected override Task DoHandleCommandAsync(IncomingCommand cmd)
@@ -537,7 +537,7 @@ namespace RabbitMQ.Client.Impl
         {
         }
 
-        public override ProtocolCommandId[] HandledProtocolCommandIds
+        public override ReadOnlySpan<ProtocolCommandId> HandledProtocolCommandIds
             => [ProtocolCommandId.QueuePurgeOk];
 
         protected override Task DoHandleCommandAsync(IncomingCommand cmd)
--- a/projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
+++ b/projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
@@ -32,6 +32,7 @@
 using System;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using RabbitMQ.Client.Events;
@@ -68,15 +69,19 @@ namespace RabbitMQ.Client.Impl
 
         private const int CommandIdBufferLength = 2;
 
-        private struct LastTimedOutCommandIds
+        // Two ProtocolCommandId values (each uint) laid out as a struct
+        // reinterpreted as a single long for lock-free atomic
+        // read-modify-write via Interlocked.Exchange.
+        // Since ProtocolCommandId: uint has no zero-valued member,
+        // default(LastTimedOutCommandIds) == 0 means "no timed-out command".
+        private readonly struct LastTimedOutCommandIds(ProtocolCommandId first, ProtocolCommandId second = 0)
         {
-            public ProtocolCommandId First;
-            public ProtocolCommandId Second;
+            public readonly ProtocolCommandId First = first;
+            public readonly ProtocolCommandId Second = second;
         }
 
         private static readonly EmptyRpcContinuation s_tmp = new EmptyRpcContinuation();
-        private LastTimedOutCommandIds _lastTimedOutCommandIds;
-        private int _lastTimedOutCommandIdsCount;
+        private long _lastTimedOutCommandIds;
         private IRpcContinuation _outstandingRpc = s_tmp;
 
         ///<summary>Enqueue a continuation, marking a pending RPC.</summary>
@@ -160,19 +165,11 @@ namespace RabbitMQ.Client.Impl
 
             // AMQP 0-9-1 RPCs handle at most 2 response command IDs
             // (e.g. BasicGetOk/BasicGetEmpty, ConnectionSecure/ConnectionTune)
-            Debug.Assert(protocolCommandIds.Length <= CommandIdBufferLength);
+            Debug.Assert(protocolCommandIds.Length is > 0 and <= CommandIdBufferLength);
 
-            int count = protocolCommandIds.Length;
-            _lastTimedOutCommandIdsCount = count;
-            _lastTimedOutCommandIds = default;
-            if (count > 0)
-            {
-                _lastTimedOutCommandIds.First = protocolCommandIds[0];
-                if (count > 1)
-                {
-                    _lastTimedOutCommandIds.Second = protocolCommandIds[1];
-                }
-            }
+            var ids = new LastTimedOutCommandIds(first: protocolCommandIds[0], second: protocolCommandIds.Length > 1 ? protocolCommandIds[1] : 0);
+
+            Interlocked.Exchange(ref _lastTimedOutCommandIds, Unsafe.As<LastTimedOutCommandIds, long>(ref ids));
         }
 
         public bool ShouldIgnoreCommand(ProtocolCommandId commandId)
@@ -180,12 +177,15 @@ namespace RabbitMQ.Client.Impl
             // rabbitmq/rabbitmq-dotnet-client#1802
             // This keeps track of ProtocolCommandId values from previous RPC
             // commands that have timed out.
-            LastTimedOutCommandIds last = _lastTimedOutCommandIds;
-            int count = _lastTimedOutCommandIdsCount;
-            _lastTimedOutCommandIds = default;
-            _lastTimedOutCommandIdsCount = 0;
+            long raw = Interlocked.Exchange(ref _lastTimedOutCommandIds, 0L);
 
-            return commandId == last.First || (count > 1 && commandId == last.Second);
+            if (raw == 0L)
+            {
+                return false;
+            }
+
+            var ids = Unsafe.As<long, LastTimedOutCommandIds>(ref raw);
+            return commandId == ids.First || (ids.Second != 0 && commandId == ids.Second);
         }
     }
 }
--- a/projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
+++ b/projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
@@ -30,9 +30,7 @@
 //---------------------------------------------------------------------------
 
 using System;
-using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
-using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using RabbitMQ.Client.Events;
@@ -67,8 +65,17 @@ namespace RabbitMQ.Client.Impl
             }
         }
 
+        private const int CommandIdBufferLength = 2;
+
+        private struct LastTimedOutCommandIds
+        {
+            public ProtocolCommandId First;
+            public ProtocolCommandId Second;
+        }
+
         private static readonly EmptyRpcContinuation s_tmp = new EmptyRpcContinuation();
-        private readonly Queue<ProtocolCommandId[]> _rpcCancellationQueue = new();
+        private LastTimedOutCommandIds _lastTimedOutCommandIds;
+        private int _lastTimedOutCommandIdsCount;
         private IRpcContinuation _outstandingRpc = s_tmp;
 
         ///<summary>Enqueue a continuation, marking a pending RPC.</summary>
@@ -143,11 +150,28 @@ namespace RabbitMQ.Client.Impl
             return false;
         }
 
-        public void RpcCanceled(bool responseReceived, ProtocolCommandId[] protocolCommandIds)
+        public void RpcCanceled(bool responseReceived, ReadOnlySpan<ProtocolCommandId> protocolCommandIds)
         {
-            if (!responseReceived)
+            if (responseReceived)
             {
-                _rpcCancellationQueue.Enqueue(protocolCommandIds);
+                return;
+            }
+
+            if (protocolCommandIds.Length > CommandIdBufferLength)
+            {
+                throw new ArgumentOutOfRangeException(nameof(protocolCommandIds));
+            }
+
+            int count = protocolCommandIds.Length;
+            _lastTimedOutCommandIdsCount = count;
+            _lastTimedOutCommandIds = default;
+            if (count > 0)
+            {
+                _lastTimedOutCommandIds.First = protocolCommandIds[0];
+                if (count > 1)
+                {
+                    _lastTimedOutCommandIds.Second = protocolCommandIds[1];
+                }
             }
         }
 
@@ -156,18 +180,27 @@ namespace RabbitMQ.Client.Impl
             // rabbitmq/rabbitmq-dotnet-client#1802
             // This keeps track of ProtocolCommandId values from previous RPC
             // commands that have timed out.
-            bool rv = false;
+            LastTimedOutCommandIds last = _lastTimedOutCommandIds;
+            int count = _lastTimedOutCommandIdsCount;
+            _lastTimedOutCommandIds = default;
+            _lastTimedOutCommandIdsCount = 0;
 
-            if (_rpcCancellationQueue.Count > 0)
+            if (count == 0)
             {
-                ProtocolCommandId[] lastErroredCommandIds = _rpcCancellationQueue.Dequeue();
-                if (lastErroredCommandIds.Contains(commandId))
-                {
-                    rv = true;
-                }
+                return false;
             }
 
-            return rv;
+            if (commandId == last.First)
+            {
+                return true;
+            }
+
+            if (count > 1 && commandId == last.Second)
+            {
+                return true;
+            }
+
+            return false;
         }
     }
 }
--- a/projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
+++ b/projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
@@ -32,7 +32,7 @@
 using System;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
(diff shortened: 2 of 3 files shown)


danielmarbach:
@paulomorgado if you have some time, I'd appreciate your review

projects/RabbitMQ.Client/Impl/AsyncRpcContinuations.cs
```diff
@@ -244,8 +244,8 @@ public SimpleAsyncRpcContinuation(ProtocolCommandId expectedCommandId, TimeSpan
             _expectedCommandId = expectedCommandId;
         }
 
-        public override ProtocolCommandId[] HandledProtocolCommandIds
-            => [_expectedCommandId];
+        public override ReadOnlySpan<ProtocolCommandId> HandledProtocolCommandIds
+            => new[] { _expectedCommandId };
```
danielmarbach, line 252: Because `HandledProtocolCommandIds` is only accessed when calling `ShouldIgnoreCommand` here it is cheaper to array an actual array when the data is needed.

projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
```diff
@@ -67,8 +66,40 @@ public void Dispose()
...
+#if NETSTANDARD
+        private unsafe struct CommandIdBuffer
+        {
+            private fixed uint _element[CommandIdBufferLength];
+
+            public ReadOnlySpan<ProtocolCommandId> AsReadOnlySpan()
+            {
+                fixed (uint* ptr = _element)
+                {
+                    return new ReadOnlySpan<ProtocolCommandId>((ProtocolCommandId*)ptr, CommandIdBufferLength);
```
gfoidl: This is not safe.
The ROS is created around a unmanaged pointer, which can be moved by GC when the `fixed`-scope is left, i.e. by the `return`. Thus the caller has potential a span with a dangling pointer.

There is no safe way to get a span out of a fixed sized buffer[ ^1], w/o pinning it (fixed statement).

[^1]: w/o lots of ceremony and w/o unsafe hacks (that are not reliable accross runtimines).

And unfortunately the `[InlineArray]` attribute can't be polyfilled for .NET Standard neither, as it needs runtime support.

danielmarbach: Yeah I was chewing on this since yesterday and concluded ultimately we only ever need a struct with two fields and that's it.

danielmarbach: The inline array variant is very elegant and that's where I started with until I hit the polyfill problem you mentioned. I completely forgot it requires runtime support

danielmarbach: Ultimately my primary interest was not to squeeze perf but make thee current design more intention revealing since there is no need for a queue in there given it only ever has one item in it

danielmarbach: @gfoidl I will address this later today but I think I have a simpler approach that also reduces the if/defs

danielmarbach: @gfoidl I think the new approach is much better

projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
```diff
@@ -143,11 +150,28 @@ public bool TryPeek<T>([NotNullWhen(true)] out T? continuation) where T : class,
...
-        public void RpcCanceled(bool responseReceived, ProtocolCommandId[] protocolCommandIds)
+        public void RpcCanceled(bool responseReceived, ReadOnlySpan<ProtocolCommandId> protocolCommandIds)
         {
-            if (!responseReceived)
+            if (responseReceived)
+            {
+                return;
+            }
+
+            if (protocolCommandIds.Length > CommandIdBufferLength)
```
gfoidl: I'd like to see a comment why max. 2 commands can be here?
But: I don't know too much about the implementation, so when this is obvious to someone familiar with the code, so ignore this comment.

---

When `2` is the correct value, is a `Debug.Assert(protocolCommandIds.Length <= 2);` enough instead of the exception?

projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
```diff
@@ -156,18 +180,27 @@ public bool ShouldIgnoreCommand(ProtocolCommandId commandId)
...
+                return true;
             }
 
-            return rv;
+            if (count > 1 && commandId == last.Second)
+            {
+                return true;
+            }
+
+            return false;
```
gfoidl: More or less a matter of taste:
```suggestion
            return commandId == last.First || (count > 1 && commandId == last.Second);
```

gfoidl (commented):
@danielmarbach besides my question, this approach is much better 👍🏻

projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
danielmarbach: Unfortunately, GitHub moaned I cannot pull in this suggestion because of the deleted lines :(

danielmarbach: Pushed

projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
danielmarbach: Done

gfoidl (approved):
Nice 👍🏻

projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
```diff
@@ -143,11 +151,27 @@ public bool TryPeek<T>([NotNullWhen(true)] out T? continuation) where T : class,
...
-            if (!responseReceived)
+            if (responseReceived)
             {
-                _rpcCancellationQueue.Enqueue(protocolCommandIds);
+                return;
+            }
+
+            // AMQP 0-9-1 RPCs handle at most 2 response command IDs
+            // (e.g. BasicGetOk/BasicGetEmpty, ConnectionSecure/ConnectionTune)
+            Debug.Assert(protocolCommandIds.Length <= CommandIdBufferLength);
```
paulomorgado: Is this enough? Should it throw an `ArgumentOutOfRangeException`?

projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
```diff
@@ -156,18 +180,12 @@ public bool ShouldIgnoreCommand(ProtocolCommandId commandId)
...
-                }
-            }
+            LastTimedOutCommandIds last = _lastTimedOutCommandIds;
+            int count = _lastTimedOutCommandIdsCount;
+            _lastTimedOutCommandIds = default;
+            _lastTimedOutCommandIdsCount = 0;
 
-            return rv;
+            return commandId == last.First || (count > 1 && commandId == last.Second);
         }
```
paulomorgado: Could this be a race condition?

`RpcCanceled` sets `_lastTimedOutCommandIdsCount` before writing `_lastTimedOutCommandIds.First`/`Second`. `ShouldIgnoreCommand` can run concurrently on the receive loop, observe `count > 0` before the IDs are populated,
clear the state, and fail to ignore the late response. That can route a timed-out RPC response into the normal continuation path and produce unexpected-command/protocol failures.

projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
gfoidl: See https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/1939#discussion_r3241339121

projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
gfoidl: <sub>Thanks for the background info, so it's easier for me to learn what's going on</sub>

Yep, could be.

https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/a62f2007e99e8b2c734c03d180b2aeb79727c5e9/projects/RabbitMQ.Client/Framing/ProtocolCommandId.cs#L34 can't have a `0` entry, so one could get rid of `_lastTimedOutCommandIdsCount` and just test against the entries (first, second) of `_lastTimedOutCommandIds` (or have a property in the struct that exposes a count based on these checks).

And to avoid torn reads/writes of `_lastTimedOutCommandIds` (two `uint`s) `Interlocked` could be used. But unfortunately only on .NET, not on .NET Standard (there a classic `lock` could be used).
Or use a `class` instead of the `struct`, because references are always atomic (at the cost of the allocation).

danielmarbach: > can't have a `0` entry, so one could get rid of `_lastTimedOutCommandIdsCount` and just test against the entries (first, second) of `_lastTimedOutCommandIds` (or have a property in the struct that exposes a count based on these checks).

I considered this too but figured it might be brittle if we ever introduce 0 there, so I wanted to avoid it, but it is definitely an elegant solution

danielmarbach: The torn reads is already a problem on master and I'm pretty sure the problem was made worse in the last PR introducing the count access. Queue has no safety guarantees in that regard. I'll look into this a bit more later but this is not a "new problem" and in some ways this code is quite likely slightly better than what is already on master but I have to think about this more

paulomorgado: How about something like this?


```suggestion
     ulong packed = 0;
     if (protocolCommandIds.Length > 0)
     {
         packed = (uint)protocolCommandIds[0];
 
         if (protocolCommandIds.Length > 1)
         {
             packed |= (ulong)(uint)protocolCommandIds[1] << 32;
         }
     }
 
     Interlocked.Exchange(ref _lastTimedOutCommandIds, unchecked((long)packed));
 }
 
 public bool ShouldIgnoreCommand(ProtocolCommandId commandId)
 {
     ulong packed = unchecked((ulong)Interlocked.Exchange(ref _lastTimedOutCommandIds, 0));
 
     uint first = (uint)packed;
     uint second = (uint)(packed >> 32);
     uint actual = (uint)commandId;
 
     return actual == first || actual == second;
 }
```

Where:

```csharp
 private const int CommandIdBufferLength = 2;
 private long _lastTimedOutCommandIds;
```

danielmarbach: I was thinking about doing the same trick we do with DecimalData but I need to double check whether that is safe

danielmarbach: Pushed

projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
```diff
@@ -67,8 +67,21 @@ public void Dispose()
...
+        // Two ProtocolCommandId values (each uint) laid out as a struct
+        // reinterpreted as a single long for lock-free atomic
+        // read-modify-write via Interlocked.Exchange.
+        // Since ProtocolCommandId: uint has no zero-valued member,
+        // default(LastTimedOutCommandIds) == 0 means "no timed-out command".
+        private readonly struct LastTimedOutCommandIds(ProtocolCommandId first, ProtocolCommandId second = 0)
+        {
+            public readonly ProtocolCommandId First = first;
+            public readonly ProtocolCommandId Second = second;
+        }
```
gfoidl: You could make this an union (in the classic sense, not the one coming to next C# version):
```suggestion
        [StructLayout(LayoutKind.Explicit)]
        private readonly struct LastTimedOutCommandIds(ProtocolCommandId first, ProtocolCommandId second = 0)
        {
            [FieldOffset(0)]
            public readonly ProtocolCommandId First = first;

            [FieldOffset(sizeof(ProtocolCommandId))]
            public readonly ProtocolCommandId Second = second;
            
            [FieldOffset(0)]
            public readonly long RawValue;
        }
```
To get rid of the `Unsafe.As` or the bit-manipulation.
And the comment above matches more, as the `long` is "overlayed" the two other fields.

gfoidl (commented):
Use a union instead?

If the current variant with `Unsafe.As` re-interpret cast is kept, I'd like to see these casts moved to the struct (as static factory method and property to the get the raw value (long)).

lukebakken (commented):
Here's what the 🧞‍♂️ and I have for our review:

## `Unsafe.As` and struct layout

`LastTimedOutCommandIds` had no `[StructLayout]` attribute, so the CLR spec leaves field ordering unspecified (auto layout). `Unsafe.As<LastTimedOutCommandIds, long>` was silently relying on `First` being at byte 0 and `Second` at byte 4 - true in practice for two same-type fields, but not guaranteed.

Following @gfoidl, the 🧞‍♂️ this with a `[StructLayout(LayoutKind.Explicit)]` union (gfoidl's suggestion), which makes the layout part of the type contract and removes `Unsafe.As` entirely:

```csharp
[StructLayout(LayoutKind.Explicit)]
private struct LastTimedOutCommandIds
{
    [FieldOffset(0)]
    public ProtocolCommandId First;
    [FieldOffset(4)]
    public ProtocolCommandId Second;
    [FieldOffset(0)]
    public long RawValue;
}
```

Packing: `ids.RawValue` after setting `First`/`Second`. Unpacking: `new LastTimedOutCommandIds { RawValue = raw }` then reading `First`/`Second`. No `Unsafe.As` in either direction. Verified with `dotnet-script` that `sizeof(uint) == 4` and the round-trip is correct. See 2972f33b1.

The fields are non-readonly deliberately. Making them `readonly` would require `Unsafe.SkipInit(out this)` in a constructor to satisfy definite-assignment rules for the overlapping `RawValue` field. Since this struct is private and only ever used as a stack-local for packing/unpacking, there is no defensive-copy risk and the mutable form is simpler.

## `SimpleAsyncRpcContinuation.HandledProtocolCommandIds` allocates

```csharp
public override ReadOnlySpan<ProtocolCommandId> HandledProtocolCommandIds
    => new[] { _expectedCommandId };
```

Every concrete subtype uses a collection expression over a compile-time constant, which the compiler lowers to a reference into the read-only data section - zero allocation. This one allocates a `ProtocolCommandId[]` on every call because `_expectedCommandId` is an instance field. Since `HandledProtocolCommandIds` is only read from `RpcCanceled`, which only fires on the timeout path, the allocation is acceptable. But the "zero heap allocation" claim in the PR description isn't correct.

## `ShouldIgnoreCommand` consumes state on any non-zero call

The old `Queue`-based implementation only dequeued (consumed) the timed-out IDs when `commandId` matched. The new implementation always consumes via `Interlocked.Exchange(..., 0L)` whenever `raw != 0`, regardless of whether the command matches. `TestShouldIgnoreCommand_EmptyAfterConsume` covers this explicitly. It is benign in practice - AMQP 0-9-1 strict ordering means the late response is always the very next incoming command - but a short comment in `ShouldIgnoreCommand` would make the intent clear.

gfoidl:
Unrelated to this PR, but

> `LastTimedOutCommandIds` had no `[StructLayout]` attribute, so the CLR spec leaves field ordering unspecified (auto layout).

is correct, though C# set layoutkind sequential. And according the packing rules it was guaranteed that the layout of the struct is as intended.

danielmarbach:
@lukebakken FYI I think the explicit type layout is also not used in the decimal conversion. I followed that pattern. I like the change!

danielmarbach:
> Every concrete subtype uses a collection expression over a compile-time constant, which the compiler lowers to a reference into the read-only data section - zero allocation. This one allocates a ProtocolCommandId[] on every call because _expectedCommandId is an instance field. Since HandledProtocolCommandIds is only read from RpcCanceled, which only fires on the timeout path, the allocation is acceptable. But the "zero heap allocation" claim in the PR description isn't correct.

That's a great genie remark, especially [because I mentioned this already](https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/1939#discussion_r3238303480) ;)

danielmarbach:
> The old `Queue`-based implementation only dequeued (consumed) the timed-out IDs when `commandId` matched.

Hu?

```
                ProtocolCommandId[] lastErroredCommandIds = _rpcCancellationQueue.Dequeue();
                if (lastErroredCommandIds.Contains(commandId))
                {
                    rv = true;
                }
```

> It is benign in practice - AMQP 0-9-1 strict ordering means the late response is always the very next incoming command - but a short comment in ShouldIgnoreCommand would make the intent clear.

I think this is a remark that applies to this AND the master version in general

lukebakken:
Thanks @danielmarbach and @gfoidl ... this is all low-level stuff I'm not very familiar with with regard to C#/.NET

paulomorgado:
> Following @gfoidl, the 🧞‍♂️ this with a `[StructLayout(LayoutKind.Explicit)]` union (gfoidl's suggestion), which makes the layout part of the type contract and removes `Unsafe.As` entirely:
> 
> ```cs
> [StructLayout(LayoutKind.Explicit)]
> private struct LastTimedOutCommandIds
> {
>     [FieldOffset(0)]
>     public ProtocolCommandId First;
>     [FieldOffset(4)]
>     public ProtocolCommandId Second;
>     [FieldOffset(0)]
>     public long RawValue;
> }
> ```
> 
> Packing: `ids.RawValue` after setting `First`/`Second`. Unpacking: `new LastTimedOutCommandIds { RawValue = raw }` then reading `First`/`Second`. No `Unsafe.As` in either direction. Verified with `dotnet-script` that `sizeof(uint) == 4` and the round-trip is correct. See [2972f33](https://github.com/rabbitmq/rabbitmq-dotnet-client/commit/2972f33b1693bf7385a0fbd1f66eacd54fb6b721).

This hack is interesting, but it should be benchmarked to see if it's really an optimization for the C# compiler.

But if you are using this, consider:

```csharp
[StructLayout(LayoutKind.Explicit)]
private struct LastTimedOutCommandIds
{
    [FieldOffset(0)]
    public ProtocolCommandId First;
    [FieldOffset(sizeof(int))]
    public ProtocolCommandId Second;
    [FieldOffset(0)]
    public long RawValue;
}
```

danielmarbach:
> > Following @gfoidl, the 🧞‍♂️ this with a `[StructLayout(LayoutKind.Explicit)]` union (gfoidl's suggestion), which makes the layout part of the type contract and removes `Unsafe.As` entirely:
> > ```cs
> > [StructLayout(LayoutKind.Explicit)]
> > private struct LastTimedOutCommandIds
> > {
> >     [FieldOffset(0)]
> >     public ProtocolCommandId First;
> >     [FieldOffset(4)]
> >     public ProtocolCommandId Second;
> >     [FieldOffset(0)]
> >     public long RawValue;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Packing: `ids.RawValue` after setting `First`/`Second`. Unpacking: `new LastTimedOutCommandIds { RawValue = raw }` then reading `First`/`Second`. No `Unsafe.As` in either direction. Verified with `dotnet-script` that `sizeof(uint) == 4` and the round-trip is correct. See [2972f33](https://github.com/rabbitmq/rabbitmq-dotnet-client/commit/2972f33b1693bf7385a0fbd1f66eacd54fb6b721).
> 
> This hack is interesting, but it should be benchmarked to see if it's really an optimization for the C# compiler.
> 
> But if you are using this, consider:
> 
> ```cs
> [StructLayout(LayoutKind.Explicit)]
> private struct LastTimedOutCommandIds
> {
>     [FieldOffset(0)]
>     public ProtocolCommandId First;
>     [FieldOffset(sizeof(int))]
>     public ProtocolCommandId Second;
>     [FieldOffset(0)]
>     public long RawValue;
> }
> ```

FieldOffset(sizeof(ProtocolCommandId))

Is IMHO better as it is now

gfoidl:
> should be benchmarked to see if it's really an optimization

What benchmark do you want to see? The JIT compiles this down to just a memmory access.
From a C#'s perspective either the two fields of the struct are accessed, or the one overlayed field.

I'd use
```diff
-[FieldOffset(sizeof(int))]
+[FieldOffset(sizeof(ProtocolCommandId))]
```
to make it obvious what offset is meant (as suggested in https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/1939#discussion_r3244118189).

danielmarbach:
> benchmarked to see if it's really an optimization

I think this comment is also not taking into account that asking for every little change to be benchmarked of something that is already proven to be faster than before puts a very high bar of entry for community contributions. I know @paulomorgado well enough that it was well meant so I'm not picking on him. My intent is to highlight those things come at a cost as well

paulomorgado:
> > benchmarked to see if it's really an optimization
> 
> I think this comment is also not taking into account that asking for every little change to be benchmarked of something that is already proven to be faster than before puts a very high bar of entry for community contributions. I know @paulomorgado well enough that it was well meant so I'm not picking on him. My intent is to highlight those things come at a cost as well

I admit this was a quick very localized review.

The C# compiler already normally emits sequential layout, not auto layout, so it is not generally free to reorder `DecimalData` fields. But I also like to be explicit.

My performance inquiry (again, without deeper review) is that structs can byte you with unnoticed copies that can introduce performance and reliability issues. I haven't noticed that, but thought I would to ask.

danielmarbach:
@gfoidl I wanted to express my gratitude. I learned a ton from your inputs. Thanks for sharing your deep knowledge here. The end result is so much better

@paulomorgado as always thank you for your inputs too. It's always amazing how by doing "small changes" we collectively discover better ways and revisit previous assumptions. 

One of the reasons I'm still around here is because of this community.

gfoidl:
> Thanks for sharing your deep knowledge here. The end result is so much better

That's how open source projects should work 😄.
Well I'm using RabbitMQ, following this repo for some time now, and where I can give my input I'll do so -- as together we are strong and can build great stuff.

lukebakken:
@gfoidl I'd like to invite you as a collaborator to this repo, but when I try to find you by your account, I'm running into an issue. Anyway, I'll figure it out. Thanks everyone! I always learn something new via these contributions.

lukebakken (approved):
Once the latest CI run passes I'll merge.

PR #1944, opened 2026-05-18
Fix URI round-trip for credentials containing special characters

lukebakken:
## Summary

- Fix `SetUri` to split `UserInfo` on the first colon only (via `IndexOf`), allowing colons in the password per RFC 3986
- Fix `GetUri` to percent-encode `UserName` and `Password` via `Uri.EscapeDataString` before assigning to `UriBuilder`, which does not encode colons in the userinfo component
- Add unit tests covering `#`, `:`, `@`, space, `/`, `?`, and `%` in both username and password, exercising both the `SetUri` (pre-encoded URI input) and `GetUri` -> `SetUri` (property round-trip) paths

## Notes

The original discussion (#1943) reports "Invalid port specified" when a password contains `#`. That error comes from .NET's `System.Uri` constructor when given an unescaped string - it is not a bug in this library. However, during investigation I found a real bug: `SetUri` uses `Split(':')` and rejects results with more than two parts, which breaks when credentials contain literal colons. `UriBuilder` does not percent-encode colons in userinfo, so the `GetUri()` -> `SetUri()` round-trip was lossy.

## Test plan

- [x] All 34 `TestAmqpUri` tests pass (including 4 previously failing colon cases)
- [x] Full unit test suite (124 tests) passes with no regressions

### DIFF
--- a/projects/RabbitMQ.Client/ConnectionFactory.cs
+++ b/projects/RabbitMQ.Client/ConnectionFactory.cs
@@ -656,12 +656,12 @@ namespace RabbitMQ.Client
 
             if (false == string.IsNullOrEmpty(UserName))
             {
-                builder.UserName = UserName;
+                builder.UserName = Uri.EscapeDataString(UserName);
             }
 
             if (false == string.IsNullOrEmpty(Password))
             {
-                builder.Password = Password;
+                builder.Password = Uri.EscapeDataString(Password);
             }
 
             if (false == string.IsNullOrEmpty(VirtualHost))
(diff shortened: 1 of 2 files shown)


michaelklishin (dismissed):

projects/RabbitMQ.Client/ConnectionFactory.cs
```diff
@@ -709,15 +709,15 @@ private void SetUri(Uri uri)
             string userInfo = uri.UserInfo;
             if (!string.IsNullOrEmpty(userInfo))
             {
-                string[] userPass = userInfo.Split(':');
-                if (userPass.Length > 2)
+                int colonIndex = userInfo.IndexOf(':');
+                if (colonIndex == -1)
```
gfoidl:
```suggestion
                if (colonIndex < 0)
```
This is one of the tiniest micro-optimizations[^1] as with `if (x == -1)` the cpu has to perform a `cmp` followed by a `jmp`. With `if (x < 0)`, which results in the same for integers, the cpu can fuse the compare and jump, so actually only one instruction needs to be executed in cpu's backend.

[^1]: I don't know if it's actually worth or measurable with a benchmark, one would need a tool like Intel's VTune Profiler to spot it

projects/RabbitMQ.Client/ConnectionFactory.cs
```diff
@@ -709,15 +709,15 @@ private void SetUri(Uri uri)
...
-                    throw new ArgumentException($"Bad user info in AMQP URI: {userInfo}");
+                    UserName = UriDecode(userInfo);
                 }
-                UserName = UriDecode(userPass[0]);
-                if (userPass.Length == 2)
+                else
                 {
-                    Password = UriDecode(userPass[1]);
+                    UserName = UriDecode(userInfo.Substring(0, colonIndex));
+                    Password = UriDecode(userInfo.Substring(colonIndex + 1));
```
gfoidl, line 720: Allocations in connection factory shouldn't matter much, as it's called seldom, but a span based overload in `UrlDecode` would be nice.

lukebakken:
Thanks everyone for your input!

PR #1953, opened 2026-07-07
Validate `shortstr` byte length in `WriteShortstr`

### LINKED ISSUE #1952: The output byte buffer is too small to contain the encoded data
### Describe the bug

Using MassTransit + RabbitMq (internally RabbitMq.Client 7.2.1).
Have a bunch of similar models:
```c#
public sealed record RegisterThatCertainMessage
    : BaseRegisterMessage<MessageDirective, RegisterThatCertainMessage.MessageContent>
{
    public override MessageDirective Code => MessageDirective.SomeDirective;
    
    [MessagePackObject]
    public struct MessageContent
    {
        [MessagePack.Key("someProp")]
        [Required]
        public int SomeProp { get; init; }
    }
}
```

All the models publishing just fine except for one certain model that throws exception.
The model itself has nothing special or unusual (at least I see nothing).
Checked it for non-ascii characters, but it's clean.
Full exchange name length is less than 100 characters, following MassTransit implicit naming pattern like "Full.Namespace.Here:ModelName".
The exchange itself visually is created in RabbitMq (deleted the exchange and then tested anew).
Renaming the message type did not change anything.
Manually overriding exchange name for super-short custom name did not change anything.

### Reproduction steps

1. Build message
2. Try to publish message
3. Exception

### Expected behavior

Message is published, no exception

### Additional context

```c#
S-FAULT rabbitmq://localhost:5673/My.Library.Backend.Shared.Name.Models.Here:RegisterThatCertainMessage dc750000-640f-5d3b-0793-08dedb9d921f My.Library.Backend.Shared.Name.Models.Here:RegisterThatCertainMessage
System.ArgumentException: The output byte buffer is too small to contain the encoded data, encoding codepage '65001' and fallback 'System.Text.EncoderReplacementFallback'. (Parameter 'bytes')
   at System.Text.Encoding.ThrowBytesOverflow()
   at System.Text.Encoding.ThrowBytesOverflow(EncoderNLS encoder, Boolean nothingEncoded)
   at System.Text.Encoding.GetBytesWithFallback(ReadOnlySpan`1 chars, Int32 originalCharsLength, Span`1 bytes, Int32 originalBytesLength, EncoderNLS encoder, Boolean throwForDestinationOverflow)
   at System.Text.Encoding.GetBytesWithFallback(Char* pOriginalChars, Int32 originalCharCount, Byte* pOriginalBytes, Int32 originalByteCount, Int32 charsConsumedSoFar, Int32 bytesWrittenSoFar, Boolean throwForDestinationOverflow)
   at System.Text.UTF8Encoding.GetBytes(Char* chars, Int32 charCount, Byte* bytes, Int32 byteCount)
   at RabbitMQ.Client.Framing.ExchangeDeclare.WriteTo(Span`1 span)
   at RabbitMQ.Client.Impl.SessionBase.TransmitAsync[T](T& cmd, CancellationToken cancellationToken)
   at RabbitMQ.Client.Impl.Channel.ExchangeDeclareAsync(String exchange, String type, Boolean durable, Boolean autoDelete, IDictionary`2 arguments, Boolean passive, Boolean noWait, CancellationToken cancellationToken)
   at MassTransit.RabbitMqTransport.ScopeChannelContext.ExchangeDeclare(String exchange, String type, Boolean durable, Boolean autoDelete, IDictionary`2 arguments, CancellationToken cancellationToken) in /_/src/Transports/MassTransit.RabbitMqTransport/RabbitMqTransport/ScopeChannelContext.cs:line 56
   at MassTransit.RabbitMqTransport.SharedChannelContext.ExchangeDeclare(String exchange, String type, Boolean durable, Boolean autoDelete, IDictionary`2 arguments, CancellationToken cancellationToken) in /_/src/Transports/MassTransit.RabbitMqTransport/RabbitMqTransport/SharedChannelContext.cs:line 51
   at MassTransit.RabbitMqTransport.SharedChannelContext.ExchangeDeclare(String exchange, String type, Boolean durable, Boolean autoDelete, IDictionary`2 arguments, CancellationToken cancellationToken) in /_/src/Transports/MassTransit.RabbitMqTransport/RabbitMqTransport/SharedChannelContext.cs:line 51
   at MassTransit.RabbitMqTransport.Middleware.ConfigureRabbitMqTopologyFilter`1.ConfigureTopology(ChannelContext context, CancellationToken cancellationToken) in /_/src/Transports/MassTransit.RabbitMqTransport/RabbitMqTransport/Middleware/ConfigureRabbitMqTopologyFilter.cs:line 68
   at MassTransit.OneTimeSetupMethod.<>c__DisplayClass5_0.<<SetupPayload>g__SetupAsync|0>d.MoveNext() in /_/src/MassTransit.Abstractions/Middleware/OneTimeSetupMethod.cs:line 37
--- End of stack trace from previous location ---
   at MassTransit.PipeExtensions.OneTimeSetup[T](PipeContext context, OneTimeSetupCallback setupMethod) in /_/src/MassTransit.Abstractions/Middleware/PipeExtensions.cs:line 90
   at MassTransit.RabbitMqTransport.Middleware.ConfigureRabbitMqTopologyFilter`1.Configure(ChannelContext context, CancellationToken cancellationToken) in /_/src/Transports/MassTransit.RabbitMqTransport/RabbitMqTransport/Middleware/ConfigureRabbitMqTopologyFilter.cs:line 55
   at MassTransit.RabbitMqTransport.RabbitMqSendTransportContext.Send[T](ChannelContext transportContext, SendContext`1 sendContext) in /_/src/Transports/MassTransit.RabbitMqTransport/RabbitMqTransport/RabbitMqSendTransportContext.cs:line 107
   at MassTransit.Transports.SendTransport`1.SendPipe`1.Send(TContext context) in /_/src/MassTransit/Transports/SendTransport.cs:line 93
```

lukebakken:
Fixes #1952.

## Problem

When a string's UTF-8 encoding exceeds 255 bytes, the `string` overload of `WireFormatting.WriteShortstr` passed a hardcoded `byte.MaxValue` as the destination byte count to `UTF8.GetBytes`, producing a cryptic framework exception:

```
System.ArgumentException: The output byte buffer is too small to contain the encoded data, encoding codepage '65001' and fallback 'System.Text.EncoderReplacementFallback'. (Parameter 'bytes')
```

This gave no indication of which value was too long. A `shortstr` is length-prefixed by a single octet, so 255 bytes is its hard maximum. The `ReadOnlySpan<byte>` overload already validated this and threw a clear message; the `string` overload is the funnel for every other shortstr site (exchange, queue, routing key, `BasicProperties`, table keys, consumer tags), so all of them shared the defect.

## Fix

Compute the encoded byte count up front and throw the existing clear `ThrowArgumentTooLong` error when it exceeds 255:

```
Value exceeds the maximum allowed length of 255 bytes, was 256 long.
```

Pass the real `byteCount` to `GetBytes`, which is safe because the buffer is pre-sized to the true byte count via `GetRequiredBufferSize`. The `unchecked` cast on the length prefix is no longer needed now that the guard bounds `bytesWritten` to 255.

## Tests

Added unit tests covering the 255-byte boundary, a 256-byte string, and a multi-byte string whose char count is small but byte count exceeds 255. Updated `TestTableEncoding_LongKey` to expect the more specific `ArgumentOutOfRangeException`.

### DIFF
--- a/projects/RabbitMQ.Client/Impl/WireFormatting.Write.cs
+++ b/projects/RabbitMQ.Client/Impl/WireFormatting.Write.cs
@@ -342,19 +342,24 @@ namespace RabbitMQ.Client.Impl
             int bytesWritten = 0;
             if (!string.IsNullOrEmpty(val))
             {
+                int byteCount = UTF8.GetByteCount(val);
+                if (byteCount > byte.MaxValue)
+                {
+                    return ThrowArgumentTooLong(byteCount);
+                }
+
                 unsafe
                 {
-
                     ref byte valDestination = ref destination.GetOffset(1);
                     fixed (char* chars = val)
                     fixed (byte* bytes = &valDestination)
                     {
-                        bytesWritten = UTF8.GetBytes(chars, val!.Length, bytes, byte.MaxValue);
+                        bytesWritten = UTF8.GetBytes(chars, val!.Length, bytes, byteCount);
                     }
                 }
             }
 
-            destination = unchecked((byte)bytesWritten);
+            destination = (byte)bytesWritten;
             return bytesWritten + 1;
         }
 
--- a/projects/RabbitMQ.Client/Impl/WireFormatting.Write.cs
+++ b/projects/RabbitMQ.Client/Impl/WireFormatting.Write.cs
@@ -342,10 +342,17 @@ namespace RabbitMQ.Client.Impl
             int bytesWritten = 0;
             if (!string.IsNullOrEmpty(val))
             {
-                int byteCount = UTF8.GetByteCount(val);
+                // GetMaxByteCount is cheap bit arithmetic that yields an upper bound
+                // on the encoded size. Only fall back to the O(n) exact count when
+                // that upper bound exceeds the shortstr maximum.
+                int byteCount = UTF8.GetMaxByteCount(val!.Length);
                 if (byteCount > byte.MaxValue)
                 {
-                    return ThrowArgumentTooLong(byteCount);
+                    byteCount = UTF8.GetByteCount(val);
+                    if (byteCount > byte.MaxValue)
+                    {
+                        return ThrowArgumentTooLong(byteCount);
+                    }
                 }
 
                 unsafe
--- a/projects/Test/Unit/TestFieldTableFormatting.cs
+++ b/projects/Test/Unit/TestFieldTableFormatting.cs
@@ -129,7 +129,7 @@ namespace Test.Unit
             int bytesNeeded = WireFormatting.GetTableByteCount(t);
             byte[] bytes = new byte[bytesNeeded];
 
-            Assert.Throws<ArgumentException>(() => WireFormatting.WriteTable(ref bytes.GetStart(), t));
+            Assert.Throws<ArgumentOutOfRangeException>(() => WireFormatting.WriteTable(ref bytes.GetStart(), t));
         }
 
         [Fact]
--- a/projects/Test/Unit/TestWriteShortstr.cs
+++ b/projects/Test/Unit/TestWriteShortstr.cs
@@ -1,0 +1,84 @@
+// This source code is dual-licensed under the Apache License, version
+// 2.0, and the Mozilla Public License, version 2.0.
+//
+// The APL v2.0:
+//
+//---------------------------------------------------------------------------
+//   Copyright (c) 2007-2026 Broadcom. All Rights Reserved.
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//       https://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//---------------------------------------------------------------------------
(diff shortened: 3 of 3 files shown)


projects/RabbitMQ.Client/Impl/WireFormatting.Write.cs
```diff
@@ -342,19 +342,24 @@ public static int WriteShortstr(ref byte destination, string? val)
             int bytesWritten = 0;
             if (!string.IsNullOrEmpty(val))
             {
-                unsafe
+                int byteCount = UTF8.GetByteCount(val);
```
gfoidl: `GetByteCount` traverses the whole input which is later done again in `UTF8.GetBytes`, so it's $O(2n)$.

If perf matters one could do something like
```c#
// Fast check for byte count (gives the upper bound)
int byteCount = UTF8.GetMaxByteCount(val.Length);

if (byteCount > byte.MaxValue)
{
    // Limit exceeded, so perform an actual byte count
    byteCount = UTF8.GetByteCount(val);

    if (byteCount > byte.MaxValue)
    {
        return ThrowArgumentTooLong(byteCount);
    }
}

// Rest a given in the PR
```
`GetMaxByteCount` is just a bit arithmetic, so it's $O(n)$.

This approach
* adds a bit more complexitiy to the code
* does a quick check if below the limit, so has advantage for these cases
* penalize cases that exceed the limit -- but then an exception is thrown, so that cost shouldn't matter

lukebakken: Good call, applied in cbbaf3377. Thanks!

gfoidl (approved):

PR #1954, opened 2026-07-16
Fix spurious "buffer too small" in WriteLongstr(string) on 32-bit LAA…

lucapersichini:
… processes

## Problem

`WireFormatting.WriteLongstr(ref byte destination, string val)` calls `UTF8.GetBytes(chars, val.Length, bytes, int.MaxValue)`, passing `int.MaxValue` as the destination byte count.

Internally the encoder computes the end-of-buffer pointer as `bytes + byteCount`. In a **32-bit large-address-aware process** (e.g. an IIS application pool with *Enable 32-Bit Applications* on 64-bit Windows, where usable address space approaches 4 GB), whenever the destination buffer rented from the `ArrayPool` happens to be located **above `0x80000000`**, `bytes + int.MaxValue` overflows and wraps around. The encoder then believes it has zero bytes available and throws:

```
System.ArgumentException: The output byte buffer is too small to contain
the encoded data, encoding 'Unicode (UTF-8)' fallback
'System.Text.EncoderReplacementFallback'.
Parameter name: bytes
   at System.Text.Encoding.ThrowBytesOverflow()
   at System.Text.Encoding.ThrowBytesOverflow(EncoderNLS encoder, Boolean nothingEncoded)
   at System.Text.UTF8Encoding.GetBytes(Char* chars, Int32 charCount, Byte* bytes, Int32 byteCount, EncoderNLS baseEncoder)
   at System.Text.UTF8Encoding.GetBytes(Char* chars, Int32 charCount, Byte* bytes, Int32 byteCount)
   at RabbitMQ.Client.Impl.WireFormatting.<WriteLongstr>g__GetBytes|39_0(Byte& destination, String val)
   at RabbitMQ.Client.Impl.WireFormatting.WriteFieldValue(Byte& destination, Object value)
   at RabbitMQ.Client.Impl.WireFormatting.WriteTable(Byte& destination, IDictionary`2 val)
   at RabbitMQ.Client.BasicProperties.RabbitMQ.Client.IAmqpWriteable.WriteTo(Span`1 span)
   at RabbitMQ.Client.Impl.Framing.Header.WriteTo[T](Span`1 span, UInt16 channel, T& header, Int32 bodyLength)
   at RabbitMQ.Client.Impl.Framing.SerializeToFrames[TMethod,THeader](...)
   at RabbitMQ.Client.Impl.Channel.BasicPublishAsync(...)
```

The failure is intermittent and content-independent: the very same publish succeeds or fails depending only on where the pooled buffer lands in the address space. In our production system (ASP.NET on IIS, 32-bit worker process) it appears once the worker's virtual address space grows past 2 GB, and disappears after an IIS recycle (address space pressure resets). Once the process is above the threshold we measured a stable **~1% failure rate** on publishes, so this is reliably reproducible, not a fluke.

Observed with 7.1.2 and 7.2.1; the code on `main` is unchanged.

Related: #1952 shows the same exception type for `shortstr`, but that case is about the 255-byte protocol limit. This one is different: no limit is actually exceeded — it is pure 32-bit pointer-arithmetic overflow caused by the `int.MaxValue` argument.

Notably, `WriteShortstr(string)` **already bounds the byte count** it passes to `UTF8.GetBytes` (since #1953, via `GetMaxByteCount` / `GetByteCount`). This PR applies the equivalent safety to `WriteLongstr(string)`.

## Reproduction

Standalone program, deterministic (100% reproducible). Must be compiled x86 **with the LargeAddressAware flag** (IIS 32-bit worker processes are LAA by default):

```csharp
// Allocate one buffer at the top of the address space (> 2 GB) and one
// normally (< 2 GB), then run the exact call WriteLongstr makes:
IntPtr high = VirtualAlloc(IntPtr.Zero, (UIntPtr)65536,
    MEM_COMMIT | MEM_RESERVE | MEM_TOP_DOWN, PAGE_READWRITE);
IntPtr low  = VirtualAlloc(IntPtr.Zero, (UIntPtr)65536,
    MEM_COMMIT | MEM_RESERVE, PAGE_READWRITE);

string val = "some.assembly.qualified.name"; // any short ASCII string
fixed (char* chars = val)
{
    // low buffer (e.g. 0x00CD0000): writes fine
    Encoding.UTF8.GetBytes(chars, val.Length, (byte*)low, int.MaxValue);

    // high buffer (e.g. 0xFFAA0000): throws the exact exception above,
    // because (bytes + int.MaxValue) wraps around 2^32
    Encoding.UTF8.GetBytes(chars, val.Length, (byte*)high, int.MaxValue);
}
```

## Fix

Pass the exact required byte count (`UTF8.GetByteCount(val)`) instead of `int.MaxValue`. Since the encoded bytes fit inside the (valid) rented buffer by construction, `destination + exactByteCount` can never wrap.

The `GetMaxByteCount` fast path used by `WriteShortstr` was deliberately not used here: for `longstr` there is no 255-byte cap to validate against, and the *max* count (≈ 3×length) could still theoretically wrap for very long strings when the buffer sits near the very top of the address space. The exact count is provably safe for any valid destination buffer. The extra `GetByteCount` pass only affects the string overload used for field table values (typically short header strings).

## Testing note

This cannot be covered by a regular unit test: it requires a 32-bit large-address-aware host whose allocations land above 0x80000000, which cannot be forced from managed code in a normal CI run. The standalone repro above demonstrates the failure deterministically.

### DIFF
--- a/projects/RabbitMQ.Client/Impl/WireFormatting.Write.cs
+++ b/projects/RabbitMQ.Client/Impl/WireFormatting.Write.cs
@@ -373,8 +373,19 @@ namespace RabbitMQ.Client.Impl
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static int WriteLongstr(ref byte destination, string val)
         {
-            static int GetBytes(ref byte destination, string val, int byteCount)
+            static int GetBytes(ref byte destination, string val)
             {
+                // On 32-bit (large-address-aware) processes "bytes + int.MaxValue"
+                // wraps around whenever the destination buffer is located above
+                // 0x80000000, making the encoder throw a spurious "output byte
+                // buffer is too small" ArgumentException. Use the exact byte count
+                // there. On 64-bit the addition cannot overflow, so the original
+                // one-pass behavior is preserved; IntPtr.Size is a JIT-time
+                // constant, so the check itself is free.
+                int byteCount = IntPtr.Size == 4
+                    ? UTF8.GetByteCount(val)
+                    : int.MaxValue;
+
                 unsafe
                 {
                     fixed (char* chars = val)
@@ -385,13 +396,7 @@ namespace RabbitMQ.Client.Impl
                 }
             }
 
-            // Do not pass int.MaxValue as the destination byte count: the encoder
-            // computes "bytes + byteCount", which wraps around in 32-bit
-            // large-address-aware processes when the destination buffer is located
-            // above 0x80000000, throwing a spurious "output byte buffer is too
-            // small" ArgumentException. The exact byte count cannot wrap for a
-            // valid destination buffer, since the encoded bytes fit within it.
-            int bytesWritten = string.IsNullOrEmpty(val) ? 0 : GetBytes(ref destination.GetOffset(4), val, UTF8.GetByteCount(val));
+            int bytesWritten = string.IsNullOrEmpty(val) ? 0 : GetBytes(ref destination.GetOffset(4), val);
             NetworkOrderSerializer.WriteUInt32(ref destination, (uint)bytesWritten);
             return bytesWritten + 4;
         }
--- a/projects/RabbitMQ.Client/Impl/WireFormatting.Write.cs
+++ b/projects/RabbitMQ.Client/Impl/WireFormatting.Write.cs
@@ -373,19 +373,25 @@ namespace RabbitMQ.Client.Impl
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static int WriteLongstr(ref byte destination, string val)
         {
-            static int GetBytes(ref byte destination, string val)
+            static int GetBytes(ref byte destination, string val, int byteCount)
             {
                 unsafe
                 {
                     fixed (char* chars = val)
                     fixed (byte* bytes = &destination)
                     {
-                        return UTF8.GetBytes(chars, val.Length, bytes, int.MaxValue);
+                        return UTF8.GetBytes(chars, val.Length, bytes, byteCount);
                     }
                 }
             }
 
-            int bytesWritten = string.IsNullOrEmpty(val) ? 0 : GetBytes(ref destination.GetOffset(4), val);
+            // Do not pass int.MaxValue as the destination byte count: the encoder
+            // computes "bytes + byteCount", which wraps around in 32-bit
+            // large-address-aware processes when the destination buffer is located
+            // above 0x80000000, throwing a spurious "output byte buffer is too
+            // small" ArgumentException. The exact byte count cannot wrap for a
+            // valid destination buffer, since the encoded bytes fit within it.
+            int bytesWritten = string.IsNullOrEmpty(val) ? 0 : GetBytes(ref destination.GetOffset(4), val, UTF8.GetByteCount(val));
             NetworkOrderSerializer.WriteUInt32(ref destination, (uint)bytesWritten);
             return bytesWritten + 4;
         }


michaelklishin:
I'm pretty sure 32-bit systems weren't a target for this client even in its early days in 2007-2010.

@lukebakken I suggest that we close this.

lucapersichini:
> I'm pretty sure 32-bit systems weren't a target for this client even in its early days in 2007-2010.
> 
> @lukebakken I suggest that we close this.

Thanks for taking a look @michaelklishin. I'd respectfully push back on closing, for three reasons:

 - The README's "Supported Platforms and .NET Releases" section states that 7.x requires ".NET framework 4.6.1 or any supported .NET version", with no restriction on process architecture or bitness. A very common deployment model for .NET Framework apps is an IIS application pool with Enable 32-Bit Applications — required whenever the app loads 32-bit native dependencies, which is still widespread in enterprise software.
 - This patch doesn't add any 32-bit-specific code path. It simply stops passing int.MaxValue as the destination buffer size to an unsafe pointer-based Encoding.GetBytes overload — an argument that is incorrect on any platform, and merely happens to be harmless on 64-bit because the address space is large enough that bytes + int.MaxValue cannot wrap. It also aligns WriteLongstr(string) with what WriteShortstr(string) already does since #1953 (pass a real byte count).
- The impact is not hypothetical: we hit this in production (ASP.NET on IIS, 32-bit worker process) with a stable ~1% publish failure rate once the worker's address space grows past 2 GB, and the PR description includes a deterministic standalone repro.
The change is a single method, with no API or behavioral change on 64-bit, and CI is green. If the team's position is that 32-bit processes are out of scope, it would be worth stating that explicitly under "Supported Platforms" — but given this is a one-line correctness fix that benefits all platforms, I hope it can be reconsidered.

michaelklishin:
@lucapersichini so you yourself confirm that you run a 32-bit process with this library that was never really designed for that.

The README can be fixed once. The burden of maintaining an architecture that is very niche is permanent.

paulomorgado:
I'm curious, @lucapersichini. What supported version of Windows is 32bit?

lucapersichini:
> I'm curious, @lucapersichini. What supported version of Windows is 32bit?

@paulomorgado To clarify: the OS is 64-bit Windows Server, fully supported. Only the IIS app pool runs 32-bit (“Enable 32-Bit Applications = True”), which is still a very common and supported IIS configuration for apps with legacy native dependencies. So no 32-bit Windows involved — just a 32-bit process on a 64-bit OS.

Thanks for picking it up!

projects/RabbitMQ.Client/Impl/WireFormatting.Write.cs
```diff
@@ -373,19 +373,25 @@ public static int WriteShortstr(ref byte destination, string? val)
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static int WriteLongstr(ref byte destination, string val)
         {
-            static int GetBytes(ref byte destination, string val)
+            static int GetBytes(ref byte destination, string val, int byteCount)
```
danielmarbach: Would this be the better fix?

```csharp
static int GetBytes(ref byte destination, string val)                                                                                                                                                                                               
   {                                                                                                                                                                                                                                                   
       int byteCount = IntPtr.Size == 4                                                                                                                                                                                                                
           ? UTF8.GetByteCount(val)                                                                                                                                                                                                                    
           : int.MaxValue;                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                       
       unsafe                                                                                                                                                                                                                                          
       {                                                                                                                                                                                                                                               
           fixed (char* chars = val)                                                                                                                                                                                                                   
           fixed (byte* bytes = &destination)                                                                                                                                                                                                          
           {                                                                                                                                                                                                                                           
               return UTF8.GetBytes(chars, val.Length, bytes, byteCount);                                                                                                                                                                              
           }                                                                                                                                                                                                                                           
       }                                                                                                                                                                                                                                               
   }
```

Rationale:                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                       
 - The bug only exists in 32-bit processes with high addresses.                                                                                                                                                                                        
 - On 64-bit, int.MaxValue cannot realistically overflow the managed heap pointer, so retain the one-pass behavior.                                                                                                                                    
 - On 32-bit, use the exact count to guarantee safe pointer arithmetic but with the drawback of the performance regression mentioned in https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/1953#discussion_r3542428690.

lucapersichini: Good call @danielmarbach, that's a better shape for the fix — updated the PR accordingly. Worth noting that IntPtr.Size is a JIT-time constant, so the branch is eliminated entirely on 64-bit and the original one-pass behavior is preserved there at zero cost, while 32-bit (the only place the wrap can happen) gets the exact count. I kept a short comment explaining why int.MaxValue is unsafe on 32-bit, so it doesn't get "simplified" back in the future.

danielmarbach: This could be a potential counter proposal https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/1956 but the machinery is a bit heavier. I ran a perf suite, which gave me roughly 

image

Running against the .NET 10 upgraded benchmark project on my M3 Max which makes me think the special handling for 32 bit is the best way forward as pushed by @lucapersichini in the latest commit

lucapersichini: Thanks for benchmarking this @danielmarbach! And agreed, on 64-bit the JIT folds the branch away, so the cost lands only on 32-bit where the bug actually lives.

lucapersichini:
The net472 job failed on TestConnectionShutdown.TestCleanClosureWithSocketClosedOutOfBand with an OperationCanceledException in the CloseAsync wait. I don't think it's related to this PR:

the failure happens in the close/shutdown wait path, which doesn't serialize any headers at all. The code I changed runs when the connection is opened (client-properties table), and in that test the connection was already up and working before the close was attempted. Also, on the 64-bit CI runners my change takes the int.MaxValue branch anyway, so it behaves exactly like master there.
FWIW the same test passed in 2 ms on net8.0 in this very run. Looking at the test itself, it closes the socket out of band and then tries a clean CloseAsync, and it already catches two different exception types as acceptable outcomes (one with a "TODO: ideally we'd not see this exception!" comment), so it seems to be a known-flaky area, especially on .NET Framework.

Does that analysis make sense, or am I missing something? If it looks right, could someone re-run the failed job? Thanks!

danielmarbach:
>  If it looks right, could someone re-run the failed job?

Re-running

danielmarbach (approved):

danielmarbach:
@lukebakken @michaelklishin please give this a final review

lukebakken (approved):
Correct, minimal fix. Passing int.MaxValue as a buffer size to the unsafe GetBytes overload is technically wrong on any platform; the IntPtr.Size branch keeps 64-bit behavior unchanged (JIT-folded, zero cost) and uses the exact byte count only on 32-bit where the pointer wrap actually occurs. Thanks!

PR #1971, opened 2026-07-29
Fix spurious RPC continuation timeouts under concurrent load (#1964)

### LINKED ISSUE #1964: Flaky tests: concurrent exchange/queue declare-delete fail with cancelled RPC and mismatched continuation
### Summary

`TestExchangeDeclare.TestConcurrentExchangeDeclareAndDelete` and `TestQueueDeclare.TestConcurrentQueueDeclareAndBindAsync` flake in full-suite integration runs. Both fail the same way: an RPC on the shared `_channel` faults with `TaskCanceledException`, and the fixture's teardown then reports a **mismatched RPC continuation** - `InvalidOperationException: Received unexpected command of type <PreviousOpOk>!`.

These are pre-existing flakes, unrelated to any in-flight shutdown work. They were found while validating an unrelated fix and confirmed against an unmodified tree.

### Evidence

Two full `Integration.csproj` runs (net8.0, Linux/WSL, local broker), same machine, minutes apart. The failing test differs between runs while the failure mode is identical:

| Run | Result | Failing test | Teardown warning |
|---|---|---|---|
| A | Failed: 2, Passed: 190, Skipped: 13 | `TestQueueDeclare.TestConcurrentQueueDeclareAndBindAsync` (1m27s) | `Received unexpected command of type QueueDeleteOk!` (also `ExchangeDeclareOk!`) |
| B (unmodified tree) | Failed: 1, Passed: 191, Skipped: 13 | `TestExchangeDeclare.TestConcurrentExchangeDeclareAndDelete` (39s) | `Received unexpected command of type ExchangeDeleteOk!` |

Run B stack trace:

```
got unexpected Exception: System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at RabbitMQ.Client.Impl.Channel.ExchangeDeleteAsync(...) in Impl/Channel.cs:line 1294
   at RabbitMQ.Client.Impl.AutorecoveringChannel.ExchangeDeleteAsync(...) in Impl/AutorecoveringChannel.cs:line 427
   at Test.Integration.TestExchangeDeclare.<...>b__1>d.MoveNext() in TestExchangeDeclare.cs:line 163
   at Test.Integration.TestExchangeDeclare.TestConcurrentExchangeDeclareAndDelete() in TestExchangeDeclare.cs:line 185

[WARNING] IntegrationFixture.CloseAsync() exception: System.InvalidOperationException: Received unexpected command of type ExchangeDeleteOk!
    at RabbitMQ.Client.Impl.Channel.CloseAsync(ShutdownEventArgs args, Boolean abort) in Impl/Channel.cs:line 243
    at Test.IntegrationFixture.DisposeAsync() in Test/Common/IntegrationFixture.cs:line 185
```

Both tests pass consistently when run in isolation - 3/3 repeat runs of just these two tests passed, and a `--filter FullyQualifiedName~TestQueueDeclare` run passed 6/6. They only fail under full-suite load, which points at timing rather than logic in the tests themselves.

### Analysis

The two symptoms are causally linked, and the second is the more interesting one:

1. Under full-suite load an RPC exceeds its `ContinuationTimeout` (the fixture sets `ContinuationTimeout = WaitSpan`, 30s locally / 60s in CI) or is otherwise cancelled, so `await k` throws and the caller sees `TaskCanceledException`.
2. The now-abandoned continuation's response still arrives from the broker. The `-Ok` for the *previous* operation is then matched against the *next* continuation in the queue, which raises `HandleUnexpectedCommand` (`Impl/AsyncRpcContinuations.cs:156`).

So a single timed-out RPC desynchronizes the channel's continuation queue for every subsequent operation. There is already machinery intended to absorb this - `RpcContinuationQueue.RpcCanceled` / `ShouldIgnoreCommand`, added for #1802 - but it records only **one** timed-out command-id pair (`Interlocked.Exchange(ref _lastTimedOutCommandIds, ...)`) and its comment assumes "a late response is always the very next incoming command". These tests drive many operations concurrently against one channel, so that single-slot, strict-ordering assumption looks like the weak point worth checking first.

Whether the *root* problem is (a) the tests being too aggressive for the default timeout under parallel load, or (b) the #1802 late-response tracking being insufficient for concurrent RPCs, needs confirmation. Note (b) would be a library bug affecting users, not just tests, since a single timed-out RPC would poison the channel.

### Reproduction

```
dotnet test projects/Test/Integration/Integration.csproj -c Release -f net8.0
```

Run the full suite (~4-5 min); does not reproduce reliably with a narrow `--filter`. Expect roughly 1-2 failures per run among the concurrent declare/delete tests.

### Suggested next steps

- Instrument `RpcContinuationQueue` to confirm whether the desync is a late response arriving after a cancelled continuation, and whether more than one timed-out RPC is ever outstanding.
- Decide whether `_lastTimedOutCommandIds` needs to track multiple outstanding ids, or whether the ordering assumption from #1802 holds here.
- Only after that, consider whether the tests also need bounded concurrency or a longer timeout.

lukebakken:
> [!NOTE]
> This description was written by Claude (Anthropic's Claude Code), which also did the investigation and wrote the fix, under the direction of @lukebakken, who reviewed and approved each commit before it was made. The measurements and test results are from actual runs, not estimates.

Fixes #1964.

## Problem

Three bugs, all in the same area. The issue framed the cause as concurrent RPCs on one channel, but `_rpcSemaphore` does serialize them, so that is not the defect.

**1. The continuation timeout was charged for time the operation could not run.** `AsyncRpcContinuation` armed its `CancellationTokenSource(continuationTimeout)` in the constructor, and every RPC method in `Channel` constructs the continuation *before* awaiting `_rpcSemaphore.WaitAsync`. So the queueing delay behind other operations was billed to the operation. Measured with 256 concurrent `ExchangeDeclareAsync` calls on one channel: 9 ms uncontended, but p50 100 ms and max 196 ms against a 200 ms budget. The broker was not slow. This also contradicts the documented meaning of `ContinuationTimeout`, "amount of time protocol operations are allowed to take" - the operation, not the queue ahead of it.

**2. Single-slot late-response tracking lost responses.** Each timeout leaves a late response in flight, and several can be outstanding at once: the semaphore serializes *requests*, but nothing makes the broker's reply arrive before the next request is sent. The #1802 machinery recorded only one entry, so each new `RpcCanceled` overwrote the previous one and the unabsorbed `-Ok` was matched against an unrelated continuation, giving `Received unexpected command of type ExchangeDeclareOk!`. Confirmed with temporary instrumentation that printed the overwrite and the resulting failure in the same run.

**3. Server-originated frames consumed the record.** `Channel.HandleCommandAsync` calls `ShouldIgnoreCommand` for *every* incoming command, so a `basic.deliver`, `basic.ack` or `channel.flow` arriving between a timeout and its late response consumed the record of that timed-out RPC. Not new in this branch - the single-slot code on `main` consumed its slot unconditionally too, under the same false invariant.

## Fix

**Timeout accounting.** The timeout source is now constructed unarmed, and `StartTimeout()` arms it once the semaphore is held. Two details make this more than a rename: `CancelAfter` cannot un-cancel, so leaving it armed at construction and merely withholding the token from `WaitAsync` would not work; and a continuation that never calls `StartTimeout()` has no timeout at all, which was the case for `MaybeConfirmSelectAsync`. All 22 construction sites now pair one-to-one with a `StartTimeout()` call. This does not introduce unbounded waiting, since whoever holds the semaphore has its own timeout armed. `BasicPublishAsync` does not take the semaphore, so the publish hot path is untouched.

**Late-response tracking.** Replaces the single-slot union struct with a `ConcurrentQueue<TimedOutRpc>`, one entry per timed-out RPC, bounded at 128. AMQP 0-9-1 enforces strict request-response ordering, so late responses arrive in send order and only the oldest entry can match; `TryDequeue` consumes it either way, since a non-match means that response is never coming.

**Server-originated frames.** `ShouldIgnoreCommand` now skips the queue for commands the broker sends of its own accord. The list is derived from the switch in `DispatchCommandAsync`, which is authoritative because it returns `false` only from its `default` case, meaning "this is an RPC response". `ChannelCloseOk`, `ConnectionSecure` and `ConnectionTune` are deliberately excluded even though dispatch handles them, because continuations name them as expected replies and they must stay absorbable. The docstring records that reasoning and the maintenance rule.

## Tests

All three bugs reproduce with no broker at all, since `RpcContinuationQueue` and `SimpleAsyncRpcContinuation` are visible to `Unit`. 13 new unit tests cover: the timeout being unarmed until the operation can run, the full budget being granted at `StartTimeout`, the dispose race, consecutive timeouts absorbing every late response in send order, the bound dropping oldest-first, the shutdown drain, 8 server-originated frame cases, and the deliberate exclusions staying absorbable.

Each was checked by reverting the relevant part of the fix and re-running, not by assuming. Two notes:

- Three pre-existing tests used `BasicAck` as their "non-matching command ID", which encoded bug 3: after the guard they would pass via the early-out rather than the mismatch path they are named for. Switched to `ExchangeDeclareOk`.
- `TestCallerTokenStillCancelsBeforeTimeoutStarts` initially hung, exposing a design limit worth stating: the continuation's task is completed only by a response, a shutdown, or the timeout callback, and the caller's token is not registered against it. Production is safe because `StartTimeout()` always precedes the await, but that is an invariant rather than an accident, and the test documents it.

## Verification

One Linux box, Toxiproxy enabled.

- **Unit** 145 passed. **Integration** 200 passed / 6 skipped on two consecutive full runs; `main` on the same box gives 197 passed / **3 failed**, failing exactly `TestConcurrentQueueDeclare`, `TestConcurrentQueueDeclareAndBindAsync` and `TestConcurrentExchangeDeclareAndDelete`. **SequentialIntegration** 64 passed / 3 skipped.
- **`TestRpcContinuationTimeout_GH1802`** passes 5/5 in isolation. This one matters most: it is the regression test for the exact #1802 mechanism replaced here, and it is Toxiproxy-gated so it does not run unless `RABBITMQ_TOXIPROXY_TESTS` is set.
- **Standalone 256-operation reproducer:** 2 mismatches in 6 runs before, 0 in 8 after.
- Clean build on both TFMs with `--no-incremental`: 0 warnings, 0 errors. `dotnet format --verify-no-changes` clean.

Skips are TLS trust store, max message body size and heartbeat cases, environmental on this box. One caveat, since this PR is about a flaky test: an intermediate full run showed `TestRpcContinuationTimeout_GH1802` and `TestQueueRecoveryWithManyQueues` failing, and both passed in the two later full runs. I attribute those to load on this box rather than to the change, and note it rather than reporting only the clean runs.

### DIFF
--- a/projects/RabbitMQ.Client/Impl/Channel.cs
+++ b/projects/RabbitMQ.Client/Impl/Channel.cs
@@ -990,11 +990,18 @@ namespace RabbitMQ.Client.Impl
                 {
                     enqueued = Enqueue(k);
 
+                    // rabbitmq/rabbitmq-dotnet-client#1964
+                    // Send before the try whose catch records a timed-out RPC, so a write
+                    // failure (cancellation, IOException, AlreadyClosedException) does not
+                    // record an entry for a request that never reached the wire. Order-based
+                    // late-response matching would then desync permanently. Only a failure
+                    // while awaiting the response, by which point the frame is sent, should
+                    // record.
+                    await ModelSendAsync(in method, k.CancellationToken)
+                        .ConfigureAwait(false);
+
                     try
                     {
-                        await ModelSendAsync(in method, k.CancellationToken)
-                            .ConfigureAwait(false);
-
                         AssertResultIsTrue(await k);
                     }
                     catch
@@ -1426,11 +1433,17 @@ namespace RabbitMQ.Client.Impl
                 {
                     enqueued = Enqueue(k);
 
+                    // rabbitmq/rabbitmq-dotnet-client#1964
+                    // Send before the try whose catch records a timed-out RPC, so a
+                    // cancellation during the write does not record an entry for a request
+                    // that never reached the wire. Order-based late-response matching would
+                    // then desync permanently. Only a cancellation while awaiting the
+                    // response, by which point the frame is sent, should record.
+                    await ModelSendAsync(in method, k.CancellationToken)
+                        .ConfigureAwait(false);
+
                     try
                     {
-                        await ModelSendAsync(in method, k.CancellationToken)
-                            .ConfigureAwait(false);
-
                         QueueDeclareOk result = await k;
                         if (false == passive)
                         {
--- a/projects/RabbitMQ.Client/Impl/Channel.cs
+++ b/projects/RabbitMQ.Client/Impl/Channel.cs
@@ -996,12 +996,9 @@ namespace RabbitMQ.Client.Impl
                     enqueued = Enqueue(k);
 
                     // rabbitmq/rabbitmq-dotnet-client#1964
-                    // Send before the try whose catch records a timed-out RPC, so a write
-                    // failure (cancellation, IOException, AlreadyClosedException) does not
-                    // record an entry for a request that never reached the wire. Order-based
-                    // late-response matching would then desync permanently. Only a failure
-                    // while awaiting the response, by which point the frame is sent, should
-                    // record.
+                    // Send outside the try: cancelling the write must not record a
+                    // timed-out RPC for a request that never reached the wire, which
+                    // would desync order-based late-response matching permanently.
                     await ModelSendAsync(in method, k.CancellationToken)
                         .ConfigureAwait(false);
 
@@ -1009,7 +1006,7 @@ namespace RabbitMQ.Client.Impl
                     {
                         AssertResultIsTrue(await k);
                     }
-                    catch
+                    catch (OperationCanceledException)
                     {
                         _continuationQueue.RpcCanceled(k.ResponseReceived, k.HandledProtocolCommandIds);
                         throw;
@@ -1052,7 +1049,7 @@ namespace RabbitMQ.Client.Impl
                 {
                     return await k;
                 }
-                catch
+                catch (OperationCanceledException)
                 {
                     _continuationQueue.RpcCanceled(k.ResponseReceived, k.HandledProtocolCommandIds);
                     throw;
@@ -1448,11 +1445,9 @@ namespace RabbitMQ.Client.Impl
                     enqueued = Enqueue(k);
 
                     // rabbitmq/rabbitmq-dotnet-client#1964
-                    // Send before the try whose catch records a timed-out RPC, so a
-                    // cancellation during the write does not record an entry for a request
-                    // that never reached the wire. Order-based late-response matching would
-                    // then desync permanently. Only a cancellation while awaiting the
-                    // response, by which point the frame is sent, should record.
+                    // Send outside the try: cancelling the write must not record a
+                    // timed-out RPC for a request that never reached the wire, which
+                    // would desync order-based late-response matching permanently.
                     await ModelSendAsync(in method, k.CancellationToken)
                         .ConfigureAwait(false);
 
--- a/projects/RabbitMQ.Client/Impl/Channel.cs
+++ b/projects/RabbitMQ.Client/Impl/Channel.cs
@@ -223,9 +223,10 @@ namespace RabbitMQ.Client.Impl
             await _rpcSemaphore.WaitAsync(k.CancellationToken)
                 .ConfigureAwait(false);
 
-            k.StartTimeout();
             try
             {
+                k.StartTimeout();
+
                 ChannelShutdownAsync += k.OnConnectionShutdownAsync;
                 enqueued = Enqueue(k);
                 ConsumerDispatcher.Quiesce();
@@ -290,9 +291,10 @@ namespace RabbitMQ.Client.Impl
             await _rpcSemaphore.WaitAsync(k.CancellationToken)
                 .ConfigureAwait(false);
 
-            k.StartTimeout();
             try
             {
+                k.StartTimeout();
+
                 enqueued = Enqueue(k);
 
                 try
@@ -336,9 +338,10 @@ namespace RabbitMQ.Client.Impl
             await _rpcSemaphore.WaitAsync(k.CancellationToken)
                 .ConfigureAwait(false);
 
-            k.StartTimeout();
             try
             {
+                k.StartTimeout();
+
                 enqueued = Enqueue(k);
 
                 try
@@ -384,9 +387,10 @@ namespace RabbitMQ.Client.Impl
             await _rpcSemaphore.WaitAsync(k.CancellationToken)
                 .ConfigureAwait(false);
 
-            k.StartTimeout();
             try
             {
+                k.StartTimeout();
+
                 enqueued = Enqueue(k);
 
                 var method = new ChannelOpen();
@@ -975,9 +979,10 @@ namespace RabbitMQ.Client.Impl
             await _rpcSemaphore.WaitAsync(k.CancellationToken)
                 .ConfigureAwait(false);
 
-            k.StartTimeout();
             try
             {
+                k.StartTimeout();
+
                 var method = new BasicCancel(consumerTag, noWait);
 
                 if (noWait)
@@ -1033,9 +1038,10 @@ namespace RabbitMQ.Client.Impl
             await _rpcSemaphore.WaitAsync(k.CancellationToken)
                 .ConfigureAwait(false);
 
-            k.StartTimeout();
             try
             {
+                k.StartTimeout();
+
                 enqueued = Enqueue(k);
 
                 var method = new BasicConsume(queue, consumerTag, noLocal, autoAck, exclusive, false, arguments);
@@ -1069,9 +1075,10 @@ namespace RabbitMQ.Client.Impl
             await _rpcSemaphore.WaitAsync(k.CancellationToken)
                 .ConfigureAwait(false);
 
-            k.StartTimeout();
             try
             {
+                k.StartTimeout();
+
                 enqueued = Enqueue(k);
 
                 var method = new BasicGet(queue, autoAck);
@@ -1125,9 +1132,10 @@ namespace RabbitMQ.Client.Impl
             await _rpcSemaphore.WaitAsync(k.CancellationToken)
                 .ConfigureAwait(false);
 
-            k.StartTimeout();
             try
             {
+                k.StartTimeout();
+
                 enqueued = Enqueue(k);
 
                 byte[] newSecretBytes = Encoding.UTF8.GetBytes(newSecret);
@@ -1161,9 +1169,10 @@ namespace RabbitMQ.Client.Impl
             await _rpcSemaphore.WaitAsync(k.CancellationToken)
                 .ConfigureAwait(false);
 
-            k.StartTimeout();
             try
             {
+                k.StartTimeout();
+
                 enqueued = Enqueue(k);
 
                 var method = new BasicQos(prefetchSize, prefetchCount, global);
@@ -1197,9 +1206,10 @@ namespace RabbitMQ.Client.Impl
             await _rpcSemaphore.WaitAsync(k.CancellationToken)
                 .ConfigureAwait(false);
 
-            k.StartTimeout();
             try
             {
+                k.StartTimeout();
+
                 var method = new ExchangeBind(destination, source, routingKey, noWait, arguments);
 
                 if (noWait)
@@ -1251,9 +1261,10 @@ namespace RabbitMQ.Client.Impl
             await _rpcSemaphore.WaitAsync(k.CancellationToken)
                 .ConfigureAwait(false);
 
-            k.StartTimeout();
             try
             {
+                k.StartTimeout();
+
                 var method = new ExchangeDeclare(exchange, type, passive, durable, autoDelete, false, noWait, arguments);
                 if (noWait)
                 {
@@ -1296,9 +1307,10 @@ namespace RabbitMQ.Client.Impl
             await _rpcSemaphore.WaitAsync(k.CancellationToken)
                 .ConfigureAwait(false);
 
-            k.StartTimeout();
             try
             {
+                k.StartTimeout();
+
                 var method = new ExchangeDelete(exchange, ifUnused, Nowait: noWait);
 
                 if (noWait)
@@ -1343,9 +1355,10 @@ namespace RabbitMQ.Client.Impl
             await _rpcSemaphore.WaitAsync(k.CancellationToken)
                 .ConfigureAwait(false);
 
-            k.StartTimeout();
             try
             {
+                k.StartTimeout();
+
                 var method = new ExchangeUnbind(destination, source, routingKey, noWait, arguments);
 
                 if (noWait)
@@ -1411,9 +1424,10 @@ namespace RabbitMQ.Client.Impl
             await _rpcSemaphore.WaitAsync(k.CancellationToken)
                 .ConfigureAwait(false);
 
-            k.StartTimeout();
             try
             {
+                k.StartTimeout();
+
                 var method = new QueueDeclare(queue, passive, durable, exclusive, autoDelete, noWait, arguments);
 
                 if (noWait)
@@ -1476,9 +1490,10 @@ namespace RabbitMQ.Client.Impl
             await _rpcSemaphore.WaitAsync(k.CancellationToken)
                 .ConfigureAwait(false);
 
-            k.StartTimeout();
             try
             {
+                k.StartTimeout();
+
                 var method = new QueueBind(queue, exchange, routingKey, noWait, arguments);
 
                 if (noWait)
@@ -1538,9 +1553,10 @@ namespace RabbitMQ.Client.Impl
             await _rpcSemaphore.WaitAsync(k.CancellationToken)
                 .ConfigureAwait(false);
 
-            k.StartTimeout();
             try
             {
+                k.StartTimeout();
+
                 var method = new QueueDelete(queue, ifUnused, ifEmpty, noWait);
 
                 if (noWait)
@@ -1584,9 +1600,10 @@ namespace RabbitMQ.Client.Impl
             await _rpcSemaphore.WaitAsync(k.CancellationToken)
                 .ConfigureAwait(false);
 
-            k.StartTimeout();
             try
             {
+                k.StartTimeout();
+
                 enqueued = Enqueue(k);
 
                 var method = new QueuePurge(queue, false);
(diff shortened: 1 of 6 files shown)


danielmarbach:
When do you need the review? I might need a few days

lukebakken:
@danielmarbach no hurry at all. I'll be busy doing 🧞- assisted reviews of the OTel library and hooks in this library prior to the `7.3.0` release.

The 🧞 has been doing some pretty nice work resolving some hairy concurrency issues around cancellation.

projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
```diff
@@ -171,34 +203,104 @@ public void RpcCanceled(bool responseReceived, ReadOnlySpan<ProtocolCommandId> p
...
 
-            var ids = new LastTimedOutCommandIds { RawValue = raw };
-            return commandId == ids.First || (ids.Second != default && commandId == ids.Second);
+            // AMQP 0-9-1 enforces strict request-response ordering on a channel, so late
+            // responses arrive in the order the requests were sent. Only the oldest
+            // outstanding entry can match; consume it either way, since a non-match means
+            // that response is never coming.
+            if (_timedOutRpcs.TryDequeue(out TimedOutRpc timedOut))
+            {
+                return timedOut.Matches(commandId);
```
danielmarbach, line 258: Should we make the struct a record struct with equality override and using that instead of matches? Might be slightly more idiomatic

projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
```diff
@@ -69,25 +69,48 @@ public void Dispose()
 
         private const int CommandIdBufferLength = 2;
 
-        // Two ProtocolCommandId values (each uint) unioned with a single long
-        // so that Interlocked.Exchange can read/write both atomically.
-        // Since ProtocolCommandId : uint has no zero-valued member,
-        // default(LastTimedOutCommandIds).RawValue == 0L means "no timed-out command".
-        [StructLayout(LayoutKind.Explicit)]
-        private struct LastTimedOutCommandIds
```
danielmarbach, line 77: My beautiful friend, we had a good time together. I'll miss you

projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
```diff
@@ -118,6 +141,15 @@ public void Enqueue(IRpcContinuation k)
         ///</remarks>
         public void HandleChannelShutdown(ShutdownEventArgs reason)
         {
+            // No further frames will arrive on this channel, so the recorded entries can
+            // never be consumed: this queue belongs to a single Channel, and a channel
+            // being recovered gets a brand new one. Release them rather than keeping them
+            // alive for as long as something holds a reference to the dead channel.
+            // Note: ConcurrentQueue<T>.Clear() is not available on netstandard2.0.
+            while (_timedOutRpcs.TryDequeue(out _))
```
danielmarbach: Given the comment that no more frames can arrive and these records are no longer useful, could we clear this by atomically replacing the queue instead of draining it?

```
Interlocked.Exchange(
    ref _timedOutRpcs,
    new ConcurrentQueue<ushort>());
```

A concurrent operation could still enqueue into the detached queue, but that should be harmless during shutdown because those records no longer need to be observed. This would also avoid the conditional implementation needed for frameworks where `ConcurrentQueue<T>.Clear()` s unavailable.

projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
```diff
@@ -171,34 +203,104 @@ public void RpcCanceled(bool responseReceived, ReadOnlySpan<ProtocolCommandId> p
...
+            // rabbitmq/rabbitmq-dotnet-client#1964
+            // Keep the record bounded. The oldest entry is the one whose response is most
+            // overdue and so the least likely to still arrive, which makes it the right one
+            // to give up on. Dropping it costs only that this library no longer recognizes
+            // that particular late response, so it would surface as "Received unexpected
+            // command of type ...!" exactly as it did before #1802.
+            //
+            // Best effort: RpcCanceled is only reached from the RPC-issuing path, which is
+            // serialized on the channel's RPC semaphore, so at most one caller trims here.
+            while (_timedOutRpcs.Count > MaxTimedOutRpcs && _timedOutRpcs.TryDequeue(out _))
```
danielmarbach: ConcurrentQueue<T>.Count is not necessarily a cheap atomic field read. Once the queue spans multiple segments, obtaining the count may acquire the cross-segment lock and inspect segment positions.

For context, we previously had a regular queue here, but replaced it in #1939 with a single struct containing overlapping command-ID fields because the design assumption was that there could only ever be one timed-out RPC awaiting a late response. The main goal was to make that invariant explicit, rather than merely optimize the queue away.

This change necessarily revisits that assumption and introduces a bounded concurrent history instead. Since RPC cancellations still appear to be serialized by the channel's RPC semaphore and each call enqueues only one record, can this queue ever exceed the limit by more than one entry at this point? If not, could the "while" be an "if"?

That would avoid repeatedly using "Count" as the eviction mechanism and would document the new concurrency invariant more clearly. If multiple producers can in fact make it exceed the bound by more than one, it may be worth maintaining the size incrementally rather than repeatedly querying "ConcurrentQueue<T>.Count".

lukebakken: Fixed in 2653cd60b, the `while` is now an `if`. All 20 `RpcCanceled` call sites are inside the `_rpcSemaphore` block and each enqueues one entry, so the bound can only be exceeded by one.

Skipped the size counter since `Count` now runs once per timeout rather than in a loop. Say the word if you'd rather not lean on the semaphore invariant.

And yes, #1939's assumption is the bug: the semaphore serializes requests, not the broker's replies, so two timeouts in a row leave two responses in flight and the single slot dropped the first.

projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
lukebakken: `Matches` needs to stay a method, it's an overlap test rather than equality. Two entries sharing one command ID would compare equal, which isn't transitive and doesn't match `GetHashCode`.

`record struct` on its own does compile on netstandard2.0 (non-positional form only, positional needs `IsExternalInit`). Left it as `readonly struct` since it'd be the only record in the library, but no strong feelings if you'd prefer otherwise.

projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
lukebakken: Adopted in 78156b302, that's nicer. Replaces the drain loop and avoids needing `ConcurrentQueue<T>.Clear()`, which netstandard2.0 doesn't have. Field is no longer `readonly` as a result.

147/147 on net8.0 and net472.

projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
danielmarbach: Good for me.

Just fyi the externals init can be shimmed

danielmarbach:
Given the invariant doesn't hold is there a hot fix required for 6.x?

projects/RabbitMQ.Client/Impl/Channel.cs
```diff
@@ -980,11 +995,18 @@ await ModelSendAsync(in method, k.CancellationToken)
...
                     enqueued = Enqueue(k);
 
+                    // rabbitmq/rabbitmq-dotnet-client#1964
+                    // Send before the try whose catch records a timed-out RPC, so a write
+                    // failure (cancellation, IOException, AlreadyClosedException) does not
+                    // record an entry for a request that never reached the wire. Order-based
+                    // late-response matching would then desync permanently. Only a failure
+                    // while awaiting the response, by which point the frame is sent, should
+                    // record.
+                    await ModelSendAsync(in method, k.CancellationToken)
```
danielmarbach, line 1002: I don't get this. Couldn't we simply do a closer matching catch then the extra comment might not be necessary?

projects/RabbitMQ.Client/Impl/Channel.cs
```diff
@@ -1398,11 +1447,17 @@ await ModelSendAsync(in method, k.CancellationToken)
                 {
                     enqueued = Enqueue(k);
 
+                    // rabbitmq/rabbitmq-dotnet-client#1964
+                    // Send before the try whose catch records a timed-out RPC, so a
+                    // cancellation during the write does not record an entry for a request
+                    // that never reached the wire. Order-based late-response matching would
+                    // then desync permanently. Only a cancellation while awaiting the
+                    // response, by which point the frame is sent, should record.
+                    await ModelSendAsync(in method, k.CancellationToken)
```
danielmarbach, line 1451: Same comment as above

projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
```diff
@@ -69,25 +69,48 @@ public void Dispose()
...
+        // rabbitmq/rabbitmq-dotnet-client#1964
+        // Upper bound on recorded timed-out RPCs. Entries are only removed by an inbound
+        // command or by channel shutdown, so a channel that receives no frames at all (a
+        // publish-only channel with publisher confirmations disabled) would otherwise grow
+        // this without limit while RPCs keep timing out against an unresponsive broker.
+        //
+        // The bound is deliberately generous. Only one RPC runs per channel at a time, so
+        // with the default 20 second ContinuationTimeout a channel can record roughly three
+        // entries per minute, and reaching this many takes about three quarters of an hour
+        // of uninterrupted timeouts. A channel in that state has much larger problems.
```
danielmarbach, line 81: > A channel in that state has much larger problems.

This comment almost indicates there should be some form of circuit breaker... but I guess that is out of scope

projects/Test/Unit/TestAsyncRpcContinuation.cs
```diff
@@ -0,0 +1,142 @@
...
+        [Fact]
+        public async Task TestTimeoutDoesNotStartAtConstruction()
+        {
+            using var k = new SimpleAsyncRpcContinuation(ProtocolCommandId.ExchangeDeclareOk,
+                s_continuationTimeout, CancellationToken.None);
+
+            // Stand in for a long wait on the channel's RPC semaphore. This is more
+            // than twice s_continuationTimeout, so an armed-at-construction
+            // continuation would already be cancelled by now.
+            await Task.Delay(s_simulatedSemaphoreWait);
```
danielmarbach, line 67: Reminds me that in a modern code base we could pass time provider and then all this awaiting and slowing down test runs would not be necessary.

danielmarbach (dismissed):

projects/RabbitMQ.Client/Impl/Channel.cs
lukebakken: You're right, done. Eighteen of the twenty `RpcCanceled` sites already caught `OperationCanceledException`; this one and `BasicConsumeAsync` were the only bare catches, which is exactly why they needed the long comment.

Narrowing them also fixed something latent. Only a continuation timeout should record a late response, and `HandleContinuationTimeout` calls `TrySetCanceled` before `TrySetException`, so a timeout arrives as `TaskCanceledException`. The bare catch was additionally recording for `OperationInterruptedException` from channel shutdown, the unexpected-command `InvalidOperationException`, and - here specifically, since `AssertResultIsTrue(await k)` sits inside the guarded try - the `BugFound` `InvalidOperationException`. `RpcCanceled`'s `if (responseReceived) return;` guard meant nothing was actually recorded in that last case, so no live bug, but the catch was claiming something untrue.

Comment is down to three lines now.

projects/RabbitMQ.Client/Impl/Channel.cs
lukebakken: This one was already `catch (OperationCanceledException)`, so I read your comment as being about the comment length rather than the catch. Trimmed it to the same three lines as the others - the ordering constraint is worth a note, but it did not need six lines to say it.

projects/RabbitMQ.Client/Impl/RpcContinuationQueue.cs
lukebakken: Agreed on both counts - it does hint at one, and it is out of scope here. The bound is a leak guard, not a health signal; a channel producing that many timed-out RPCs is already failing in a way this queue cannot fix. Worth its own issue if we want to act on it.

projects/Test/Unit/TestAsyncRpcContinuation.cs
lukebakken: It would, yes. The blocker is that `TimeProvider` needs .NET 8 or a `Microsoft.Bcl.TimeProvider` reference, and netstandard2.0 is still a hard requirement here - so it is a dependency decision rather than a refactor, and not one to make inside a bugfix PR. Happy to open an issue for it.

lukebakken:
> Given the invariant doesn't hold is there a hot fix required for 6.x?

No - I checked 6.x and neither half of this bug can occur there.

The late-response tracking that #1939 introduced and this PR fixes is 7.x-only. 6.x has no `ShouldIgnoreCommand`, no `RpcCanceled`, and no `TimedOutRpc`; there is no order-based matching invariant to break.

The spurious-timeout half does not apply either. 6.x is synchronous: `ModelBase` takes `lock (_rpcLock)` and calls `k.GetReply(ContinuationTimeout)` *inside* the lock, so the timeout is armed after the lock is acquired and never charges queueing time against the operation's budget. It also means a timed-out RPC blocks its successors rather than racing them.

6.x is not clean here, just broken differently. A timeout throws a bare `TimeoutException` out of `BlockingCell.WaitForValue` and nothing resets `_outstandingRpc`, so the stale continuation stays put and the next `Enqueue` on that channel throws `NotSupportedException("Pipelining of requests forbidden")` - the `CompareExchange` against `s_tmp` fails. A late response is delivered to the timed-out continuation, whose `BlockingCell` nobody is waiting on any more, so it is silently dropped rather than misattributed.

So the channel is effectively dead after one RPC timeout, but there is no cross-request desync and no way to get one request's reply attributed to another. That is a usability problem in 6.x, not the correctness problem this PR fixes, and it is not worth a hotfix on a branch in extended support.

Also pushed the catch narrowing you asked for. Details in the two inline replies.

Issue #1976, opened 2026-07-30
Disposing a SemaphoreSlim without releasing it can permanently strand waiters in MainSession and Channel


lukebakken:
> [!NOTE]
> This issue was researched and written by Claude Code, prompted by @lukebakken while fixing #1968. The `SemaphoreSlim` disposal behaviour described below was verified empirically with a standalone probe, not inferred from documentation.
>
> **Corrected 2026-07-31** while implementing #1987: the original audit under Site 2 claimed `AutorecoveringConnection`'s two semaphores were never disposed. They were. Site 3 below is the correction, and the site and call counts have been updated with it.

## The pattern

`SemaphoreSlim` gives `Dispose()` no special meaning for pending async waiters. Disposing one while another task is parked in `WaitAsync` leaves that waiter **pending forever**: it does not fault, it does not cancel, and not even the waiter's own cancellation token or wait timeout releases it. `Dispose()` itself does not throw, so the disposing caller sees nothing wrong.

Verified directly with a standalone probe. With a 3s bound on the wait and the semaphore disposed while the waiter is parked, all three wait forms were still pending at 8s:

| wait form | state at 8s |
|---|---|
| `WaitAsync(token)` | pending |
| `WaitAsync(TimeSpan)` | pending |
| `WaitAsync(TimeSpan, token)` | pending |

The important consequence: **a timeout on the wait does not mitigate this.** Code that looks defensive because it passes `DefaultConnectionAbortTimeout` is just as exposed as code that waits unbounded.

#1968 is the confirmed instance of this pattern. `SocketFrameHandler.CloseAsync` acquired `_closingSemaphore` and disposed it in `finally` without ever releasing it. `MainLoop`'s `FinishCloseAsync` is itself a frame-handler closer, so when it lost that race it never returned, `_mainLoopTask` never completed, and `Connection.CloseAsync` burned the full 30s `DefaultConnectionCloseTimeout` before surfacing a bare `OperationCanceledException`.

Three more places dispose a `SemaphoreSlim` that other code paths wait on, across seven disposal calls. None has a known stranding today, but the hazard is identical and the failure mode is a permanent hang rather than an exception.

## Site 1: `MainSession._closingSemaphore`

`projects/RabbitMQ.Client/Impl/MainSession.cs`

[...]

PR #1977, opened 2026-07-30
Release the frame handler's closing semaphore instead of disposing it

### LINKED ISSUE #1968: Flaky test: TestCleanClosureWithSocketClosedOutOfBand times out on net472 with a bare OperationCanceledException
### What happened

`Test.Integration.TestConnectionShutdown.TestCleanClosureWithSocketClosedOutOfBand` fails intermittently on the `integration-win32` CI leg (net472) with an unexpected `OperationCanceledException` escaping `IConnection.CloseAsync`.

Observed on the `feat/add_activity_on_connection` branch (PR #1734), but the test and every frame in the stack are unmodified from `main` — that branch's only `Connection.cs` change is an activity-tag guard in `WriteAsync`, which is not on this path. So this is a pre-existing flake in `main`, not a regression from that PR.

```
Failed Test.Integration.TestConnectionShutdown.TestCleanClosureWithSocketClosedOutOfBand [30 s]
  Error Message:
   System.OperationCanceledException : The operation was canceled.
  Stack Trace:
     at RabbitMQ.Client.TaskExtensions.DoWaitAsync(...) in projects\RabbitMQ.Client\TaskExtensions.cs:line 121
   at RabbitMQ.Client.Impl.Connection.CloseAsync(...) in projects\RabbitMQ.Client\Impl\Connection.cs:line 464
   at RabbitMQ.Client.Impl.Connection.CloseAsync(...) in projects\RabbitMQ.Client\Impl\Connection.cs:line 480
   at RabbitMQ.Client.Impl.AutorecoveringConnection.CloseAsync(...) in projects\RabbitMQ.Client\Impl\AutorecoveringConnection.cs:line 259
   at RabbitMQ.Client.Impl.AutorecoveringConnection.CloseAsync(...) in projects\RabbitMQ.Client\Impl\AutorecoveringConnection.cs:line 274
   at Test.Integration.TestConnectionShutdown.TestCleanClosureWithSocketClosedOutOfBand() in projects\Test\Integration\TestConnectionShutdown.cs:line 69
```

### What the test does

It closes the frame handler out of band and *then* calls `CloseAsync`, i.e. it deliberately drives a clean close over an already-dead socket. It tolerates `AlreadyClosedException` (asserting an `IOException` inner) and `ChannelClosedException`, but not `OperationCanceledException`.

### Reading of the stack

This is a non-abort close (`abort: false`), so:

- `Connection.cs:464` is `await _mainLoopTask.WaitAsync(mainLoopWaitToken)`. For a non-abort close `mainLoopWaitToken` is `cts.Token` — the token linked to `timeoutCts`.
- `Connection.cs:480` is the `if (false == abort) { throw; }` in the surrounding `catch`, which rethrows after the best-effort `_frameHandler.CloseAsync`.
- `TaskExtensions.cs:121` is `DoWaitAsync`, which throws `new OperationCanceledException(cancellationToken)` when the token wins the `Task.WhenAny` race. It lives inside `#if !NET`, which is why this is net472-only — on `net8.0` the built-in `Task.WaitAsync` is used instead.

So `_mainLoopTask` did not complete before the close timeout expired, and the timeout surfaced as a bare `OperationCanceledException`.

The **30s duration is the tell**: the test passes `_waitSpan` = 6s, but `CloseAsync` floors non-abort timeouts at `InternalConstants.DefaultConnectionCloseTimeout` (30s):

```csharp
if (false == abort && timeout < InternalConstants.DefaultConnectionCloseTimeout)
{
    timeout = InternalConstants.DefaultConnectionCloseTimeout;
}
```

So the test waited the full 30s for the main loop and never got it. That points at `MainLoop` staying parked rather than at a merely-slow close — plausibly the same net472 behavior documented at `Connection.cs:443-461`: cancelling `_mainLoopCts` does not interrupt a `PipeReader.ReadAsync` already parked on the `NetworkStream`. That comment explains why the `abort` path closes the socket up front (issue #1921); the non-abort path has no equivalent, and here the socket is *already* closed out of band, so which side wins looks timing-dependent.

That last paragraph is a hypothesis, not a diagnosis — hence `S-needs-investigation`.

### Open questions

1. Is a bare `OperationCanceledException` even the right thing for a caller to see when a clean close times out? `AlreadyClosedException` / `ChannelClosedException` are what this test was written to expect, and the `TODO` already in the test suggests the exception surface here was never settled.
2. Should the non-abort path get the same net472 up-front socket close the abort path has, when the frame handler is already known to be closed?
3. Is the 30s floor on a caller-supplied 6s timeout intended? It makes a deliberately short close timeout ineffective, and it is what turns this flake into a 30-second CI stall.

### Notes

- net472 / `integration-win32` only so far; `integration-ubuntu` (net8.0) has been green across the runs I looked at.
- Frequency is unquantified — seen once while watching PR #1734. Worth a `--filter TestCleanClosureWithSocketClosedOutOfBand` loop on Windows to get a real rate before spending time on a fix.
- Related in spirit to the shutdown/cancellation work in #1921 and #1960; `docs/internal/connection-shutdown-and-cancellation.md` is the relevant background.

lukebakken:
> [!NOTE]
> The root-cause investigation, the fix, and this description were produced by Claude Code, prompted and reviewed by @lukebakken. The `SemaphoreSlim` disposal behaviour was verified empirically with a standalone probe rather than inferred from documentation.

Fixes #1968.

## The bug

`SocketFrameHandler.CloseAsync` acquired `_closingSemaphore` and, in its `finally`, disposed it without ever releasing it.

`SemaphoreSlim` gives `Dispose()` no special meaning for pending async waiters. A second `CloseAsync` already parked in `WaitAsync` when the first disposes the semaphore stays pending **forever**. Verified directly with a standalone probe, and worth internalising because it is counter-intuitive:

| wait form | state 8s after the semaphore is disposed |
|---|---|
| `WaitAsync(token)`, 3s CTS | pending |
| `WaitAsync(TimeSpan.FromSeconds(3))` | pending |
| `WaitAsync(3s, token)` | pending |

Not even the waiter's own cancellation token releases it, and a timeout on the wait is not an escape hatch. `Dispose()` itself does not throw, so the disposing closer sees nothing wrong.

## Why that reaches a 30-second stall

`MainLoop`'s `FinishCloseAsync` calls `_frameHandler.CloseAsync`, so it is itself one of the concurrent closers. When it lost this race it never returned, so:

1. `MainLoop` never completes, so `_mainLoopTask` never completes.
2. `Connection.CloseAsync` waits on `_mainLoopTask` (`Connection.cs:464`) and burns its full timeout.
3. That timeout is `InternalConstants.DefaultConnectionCloseTimeout` (30s), not the caller's value, because a non-abort close floors the requested timeout.

That is the exact shape of what #1968 recorded: `TestCleanClosureWithSocketClosedOutOfBand` failing after exactly 30s with a bare `OperationCanceledException`.

**The platform was a red herring.** #1968 was filed against net472 and I originally speculated in the issue that netstandard2.0's `PipeReader.ReadAsync` stayed parked. That was wrong. The TFM determines only the *exception type*: the throwing frame `TaskExtensions.DoWaitAsync` sits inside `#if !NET`, so netstandard2.0 surfaces a bare `OperationCanceledException` where net8.0 would use the built-in `Task.WaitAsync`. The hang itself reproduces on net8.0/Linux. The 30-second duration was the tell that it tracked the close timeout rather than a merely-slow close.

## The fix

Release the semaphore instead of disposing it, and set `_closed` before releasing, so the next waiter observes the close as done and returns rather than repeating it. A `WaitAsync` that faults now returns instead of proceeding, since it must not run the close body without holding the semaphore.

The semaphore is deliberately **not** disposed. `SemaphoreSlim` only requires disposal once `AvailableWaitHandle` has been read, and this one never exposes it, so there is nothing to reclaim.

`_closed` is now `volatile`. Both `CloseAsync` and `WriteAsync` read it outside any lock, on a field written by whichever thread happens to own the close, so those reads need acquire semantics rather than a value the JIT may keep in a register. The store paired with `Release()` is already ordered by the semaphore itself; the fast-path reads are not. `Connection._closed` (`Connection.cs:51`) is `volatile` for the same reason, so this matches existing convention. `Interlocked` would be overkill here, since nothing needs read-modify-write atomicity: only one thread can reach the store, because it holds the semaphore.

## Regression test

`TestConcurrentFrameHandlerCloseDoesNotHang_GH1968` closes the frame handler twice concurrently. That is the minimal deterministic form of the `MainLoop` race and needs no platform-specific timing. Discriminating power confirmed: with the fix reverted it fails at the 6s bar with a `TimeoutException`, the second task left in `WaitingForActivation`. With the fix it passes in ~140ms.

## Verification

On net8.0 against a local broker: Integration 194 passed, SequentialIntegration 64 passed, Unit 127 passed, no failures. `TestConnectionShutdown` ran 10 consecutive times at 12/12. Both target frameworks build clean with no new warnings.

## Not in this PR

The same dispose-without-release pattern exists in `MainSession` (`_closingSemaphore`) and `Channel` (`_rpcSemaphore`, `_confirmSemaphore`, four dispose sites). Neither is known to strand a caller today, but the hazard is identical. Tracked separately as #1976 rather than widened into this fix.

Background, the probe results, and the diagnostic reasoning are written up in `docs/internal/connection-shutdown-and-cancellation.md`.

### DIFF
--- a/projects/RabbitMQ.Client/Impl/SocketFrameHandler.cs
+++ b/projects/RabbitMQ.Client/Impl/SocketFrameHandler.cs
@@ -55,7 +55,10 @@ namespace RabbitMQ.Client.Impl
         private readonly PipeReader _pipeReader;
         private Task? _writerTask;
 
-        private bool _closed;
+        // volatile for the unsynchronized fast-path reads at the top of
+        // CloseAsync and WriteAsync. The read inside CloseAsync's critical
+        // section is already ordered by the semaphore.
+        private volatile bool _closed;
 
         private static ReadOnlyMemory<byte> Amqp091ProtocolHeader => new byte[] { (byte)'A', (byte)'M', (byte)'Q', (byte)'P', 0, 0, 9, 1 };
 
@@ -183,6 +186,26 @@ namespace RabbitMQ.Client.Impl
             {
                 await _closingSemaphore.WaitAsync(cancellationToken)
                     .ConfigureAwait(false);
+            }
+            catch
+            {
+                /*
+                 * Either the caller's token fired or another closer disposed the
+                 * semaphore. Both mean this call does not own the close, and the
+                 * owner sets _closed. Returning is correct: it must NOT fall
+                 * through to the close body without holding the semaphore.
+                 */
+                return;
+            }
+
+            try
+            {
+                if (_closed)
+                {
+                    // A concurrent closer finished while we waited.
+                    return;
+                }
+
                 try
                 {
                     // TryComplete rather than Complete: WriteLoopAsync may have
@@ -214,17 +237,32 @@ namespace RabbitMQ.Client.Impl
                 {
                     // ignore, we are closing anyway
                 }
-            }
-            catch
-            {
-            }
-            finally
-            {
-                _closingSemaphore.Dispose();
+
 #if NET
                 _flushTimeoutCts?.Dispose();
 #endif
+            }
+            finally
+            {
+                /*
+                 * Set _closed before releasing so the next waiter sees the close
+                 * as done and returns rather than repeating it.
+                 *
+                 * The semaphore is deliberately NOT disposed. Disposing it while a
+                 * concurrent CloseAsync is parked in WaitAsync leaves that waiter
+                 * pending forever -- not even its own cancellation token releases
+                 * it, which was verified directly. MainLoop's FinishCloseAsync is
+                 * one such concurrent closer, so a frame-handler close racing it
+                 * stranded MainLoop, _mainLoopTask never completed, and
+                 * Connection.CloseAsync burned its full 30s timeout before
+                 * surfacing a bare OperationCanceledException. See issue #1968.
+                 *
+                 * SemaphoreSlim only needs disposal when AvailableWaitHandle has
+                 * been read, and this one never exposes it, so there is nothing
+                 * to reclaim.
+                 */
                 _closed = true;
+                _closingSemaphore.Release();
             }
         }
 
--- a/docs/internal/connection-shutdown-and-cancellation.md
+++ b/docs/internal/connection-shutdown-and-cancellation.md
@@ -232,6 +232,45 @@ Two gates, in `projects/Applications/GH-1960/`:
 
 The token state in the app's output is the discriminating signal: `cancellable,not-yet` on a `Library/541` reason is the bug; `cancellable,already-cancelled` is the fix.
 
+## Issue #1968: a disposed `SemaphoreSlim` strands the second frame-handler closer
+
+This one presented as a flaky net472 test - `TestCleanClosureWithSocketClosedOutOfBand` failing after exactly 30s with a bare `OperationCanceledException` - and was filed as `S-needs-investigation`. It is a library bug, and the platform and the test were both red herrings.
+
+### The mechanism
+
+`SocketFrameHandler.CloseAsync` acquired `_closingSemaphore` and, in its `finally`, **disposed** it without ever releasing it. `SemaphoreSlim` grants disposal no special meaning for pending async waiters: a second `CloseAsync` already parked in `WaitAsync` when the first disposes the semaphore stays pending **forever**.
+
+Verified directly, and worth internalising because it is counter-intuitive - all three wait forms hang, so a timeout is not an escape hatch:
+
+| Wait form | Outcome after the semaphore is disposed |
+|---|---|
+| `WaitAsync(token)` with a 3s CTS | still pending at 8s |
+| `WaitAsync(TimeSpan.FromSeconds(3))` | still pending at 8s |
+| `WaitAsync(3s, token)` | still pending at 8s |
+
+Not even the waiter's own cancellation token releases it. `Dispose()` itself does not throw, so the first closer sees nothing wrong.
+
+### Why that reaches a user-visible 30s stall
+
+`MainLoop`'s `FinishCloseAsync` calls `_frameHandler.CloseAsync`, so it is itself one of the concurrent closers. When it lost this race it never returned, which means:
+
+1. `MainLoop` never completes, so `_mainLoopTask` never completes.
+2. `Connection.CloseAsync` waits on `_mainLoopTask` (`Connection.cs:464`) and burns its full timeout.
+3. That timeout is `InternalConstants.DefaultConnectionCloseTimeout` (30s), **not** the caller's 6s - a non-abort close floors the caller's value at 30s.
+4. On netstandard2.0 the throwing frame is `TaskExtensions.DoWaitAsync`, which is inside `#if !NET`, so it surfaces as a bare `OperationCanceledException`.
+
+Item 4 is the only platform-specific part, and it affects the *exception type*, not the hang. The 30-second duration was the tell that this tracked the close timeout rather than a merely-slow close.
+
+### The fix
+
+Release the semaphore instead of disposing it, and set `_closed` before releasing so the next waiter returns rather than repeating the close. A `WaitAsync` that faults now returns instead of proceeding - it must not run the close body without holding the semaphore. `SemaphoreSlim` only requires disposal once `AvailableWaitHandle` has been read, and this one never exposes it, so nothing leaks.
+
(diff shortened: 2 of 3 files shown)


projects/RabbitMQ.Client/Impl/SocketFrameHandler.cs
```diff
@@ -214,17 +237,32 @@ await _pipeReader.CompleteAsync()
                 {
                     // ignore, we are closing anyway
                 }
-            }
-            catch
-            {
-            }
-            finally
-            {
-                _closingSemaphore.Dispose();
```
danielmarbach, line 223: FYI: Normally when you have analyzers enabled, it would raise this

```
[SuppressMessage(
    "Reliability",
    "CA2213:Disposable fields should be disposed",
```

so given all of this is internal you could also go extra safe and make socket frame handler disposable which then gets carried into the connection and disposed there. But yeah when the wait handle is not accessed no need to dispose.

lukebakken: Thanks - I went and checked both halves of this, and the answer is a bit more interesting than "analyzers are off".

**The analyzer would not raise CA2213 here, and not because it is disabled.** `EnableNETAnalyzers` is not set in this repo, but I forced the full set on to see what it would say:

```
dotnet build projects/RabbitMQ.Client/RabbitMQ.Client.csproj \
  -p:EnableNETAnalyzers=true -p:AnalysisMode=All -p:AnalysisLevel=latest
```

CA2213 fires 8 times, all of them on `PollingCounter` fields in `RabbitMqClientEventSource`. Nothing on `_closingSemaphore`. The rule only applies to types that are themselves `IDisposable`, and `SocketFrameHandler` is not - so the field is invisible to it either way.

What is worth knowing is what happens if we take your suggestion. I made `SocketFrameHandler : IFrameHandler, IDisposable` with an empty `Dispose()` and rebuilt with the same flags:

```
SocketFrameHandler.cs(47,37):  CA2213: field '_socket' ... never disposed
SocketFrameHandler.cs(48,33):  CA2213: field '_stream' ... never disposed
SocketFrameHandler.cs(52,40):  CA2213: field '_closingSemaphore' ... never disposed
SocketFrameHandler.cs(148,42): CA2213: field '_flushTimeoutCts' ... never disposed
```

So making it disposable is what *creates* the CA2213 obligation, on four fields rather than one. Three of those are deliberate today: `_socket` is closed via `_socket.Close()` in `CloseAsync` rather than disposed, `_stream` is owned by the socket, and `_flushTimeoutCts` is disposed inside `CloseAsync` under `#if NET` because it is reset and reused per flush. Adopting the pattern means either wiring all four into a real `Dispose` or suppressing the rule on the ones we do not want to change, and it would have to be threaded through `IFrameHandler` (the only implementer, but `Connection` holds the interface) plus `Connection.DisposeAsync`. That is a bigger change than the bug warrants, for a semaphore that has nothing to reclaim.

Which is your own last sentence, and it is the part that settles it: `SemaphoreSlim`'s finalizable state is the lazily-created `AvailableWaitHandle`. Grepping the whole repo, nothing reads that property on any semaphore, so `Dispose()` on this one is a no-op over unmanaged resources. That is why releasing rather than disposing is not just safer but complete.

The reason it has to be a release and not just a dropped `Dispose()`, for the record: a disposed `SemaphoreSlim` does not fault or cancel already-pending async waiters. I verified all three wait forms - `WaitAsync(token)` with a 3s CTS, `WaitAsync(3s)`, and `WaitAsync(3s, token)` - and every one was still pending at 8s. Not even the waiter's own token releases it. Since `MainLoop`'s `FinishCloseAsync` is itself one of the concurrent closers, losing that race stranded `MainLoop`, `_mainLoopTask` never completed, and `Connection.CloseAsync` burned the full 30s `DefaultConnectionCloseTimeout`. That is the 30-second net472 failure in #1968.

I did keep the safety property you are pointing at, just without the disposable plumbing: the `catch` around the wait returns instead of falling through, so a call that does not own the close can never run the close body unsynchronized.

Worth noting the same anti-pattern is still live in `MainSession.Dispose` (`_closingSemaphore`) and `Channel` (`_rpcSemaphore`, `_confirmSemaphore`), which do dispose semaphores that other paths wait on. Not known to strand anyone today, but the hazard is identical and a timeout on the wait does not mitigate it. Tracked as #1976 - if you think the disposable-frame-handler approach is the right shape for those, that is the better place to apply it, since `Channel` is already `IAsyncDisposable`.

PR #1983, opened 2026-07-30
feat: Publish ReadOnlySequence

PauloHMattos:
> [!NOTE]
> This PR was mostly written by Claude Opus 5 with occasional input from me.
> The PR description with the changes summary were written just by Claude.

## Why

Message bodies are frequently produced in pieces: `PipeReader.ReadAsync` hands back a `ReadOnlySequence<byte>`, serializers write into chained buffers, and pooled allocators hand out fixed-size blocks. To publish such a body today the caller must gather it into one contiguous `byte[]`/`IMemoryOwner<byte>` first, which costs an extra full-body copy and an allocation on every publish.

That copy is not required by the protocol. AMQP 0-9-1 already frames a body as a *series* of body frames, so the wire format never needed contiguity - only the client API did.

This adds `ReadOnlySequence<byte>` publish overloads that write the segments straight to the wire, following the shape of the `ReadOnlyMemory<byte>` + `IDisposable bodyOwner` overloads added in #1922 and reusing their memory-ownership semantics.

```csharp
// body arrives as a multi-segment sequence, e.g. from a Pipe
ReadResult read = await pipeReader.ReadAsync();
await channel.BasicPublishAsync(exchange, routingKey, read.Buffer, bodyOwner: owner);
```

### Public API (additive only, 7 new members)

Two `IChannel` overloads, mirroring the existing memory + `bodyOwner` pair:

```csharp
ValueTask BasicPublishAsync<TProperties>(string exchange, string routingKey,
    bool mandatory, TProperties basicProperties, ReadOnlySequence<byte> body,
    IDisposable? bodyOwner, CancellationToken cancellationToken = default)
    where TProperties : IReadOnlyBasicProperties, IAmqpHeader;

// ...and the CachedString variant
```

Plus the five matching `IChannelExtensions` conveniences (`PublicationAddress`, and `string`/`CachedString` with and without `mandatory`). Entries were added to both `PublicAPI.Unshipped.net8.0.txt` and `PublicAPI.Unshipped.netstandard2.0.txt`. No existing signature changed.

**Ownership contract** (documented on every new overload, identical to the memory overloads): ownership of `bodyOwner` transfers to the client, which disposes it **exactly once** - after the message reaches the wire, or when publication fails. Every segment must stay valid and unmodified until then, and the caller must not reuse, mutate, or release that memory itself. Passing `null` is valid and selects the copy path.

### Internals

`OutgoingFrame._body` changed from `ReadOnlyMemory<byte>` to `ReadOnlySequence<byte>`, and the publish path (`ISession.TransmitAsync` → `Channel.ModelSendAsync` → `BasicPublishCoreAsync`) was unified on that one representation rather than duplicated. All of it is `internal`.

The existing ReadOnlyMemory path retains a dedicated TransmitAsync/SerializeToFrames overload and incurs no wrapping cost.

For the ReadOnlySequence path, every hot spot short-circuits on `IsSingleSegment`: `Framing.SerializeToFrames` delegates single-segment sequences to the existing `ReadOnlyMemory` implementation, and `OutgoingFrame.WriteTo` keeps its original span loop. Multi-segment bodies use a new `BodySegment.WriteTo(IBufferWriter<byte>, …)` that writes the 7-byte frame header, each segment, then the end marker independently - so it never asks the `PipeWriter` for a span covering more than one segment. Body frame boundaries are sliced out of the sequence independently of segment boundaries.

Both branches are preserved for sequences: with no `bodyOwner` the body is copied into a single pooled buffer (the caller still owns it); with a `bodyOwner` only the method + header are pooled and the segments go straight out, then the owner is disposed.

### Struct size

`OutgoingFrame._body` changed from `ReadOnlyMemory<byte>` (16 bytes) to `ReadOnlySequence<byte>` (24 bytes), widening the struct from 48 to 56 bytes. This also widens the bounded 128-entry `Channel<OutgoingFrame>` queue at `SocketFrameHandler.cs` by 1 KB total.

### Validation

`body.Length` is a `long`, while AMQP content headers, the tracing APIs, and the pooled frame buffer are all `int`-based. Publishing now rejects bodies that cannot be framed, disposing the owner first:

- `ArgumentOutOfRangeException` when `body.Length > int.MaxValue`, thrown at the public entry point before any publisher-confirm sequence number is consumed.
- `ArgumentOutOfRangeException` when the *total frame set* (method + header + body + per-frame overhead) would exceed `int.MaxValue`, checked against the negotiated `MaxPayloadSize`. `SessionBase.TransmitAsync`'s existing `bytes.Size == 0` branch already disposes the owner on this path.

## Incidental fixes

Both are in `Framing`, both were reachable before this PR, and both are covered by new tests:

1. **`int` overflow in the frame-count math.** `(length + maxPayloadBytes - 1) / maxPayloadBytes` overflows for a body near `int.MaxValue`, producing a negative count and a bogus buffer size. Rewritten as `((length - 1) / maxPayloadBytes) + 1`, which cannot overflow.
2. **Empty body with unlimited frame size.** `GetBodyFrameCount` returned `1` for a zero-length body when `maxPayloadBytes == int.MaxValue` (i.e. a negotiated `frame_max` of 0). The frame set was sized 8 bytes too large while the write loop emitted no body frame, so the client sent 8 bytes of uninitialized pooled memory and tripped the `offset == size` assertion in debug builds. It now returns `0`.

## Benchmarks

`projects/Benchmarks/WireFormatting/MethodFraming.cs` gained single-segment and multi-segment `ReadOnlySequence` variants (`BasicPublishWriteSingleSegmentSequence`, `BasicPublishWriteMultiSegmentSequence`, and their `WithOwner` counterparts).

What the code paths guarantee, independent of measured numbers:

- The `ReadOnlyMemory` overloads keep their own `SerializeToFrames`/`TransmitAsync` path and are never routed through a `ReadOnlySequence`, so the pre-existing publish paths take on no wrapping cost from this change.
- A single-segment `ReadOnlySequence` short-circuits on `IsSingleSegment` and delegates to that same `ReadOnlyMemory` implementation, so it runs the memory path rather than the multi-segment one.
- A multi-segment body with a `bodyOwner` writes its segments straight to the wire instead of gathering them into one contiguous buffer, trading the gather-copy for independent per-segment frame writes; with no `bodyOwner` it falls back to a single gather-copy that the caller still owns.

Numeric results are intentionally omitted for now: the earlier table predates the split of the `ReadOnlyMemory` and `ReadOnlySequence` paths, and I would rather add fresh figures from a clean benchmark run than quote stale ones.

## Testing

New unit tests (no broker required):

- `TestSegmentedBodyFraming` - a multi-segment body produces **byte-identical** output to its contiguous equivalent, across `maxBodyPayloadBytes` smaller than, equal to, and larger than the segment size, plus ragged lengths and `int.MaxValue`; correct body-frame count, payload limits, and end markers; `Size` equals bytes actually written; empty sequences, all-empty-segment sequences, and empty segments interleaved with data.
- `TestSerializeToFramesWithSequence` - copy path and zero-copy path emit identical bytes; single-segment sequences match the memory overload; owner disposed exactly once by `frame.Dispose()`; both oversize guards reject without leaking the owner (using a `MemoryManager<byte>`-backed sequence that reports a huge length without allocating).

New integration tests in `TestBasicPublishAsync` - round trip via the `IChannel`, extension, and `CachedString` overloads; a body spanning multiple body frames *and* multiple segments; confirms disabled; channel already closed; pre-cancelled token; `mandatory: true` producing a `basic.return`; oversize rejection. Each asserts the owner was disposed **exactly once** (waiting for the write loop rather than racing it), and the three pre-existing memory tests were upgraded to the same assertion.


PauloHMattos:
**Question for reviewers:** Should a no-owner `ReadOnlySequence<byte>` overload be added for symmetry with the plain
`ReadOnlyMemory<byte>` API? Right now callers who don't need ownership transfer pass
`bodyOwner: null`, which works and takes the copy path, but there's no `bodyOwner`-free signature as
there is for memory bodies. I left it out to keep the added surface minimal and happy to add it.

PauloHMattos:
@lukebakken finally got time to contribute this change, but I saw you are putting the work to release 7.3.0 and I don't want to add to the burden, so if you think it's better we can leave this for 7.4.0

cc @danielmarbach I think you might be interested in this

projects/RabbitMQ.Client/Impl/Frame.cs
```diff
@@ -239,15 +286,143 @@ public static OutgoingFrame SerializeToFrames<TMethod, THeader>(ref TMethod meth
...
+                {
+                    return (int)totalSize;
+                }
+            }
+
+            ThrowBodyTooLarge(framingSize, bodyLength, maxBodyPayloadBytes);
+            return 0; // unreachable
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
```
danielmarbach: Would this benefit from DoesNotReturn?

danielmarbach (commented):
I need to look into it with a fresh mind. Did a quick squim only for now

danielmarbach:
> ~9 ns more, from wrapping the ReadOnlyMemory into a ReadOnlySequence inside OutgoingFrame; 

I guess having a dual path in there would also add overhead right? And if so do we know how much? So should this be seen as the cost of supporting this additional machinery or how do you see it?

lukebakken:
@PauloHMattos hello! I added this to `7.3.0` as I expect the OpenTelemetry work to take a while.

projects/RabbitMQ.Client/Impl/Frame.cs
```diff
@@ -239,15 +286,143 @@ public static OutgoingFrame SerializeToFrames<TMethod, THeader>(ref TMethod meth
...
+
+            int framingSize = Method.FrameSize + Header.FrameSize +
+                              method.GetRequiredBufferSize() + header.GetRequiredBufferSize();
+            int totalSize = GetTotalFrameSetSize(framingSize, body.Length, maxBodyPayloadBytes);
+            int bodyLength = (int)body.Length;
+
+            if (bodyOwner is null)
+            {
+                // Will be returned by SocketFrameWriter.WriteLoop
+                byte[] buffer = ArrayPool<byte>.Shared.Rent(totalSize);
```
PauloHMattos, line 317: Should we create a new ReadOnlySequence here to prevent renting a LOH sized array?

projects/RabbitMQ.Client/Impl/Frame.cs
```diff
@@ -239,15 +286,143 @@ public static OutgoingFrame SerializeToFrames<TMethod, THeader>(ref TMethod meth
...
+        /// <summary>
+        /// Serializes a method, header and a (potentially non-contiguous) body into a set of AMQP frames.
+        /// </summary>
+        /// <remarks>
+        /// When <paramref name="bodyOwner"/> is <c>null</c> the body is copied into a single pooled
+        /// buffer, because the caller retains ownership of the memory backing the sequence. Otherwise
+        /// ownership is transferred to the returned <see cref="OutgoingFrame"/>, which writes the
+        /// segments straight to the wire and disposes <paramref name="bodyOwner"/> afterwards.
+        /// </remarks>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
gfoidl: This method is quite huge for inlining.
Split it into two: one for the `IsSingleSegment` path (with aggressive inlining) and the other one for the multi segment path w/o inlining.

projects/RabbitMQ.Client/Impl/Frame.cs
gfoidl: The arrays in the array pool will probably land in Gen-2 anyway.
LOH is conceptionally also Gen-2 so there shouldn't be a problem (especially as value types (`byte`) and no GC references are stored in the `buffer`).

projects/RabbitMQ.Client/Impl/Frame.cs
```diff
@@ -239,15 +286,143 @@ public static OutgoingFrame SerializeToFrames<TMethod, THeader>(ref TMethod meth
...
+                }
+
+                System.Diagnostics.Debug.Assert(offset == totalSize, $"Serialized to wrong size, expect {totalSize}, offset {offset}");
+                return new OutgoingFrame(buffer, totalSize);
+            }
+            else
+            {
+                // Rent a smaller buffer exclusively for the Method and Header
+                byte[] buffer = ArrayPool<byte>.Shared.Rent(framingSize);
+                Span<byte> bufferSpan = buffer.AsSpan();
```
gfoidl: Is a stack-alloc buffer meaningful here (w/ fallback to the array pool when more than e.g. 256 bytes are needed)?

projects/RabbitMQ.Client/Impl/Frame.cs
gfoidl: The `NoInlining` should be dropped here.

With the attribute the JIT sees it should not bothter with that method, just emit a call.

W/o the attribute the JIT sees that this method is small[^1], so eligible for inlining, and examines the body. There it sees the method is never returning (i.e. throwing), thus treat it specially (i.e. this method won't be inlined and put into a rather cold segment of the code layout as exceptional path).
[^1]: among other things to consider for inlining

`DoesNotReturn` could / should be applied as suggested.
It kind of documents the behavior in method meta-data, and in the future tooling (compilers, etc) could take benefit of it -- which we get for free then.

paulomorgado: Methods that throw exceptions are not inlineable, anyway,

I also like to make clear when it comes to these throw methods.

A `[DoesNotReturn]` may help the compiler's flow analysis. I would add to every one of these methods.

I'd also add a `[StackTraceHidden]` to not pollute the stack trace, in case of an exception.

paulomorgado:
> **Question for reviewers:** Should a no-owner `ReadOnlySequence<byte>` overload be added for symmetry with the plain `ReadOnlyMemory<byte>` API? Right now callers who don't need ownership transfer pass `bodyOwner: null`, which works and takes the copy path, but there's no `bodyOwner`-free signature as there is for memory bodies. I left it out to keep the added surface minimal and happy to add it.

The `using` statement does not require that the returned disposable is not `null`.

So, it's better to have a `bodyOwner` parameter that may be `null`. And make that the argument's default value.

projects/RabbitMQ.Client/Impl/Frame.cs
paulomorgado: Agreed!

And rented memory. should be released by the renter.

projects/RabbitMQ.Client/Impl/Frame.cs
paulomorgado: Unless it's proven to be beneficial, I would let the JIT decide.

projects/RabbitMQ.Client/Impl/Frame.cs
danielmarbach: But then you probably also want skip locals init

projects/RabbitMQ.Client/Impl/Frame.cs
gfoidl: What should the JIT decide when it's hinted to be force-inlined?
Either split the method (happy- / unhappy-path, which is common practice) or remove the aggressive-inline attribute.

Put it into other words: unless proven that force-inlining is beneficial let the JIT decide (I think this is what you meant, or?).

danielmarbach: My understanding is with modern jit it is more beneficial to do less forced aggressive inlining and my interpretation was that's what @paulomorgado meant

projects/RabbitMQ.Client/Impl/Frame.cs
```diff
@@ -134,6 +150,41 @@ public static int WriteTo(IBufferWriter<byte> bufferWriter, ushort channel, Read
...
+            {
+                System.Diagnostics.Debug.Assert(body.Length <= int.MaxValue - BaseFrameSize);
+                int payloadLength = (int)body.Length;
+
+                /*
+                 * The 7 byte frame header is written with a single 64 bit store (the 8th byte
+                 * belongs to the payload), so request one extra byte and commit only the 7 bytes
+                 * that make up the header.
+                 */
+                Span<byte> headerSpan = bufferWriter.GetSpan(StartPayload + 1);
```
lukebakken, line 171: This `+ 1` is load-bearing and correct, but nothing pins it.

I changed it to `GetSpan(StartPayload)` and all 186 unit and 14 integration tests still passed. So I checked why it matters: `NetworkOrderSerializer.WriteUInt64` does `Unsafe.WriteUnaligned` on a raw `ref byte` from `MemoryMarshal.GetReference`, with no bounds check. I wrote a strictly-conforming `IBufferWriter` whose `GetSpan(n)` returns exactly `n` bytes - legal, since the contract only promises *at least* `n` - and the 8-byte store into a 7-byte span succeeds silently, writing past the end.

So the invariant holds today only because `PipeWriter.Create(stream)` happens to over-deliver spans. A different writer, or a `PipeOptions` change, turns this into silent heap corruption on the publish path, and the test suite would stay green.

Could you add a test with an exact-size `IBufferWriter` that fails if the `+ 1` is removed? Your comment above already explains the why; this would keep someone from "simplifying" it in two years.

projects/RabbitMQ.Client/OutgoingFrame.cs
```diff
@@ -40,7 +40,7 @@ internal struct OutgoingFrame : IDisposable
     {
         private byte[]? _methodAndHeaderArray;
         private readonly int _methodAndHeaderLength;
-        private ReadOnlyMemory<byte> _body;
+        private ReadOnlySequence<byte> _body;
```
lukebakken, line 43: Worth a line in the PR description: this widens `OutgoingFrame` from 48 to 56 bytes (`ReadOnlySequence<byte>` is 24 vs `ReadOnlyMemory<byte>`'s 16 - measured with `Unsafe.SizeOf`), which also widens the bounded 128-entry `Channel<OutgoingFrame>` at `SocketFrameHandler.cs:72`.

Not arguing against the unification, just that the struct-size change is the kind of thing a reader wants stated rather than discovered.

projects/RabbitMQ.Client/Impl/Channel.BasicPublish.cs
```diff
@@ -51,7 +51,7 @@ public ValueTask BasicPublishAsync<TProperties>(string exchange, string routingK
             where TProperties : IReadOnlyBasicProperties, IAmqpHeader
         {
             var cmd = new BasicPublish(exchange, routingKey, mandatory, default);
-            return BasicPublishCoreAsync(cmd, basicProperties, body, bodyOwner: null, exchange, routingKey, cancellationToken);
+            return BasicPublishCoreAsync(cmd, basicProperties, new ReadOnlySequence<byte>(body), bodyOwner: null, exchange, routingKey, cancellationToken);
```
lukebakken: The description says "to keep the existing path free of new cost," but all four pre-existing `ReadOnlyMemory` overloads now wrap in `new ReadOnlySequence<byte>(body)` (here and lines 63, 72, 81), so every existing publish pays something. Your own benchmark table is honest about it (~9ns on the owner path, attributed to exactly this), so it is really the prose that oversells rather than the numbers.

One gap in the measurement: `MethodFraming` benchmarks `SerializeToFrames`, not the `BasicPublishAsync` entry point where the wrap actually happens. Could you re-measure at the entry point, and either soften that sentence or keep a `ReadOnlyMemory` overload of `BasicPublishCoreAsync`/`ModelSendAsync` to avoid the round trip?

projects/RabbitMQ.Client/Impl/Channel.BasicPublish.cs
```diff
@@ -69,20 +69,56 @@ public ValueTask BasicPublishAsync<TProperties>(CachedString exchange, CachedStr
...
             return BasicPublishCoreAsync(cmd, basicProperties, body, bodyOwner, exchange.Value, routingKey.Value, cancellationToken);
         }
 
+        /// <summary>
+        /// A message body must be addressable with an <see cref="int"/>, both because that is the
+        /// limit of a single AMQP content header and because the set of frames used to send it is
+        /// allocated as one contiguous block. Ownership of <paramref name="bodyOwner"/> has already
+        /// been transferred by the caller, so it has to be disposed here when the body is rejected.
+        /// </summary>
+        private static void ValidateBodyLength(in ReadOnlySequence<byte> body, IDisposable? bodyOwner)
```
lukebakken, line 112: Minor: this duplicates a gate that `Framing.GetTotalFrameSetSize` already applies, with the same exception type and `paramName`. The two differ in behavior, though - this one throws synchronously and disposes `bodyOwner` itself, while the `GetTotalFrameSetSize` path throws inside the async state machine and relies on `TransmitAsync`'s catch. So a body of `int.MaxValue - 4` fails differently from one of `int.MaxValue + 1`.

I verified dispose-exactly-once on both, so this is not a leak - just a split surface that is easy to regress. Your XML doc explains why the check is here; consider noting that the stricter total-frame-set check lives downstream.

projects/Benchmarks/WireFormatting/MethodFraming.cs
```diff
@@ -1,4 +1,4 @@
-﻿// This source code is dual-licensed under the Apache License, version
+// This source code is dual-licensed under the Apache License, version
```
lukebakken: The UTF-8 BOM was stripped from this file. It is now the only non-generated file in `projects/Benchmarks/` without one - the other 9 all have it. `.editorconfig` sets no `charset`, so nothing enforces either way; probably worth restoring to keep the diff clean.

projects/Benchmarks/Program.cs
```diff
@@ -19,7 +19,7 @@ class Config : ManualConfig
     {
         public Config()
         {
-            AddJob(Job.Default.WithRuntime(CoreRuntime.Core31));
+            AddJob(Job.Default.WithRuntime(CoreRuntime.Core80));
```
lukebakken: `CoreRuntime.Core31` to `Core80` is unrelated to the feature and unmentioned in the description. It is a reasonable fix on its own (netcoreapp3.1 is long EOL), but note `Benchmarks.csproj` targets `net8.0;net472` while this config requests `Core80` plus `ClrRuntime.Net48`, so the `net472` build has no matching job. Better as its own commit.

lukebakken (commented):
> [!NOTE]
> Written by Claude (Anthropic's Claude Code), directed and approved by @lukebakken. The findings were verified against 5f9e25f336955f5487b500cc5a1fff34d77aca2e by build, test runs, and mutation testing.

Nice work. I could not find a correctness bug, and I tried reasonably hard to: beyond reading the diff I ran two mutation tests to check whether the new tests actually have teeth.

What I verified locally at 5f9e25f336955f5487b500cc5a1fff34d77aca2e: client builds 0 warnings / 0 errors with `-warnaserror`, with both `net8.0` and `netstandard2.0` assemblies actually produced. Unit 186/186, `TestBasicPublishAsync` 14/14. I could not build `net472` here (no targeting pack), so CI's `integration-win32` and `sequential-integration-win32` are what cover that TFM - both green.

The dispose contract is genuinely pinned. I replaced the guarded `if (!bodyOwnerTransferred)` in the `finally` with an unconditional `bodyOwner?.Dispose()` and 9 of 14 tests failed. `DisposalTracker` is the reason: `Interlocked.Increment` plus a 250ms window after the first disposal to catch a second one. That is a real improvement over the `Assert.True(body.Disposed)` on a plain `bool` that `main` had, which could not detect a double-dispose at all. The ownership handoff reads correctly too - the flag is set immediately before `ModelSendAsync`, and `TransmitAsync` discriminates on `bytes.Size == 0` for the never-captured case.

The one thing I would like fixed before merge is an untested invariant in the multi-segment writer - inline below. Everything else is minor.

Also confirmed: the empty-body fix is real (on `main`, `GetBodyFrameCount(int.MaxValue, 0)` returns 1 because the `maxPayloadBytes == int.MaxValue` early-return precedes any length check, and `Connection.Commands.cs:180` sets `MaxPayloadSize = int.MaxValue` whenever the broker negotiates `frame_max = 0`), and it is pinned by `EmptyBodyWithUnlimitedFrameSizeProducesNoBodyFrames`. Your "Incidental fixes" write-up matches what I found in the code.

projects/RabbitMQ.Client/Impl/Channel.BasicPublish.cs
paulomorgado: I haven't looked at all the code, but I would prefer code that special cases a single item `ReadOnlySequence<byte>` as a `ReadOnlyMemory<byte>` than the other way around. A new `ReadOnlySequence<byte>` is always heap allocated.

danielmarbach: That's what I tried to ask here https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/1983#issuecomment-5135781910

projects/Benchmarks/Program.cs
PauloHMattos: This was already in its own separate commit, but I will split it into a new PR

projects/RabbitMQ.Client/Impl/Frame.cs
PauloHMattos: Keep this fix and the test in this PR or open another to not mix it from the new feature?

projects/RabbitMQ.Client/OutgoingFrame.cs
PauloHMattos: Added to the PR description

projects/RabbitMQ.Client/Impl/Frame.cs
PauloHMattos: Done. Replaced [NoInlining] with [DoesNotReturn] + [StackTraceHidden] (the latter #if NET-guarded since it's .NET 6+ only). Per @gfoidl's explanation, the JIT already won't inline a throwing method, so NoInlining was redundant — but [DoesNotReturn] documents intent for flow analysis and future tooling

projects/RabbitMQ.Client/Impl/Frame.cs
PauloHMattos: Done. Removed [AggressiveInlining] from both SerializeToFrames(sequence) and GetTotalFrameSetSize

projects/RabbitMQ.Client/Impl/Frame.cs
PauloHMattos: Can't use stackalloc here. The buffer is stored in OutgoingFrame and written asynchronously by SocketFrameWriter.WriteLoop, so it must outlive the stack frame

PauloHMattos:
Thank you all for the thorough review! It's genuinely a pleasure working with this level of collaboration.
I just pushed a few commits that I hope address all the concerns for now. I just wasn't able to re run the benchmarks yet, but will do later today and update the results in the PR body.

Looking forward to the next round of reviews

projects/RabbitMQ.Client/Impl/Channel.BasicPublish.cs
```diff
@@ -81,6 +81,43 @@ public ValueTask BasicPublishAsync<TProperties>(CachedString exchange, CachedStr
             return BasicPublishCoreAsync(cmd, basicProperties, body, bodyOwner, exchange.Value, routingKey.Value, cancellationToken);
         }
 
+        public ValueTask BasicPublishAsync<TProperties>(string exchange, string routingKey,
+            bool mandatory, TProperties basicProperties, ReadOnlySequence<byte> body, IDisposable? bodyOwner,
```
gfoidl:
```suggestion
            bool mandatory, TProperties basicProperties, in ReadOnlySequence<byte> body, IDisposable? bodyOwner,
```
`ReadOnlySequence` is quite big, so shouldn't be passed as copy, rather as "reference" for perf-reasons.

Same on the ther other places. Some of them are already with `in`, whilst others (like t his one) not.

---

If passing via `in` shouldn't be possible for any reason, so please add a short comment on why.

projects/RabbitMQ.Client/Impl/Frame.cs
```diff
@@ -239,15 +287,141 @@ public static OutgoingFrame SerializeToFrames<TMethod, THeader>(ref TMethod meth
...
+                }
+
+                System.Diagnostics.Debug.Assert(offset == totalSize, $"Serialized to wrong size, expect {totalSize}, offset {offset}");
+                return new OutgoingFrame(buffer, totalSize);
+            }
+            else
+            {
+                // Rent a smaller buffer exclusively for the Method and Header
+                byte[] buffer = ArrayPool<byte>.Shared.Rent(framingSize);
+                Span<byte> bufferSpan = buffer.AsSpan();
```
gfoidl:
```suggestion
                Span<byte> bufferSpan = buffer.AsSpan(0, framingSize);
```
To add a bit of extra safety in case that anywhere a read goes beyond `framingSize`.
W/o garbage from the pooled array would be read, this way an IndexOutOfRange-ex is thrown.

Perf cost is negligible.

projects/RabbitMQ.Client/Impl/Frame.cs
```diff
@@ -239,15 +287,141 @@ public static OutgoingFrame SerializeToFrames<TMethod, THeader>(ref TMethod meth
...
+            }
+
+            ThrowBodyTooLarge(framingSize, bodyLength, maxBodyPayloadBytes);
+            return 0; // unreachable
+        }
+
+        [DoesNotReturn]
+#if NET
+        [System.Diagnostics.StackTraceHidden]
+#endif
```
gfoidl, line 383: Instead of #ifdef one could "polyfill" that attribute.
I.e. just create it in a fiel with the exact same namespace, and conditionally compile that file.
So the code here become more straight.

But this is more a matter of taste.

danielmarbach: I agreed it should be polyfilled. We had a "discussion" about this see https://github.com/rabbitmq/rabbitmq-dotnet-client/issues/1688 and https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/1687#issuecomment-2377963224 but did decide to not yet take a dependency to a polyfill lib, so it would have to polyfill by copying although Polysharp would be super lightweight compile time only dependency, which might be worthwhile bringing in

PauloHMattos: I will leave as is for now since and we can make a broader discussion about polyfilling

projects/RabbitMQ.Client/Impl/Frame.cs
lukebakken: Applied in ec4c674eb. This turned out to be a correctness fix, not just extra safety: with `buffer.AsSpan(framingSize)` the method and header were written at the wrong offset and `OutgoingFrame` transmitted the uninitialized head of the pooled buffer. The new `TestSerializeToFramesWithSequence` cases fail without this and pass with `buffer.AsSpan(0, framingSize)`.

projects/RabbitMQ.Client/Impl/Frame.cs
lukebakken: Added `[DoesNotReturn]` (with `[StackTraceHidden]`) to the throw helper in 4bf33d0dd.

projects/RabbitMQ.Client/Impl/Frame.cs
lukebakken: Done in c93e5c04e and 4bf33d0dd: the sequence overload delegates the single-segment case to the aggressively inlined `ReadOnlyMemory` overload and keeps the multi-segment path out of line.

projects/RabbitMQ.Client/Impl/Channel.BasicPublish.cs
lukebakken: Changed to `in ReadOnlySequence<byte>` across the synchronous overloads in 3c3de557c. The async `BasicPublishCoreAsync` core cannot take `in` (it is an async method), which is noted in a comment there.

projects/RabbitMQ.Client/Impl/Channel.BasicPublish.cs
lukebakken: Fixed in c93e5c04e: the four `ReadOnlyMemory` overloads no longer wrap in `new ReadOnlySequence<byte>(body)`. They call a dedicated `ReadOnlyMemory` core, so the existing path no longer pays the sequence-wrapping cost.

projects/Benchmarks/WireFormatting/MethodFraming.cs
lukebakken: Restored the UTF-8 BOM in 9bf7ffa6d.

projects/RabbitMQ.Client/Impl/Frame.cs
lukebakken: Good catch. Fixed in 8d58a4ab0: `ExactSizeBufferWriter` now hands back exactly the requested size backed by a sentinel guard region and verifies it on Advance/flush. Removing the `+ 1` makes the 8-byte header store overrun the guard, so `BodySegmentWriteToWithExactSizeBufferWriter` throws; with it in place the suite stays green. Confirmed by removing the `+ 1` locally and watching only that test fail.

projects/RabbitMQ.Client/Impl/Channel.BasicPublish.cs
lukebakken: Kept intentional and documented in 1c358287a. The synchronous check rejects an oversize body before a publisher confirmation sequence number is consumed and disposes the owner itself, which the downstream `GetTotalFrameSetSize` check (inside the async state machine) does not do. Added a line to the summary spelling that out.

projects/RabbitMQ.Client/Impl/Frame.cs
lukebakken: The 🧞 and I re-opened this issue to continue this discussion - https://github.com/rabbitmq/rabbitmq-dotnet-client/issues/1688

lukebakken:
> [!NOTE]
> Drafted by Claude (Anthropic's Claude Code) under @lukebakken's direction; he reviewed it before posting. The CS0121 ambiguity was verified against the branch.

@PauloHMattos @paulomorgado on the two options for the no-owner case:

Defaulting `bodyOwner` to `null` cannot be done on the `ReadOnlyMemory<byte>` overloads. They coexist with the already-shipped no-owner overloads (`..., ReadOnlyMemory<byte> body, CancellationToken`), so adding `bodyOwner = null` makes a call like `BasicPublishAsync(exchange, routingKey, mandatory, props, body)` ambiguous (CS0121) between the two candidates. Both would take the same copy path anyway, so there is nothing to gain.

The default does compile on the `ReadOnlySequence<byte>` overloads, and a separate no-owner sequence overload would also work, but I would rather do neither. `bodyOwner` is the zero-copy switch: passing `null` copies the whole body into a pooled buffer, which is exactly the cost a caller reaching for `ReadOnlySequence<byte>` is usually trying to avoid. Keeping it required forces that choice to be explicit at each call site instead of silently falling back to a copy.

I added a paragraph to the sequence overload docs in 4cd60e951 making it clear that a null owner means the body is copied.

lukebakken:
> [!NOTE]
> Drafted by Claude (Anthropic's Claude Code) under @lukebakken's direction; he reviewed it before posting. The described code paths were verified against the branch.

@danielmarbach on the wrapping overhead: after the core split, the existing no-owner `ReadOnlyMemory<byte>` path no longer wraps into a `ReadOnlySequence<byte>`. It copies straight into a pooled buffer as before, so existing publishers pay nothing new. The wrap only remains on the new owner-carrying path (non-null `bodyOwner`), which is opt-in new surface, so there is no regression for current callers.

The original ~9 ns figure came from `MethodFraming`, which benchmarks `SerializeToFrames` rather than the `BasicPublishAsync` entry point where the wrap actually happens, so it is not measuring the real entry-point cost. The remaining owner-path wrap is the cost of the zero-copy machinery rather than something worth pushing down into `OutgoingFrame`. @PauloHMattos planned to re-run the benchmarks; entry-point numbers for the owner path can go in the PR body then if a hard figure is useful.

lukebakken (dismissed):
Thank you @PauloHMattos. Once I get one more 👍 I'll merge. I'll clean up history a bit but will ensure you have attribution.

gfoidl (dismissed):
Great work 👍🏻

projects/RabbitMQ.Client/Impl/Frame.cs
```diff
@@ -239,15 +287,141 @@ public static OutgoingFrame SerializeToFrames<TMethod, THeader>(ref TMethod meth
...
+
+            int framingSize = Method.FrameSize + Header.FrameSize +
+                              method.GetRequiredBufferSize() + header.GetRequiredBufferSize();
+            int totalSize = GetTotalFrameSetSize(framingSize, body.Length, maxBodyPayloadBytes);
+            int bodyLength = (int)body.Length;
+
+            if (bodyOwner is null)
+            {
+                // Will be returned by SocketFrameWriter.WriteLoop
+                byte[] buffer = ArrayPool<byte>.Shared.Rent(totalSize);
```
danielmarbach, line 317: Do we assume here that the multiple writes can never throw and therefore we don't want to blow up the stack frames with a safety catch and return?

projects/RabbitMQ.Client/Impl/Frame.cs
```diff
@@ -239,15 +287,141 @@ public static OutgoingFrame SerializeToFrames<TMethod, THeader>(ref TMethod meth
...
+                    remainingBody = remainingBody.Slice(payloadSize);
+                }
+
+                System.Diagnostics.Debug.Assert(offset == totalSize, $"Serialized to wrong size, expect {totalSize}, offset {offset}");
+                return new OutgoingFrame(buffer, totalSize);
+            }
+            else
+            {
+                // Rent a smaller buffer exclusively for the Method and Header
+                byte[] buffer = ArrayPool<byte>.Shared.Rent(framingSize);
```
danielmarbach, line 337: Same as before. I know this is largely the same as we had already before and I was also involved in the code, but it only occurred to me now seeing the bigger machinery for ROS

projects/RabbitMQ.Client/IChannelExtensions.cs
```diff
@@ -313,6 +313,136 @@ public static ValueTask BasicPublishAsync(this IChannel channel,
...
+        /// <param name="body">The message body, which may be spread over several segments.</param>
+        /// <param name="bodyOwner">An optional <see cref="IDisposable"/> instance responsible for releasing or returning the memory used by the message body once publication is complete. Pass <c>null</c> if the body memory does not require disposal.</param>
+        /// <param name="cancellationToken">CancellationToken for this operation.</param>
+        /// <remarks>
+        /// Routing key must be shorter than 255 bytes.
+        /// The publication occurs with mandatory=false and empty BasicProperties.
+        /// Throws <see cref="Exceptions.PublishException"/> if a nack or basic.return is returned for the message.
+        /// Ownership of <paramref name="bodyOwner"/> is transferred to the client, which disposes it exactly once.
+        /// </remarks>
+        public static ValueTask BasicPublishAsync(this IChannel channel,
```
danielmarbach, line 409: Do we really want to expose `in` as part of the public API here? While it works today because this method is just a non-async forwarding method returning a `ValueTask`, it constrains future implementation changes. If we ever need to make this method itself `async`, we'd have to preserve a synchronous shim because `async` methods cannot have `in` parameters.

It also makes removing `in` later an API compatibility concern, while callers get relatively little benefit from it unless we've measured that avoiding the `ReadOnlySequence<byte>` struct copy matters on this path. I'd lean toward passing the sequence by value unless we have benchmark data showing that `in` is worthwhile here.

paulomorgado:
> Note
> 
> Drafted by Claude (Anthropic's Claude Code) under @lukebakken's direction; he reviewed it before posting. The CS0121 ambiguity was verified against the branch.
> 
> @PauloHMattos @paulomorgado on the two options for the no-owner case:
> 
> Defaulting `bodyOwner` to `null` cannot be done on the `ReadOnlyMemory<byte>` overloads. They coexist with the already-shipped no-owner overloads (`..., ReadOnlyMemory<byte> body, CancellationToken`), so adding `bodyOwner = null` makes a call like `BasicPublishAsync(exchange, routingKey, mandatory, props, body)` ambiguous (CS0121) between the two candidates. Both would take the same copy path anyway, so there is nothing to gain.

We could make the overload without `bodyOwner` [obsolete](https://learn.microsoft.com/en-us/dotnet/api/system.obsoleteattribute) and set its [priority](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-13.0/overload-resolution-priority). It would be binary compatible and source friendly. I 8.x, the overload without `bodyOwner` should be removed.

> The default does compile on the `ReadOnlySequence<byte>` overloads, and a separate no-owner sequence overload would also work, but I would rather do neither. `bodyOwner` is the zero-copy switch: passing `null` copies the whole body into a pooled buffer, which is exactly the cost a caller reaching for `ReadOnlySequence<byte>` is usually trying to avoid. Keeping it required forces that choice to be explicit at each call site instead of silently falling back to a copy.
> 
> I added a paragraph to the sequence overload docs in [4cd60e9](https://github.com/rabbitmq/rabbitmq-dotnet-client/commit/4cd60e9517542caa7968b936037710e7a520a345) making it clear that a null owner means the body is copied.

I don't think copying the `ReadOnlySequence<byte> body` if `bodyOwner` is `null` is a good idea. It's up to the caller to decide it. Doing this would hurt callers that have a throw away `ReadOnlySequence<byte> body`

danielmarbach:
> We could make the overload without `bodyOwner` [obsolete](https://learn.microsoft.com/en-us/dotnet/api/system.obsoleteattribute) and set its [priority](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-13.0/overload-resolution-priority). It would be binary compatible and source friendly. I 8.x, the overload without `bodyOwner` should be removed.

In my books, obsoleting interface members is a little weird because during a minor upgrade as an implementer of that interface, for example, for, let's say faking, mocking purposes and whatnot, I have no way to deal with the observations because I still have to implement it. So I think there is little value in obsoleting a member in an interface. I think it's better to add the overloads, do the overload resolution priority, and then in the next major drop the interface method entirely without actually adding obsoletes. These are my two cents, though.

paulomorgado:
> > We could make the overload without `bodyOwner` [obsolete](https://learn.microsoft.com/en-us/dotnet/api/system.obsoleteattribute) and set its [priority](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-13.0/overload-resolution-priority). It would be binary compatible and source friendly. I 8.x, the overload without `bodyOwner` should be removed.
> 
> In my books, obsoleting interface members is a little weird because during a minor upgrade as an implementer of that interface, for example, for, let's say faking, mocking purposes and whatnot, I have no way to deal with the observations because I still have to implement it. So I think there is little value in obsoleting a member in an interface. I think it's better to add the overloads, do the overload resolution priority, and then in the next major drop the interface method entirely without actually adding obsoletes. These are my two cents, though.

Default interface methods?

I'm all on board with keeping binary compatibility as much as possible. But crippling the API because of language features, no.

We need to keep users of the current major working, but plan for obsolence.

danielmarbach:
On board with that. My point was more about the fact that an obsoletion with warn should express a meaningful thing a consumer or an inheritor of an interface can do. If that's not possible the obsolete may as well simply not exist and the method can go in the next major. Tracking it for the maintainers is a different angle

projects/RabbitMQ.Client/Impl/Frame.cs
lukebakken: > [!NOTE]
> Drafted by Claude, reviewed by @lukebakken.

These writes only throw if the frame set was mis-sized, which is a programming error rather than a runtime condition: `ValidateBodyLength` and `GetTotalFrameSetSize` both reject a body that can't be addressed with an `int` before we rent, so the buffer is always large enough for the writes that follow. This mirrors the existing `ReadOnlyMemory` overload, which has no catch either. A try/catch would only return an already-rented buffer to the pool on a path that's already broken (a leaked pooled array is collected, not lost), so I'd rather keep the two overloads consistent and leave it out.

projects/RabbitMQ.Client/Impl/Frame.cs
lukebakken: > [!NOTE]
> Drafted by Claude, reviewed by @lukebakken.

Same reasoning as the `null`-owner branch above - the writes only throw on a mis-sized frame set, never on validated input. One wrinkle specific to this branch: such a throw would also skip disposing `bodyOwner` (it isn't yet handed to the `OutgoingFrame`, and `BasicPublishCoreAsync` has already set `bodyOwnerTransferred`). Since that's unreachable on valid input and matches the existing memory overload, I'm keeping it consistent rather than hardening one path alone.

projects/RabbitMQ.Client/IChannelExtensions.cs
lukebakken: > [!NOTE]
> Drafted by Claude, reviewed by @lukebakken.

Agreed - dropped `in` here. The benefit was smaller than it looks: the async core (`BasicPublishCoreAsync`) has to take the body by value anyway (async methods can't have `in` parameters), so the `ReadOnlySequence<byte>` is copied at the sync-to-async boundary regardless. `in` on the public surface saved at most one 24-byte copy before that unavoidable one, at the cost of baking the by-reference form into the public API (and blocking a future async forwarder without a synchronous shim).

Applied consistently everywhere this PR introduced `in` for the sequence body - the interface and extension overloads, their implementations, the internal transmit/framing plumbing, both PublicAPI baselines, and the tests - in bab89a62c. The rationale is documented at the `IChannel` overload and in the note in `Channel.BasicPublish.cs`.

cc @gfoidl

lukebakken (approved):
Approving. The three open threads from the 2026-08-13 review (the two `Frame.cs` rent/write questions and the `in`-on-public-API question) are addressed and resolved, and CI is green on bab89a62c across build, integration, sequential-integration, oauth2, and CodeQL.

lukebakken (approved):

PR #1994, opened 2026-08-04
Retry recovery when a topology operation times out

### LINKED ISSUE #1993: Topology recovery silently loses the consumer when `basic.consume` exceeds `ContinuationTimeout`
### Describe the bug

If the `basic.consume` issued by **topology recovery** does not complete within `ContinuationTimeout`
(default 20s), the client ends up with a **phantom consumer**:

* the broker *does* eventually process the `basic.consume` and registers an active consumer;
* the client discards the late `basic.consume-ok`, so the consumer is **never added to the new
  channel's `ConsumerDispatcher`**;
* recovery nevertheless reports success — `RecoverySucceededAsync` fires and
  `"Connection recovery completed"` is logged;
* every subsequent `basic.deliver` resolves to the internal `FallbackConsumer`, which logs at
  **`Info`** level and silently drops the message.

The connection and channel stay `IsOpen == true`. No exception is thrown, no event is raised. With
manual acks the delivery is never acked, so with `BasicQos(prefetchCount: 1)` the broker never
delivers again and the queue grows without bound. **The application is permanently wedged and has no
way to detect it.** With `autoAck: true` the messages are silently lost instead.

This is easy to hit in production with **quorum queues**: while a quorum queue has no leader, `basic.consume` legitimately takes longer than 20s to complete.

Not a duplicate of #1930 — that one is about `BasicPublishAsync` hanging, and its fix does not affect
this. Verified below.

### Reproduction steps

Two ways. The first is deterministic, needs a single node, and takes ~60s.

#### A. Minimal deterministic repro (single node, no cluster)

Self-contained program below. It embeds a tiny TCP proxy so it can (a) sever the connection to force
auto-recovery and (b) hold the broker's reply to topology recovery's `basic.consume` for 25s — i.e.
longer than `ContinuationTimeout`. That artificially produces the same condition a leaderless quorum
queue produces naturally.

```
dotnet new console -n GhIssueRepro && cd GhIssueRepro
dotnet add package RabbitMQ.Client --version 7.2.1
# replace Program.cs with the file attached to this issue
dotnet run                       # expects a broker on 127.0.0.1:5672, guest/guest
```

<details>
<summary>Program.cs</summary>

```csharp
// Minimal, self-contained reproduction for:
//   "Topology recovery silently loses the consumer when basic.consume exceeds ContinuationTimeout"
//
// Needs ONE RabbitMQ node on 127.0.0.1:5672 (guest/guest). No cluster required.
//   dotnet run
//
// The program embeds a tiny TCP proxy so it can (a) sever the connection to force auto-recovery
// and (b) make the broker's reply to the basic.consume issued by *topology recovery* arrive later
// than ContinuationTimeout. In a real 3-node cluster that delay happens naturally: after quorum
// loss a quorum queue has no leader, so basic.consume does not complete until a leader is elected.

using System.Diagnostics.Tracing;
using System.Net;
using System.Net.Sockets;
using System.Text;
using RabbitMQ.Client;
using RabbitMQ.Client.Events;

const int ProxyPort = 5690;
int BrokerPort = int.TryParse(Environment.GetEnvironmentVariable("BROKER_PORT"), out var bp) ? bp : 5672;
string User = Environment.GetEnvironmentVariable("AMQP_USER") ?? "guest";
string Pass = Environment.GetEnvironmentVariable("AMQP_PASS") ?? "guest";
// Must exceed ConnectionFactory.ContinuationTimeout (default 20s).
var StallForConsumeOk = TimeSpan.FromSeconds(25);

void Log(string m) => Console.WriteLine($"{DateTime.Now:HH:mm:ss.fff} {m}");

// ---------------------------------------------------------------- client EventSource capture
using var esListener = new EsListener();

// ---------------------------------------------------------------------------- stalling proxy
var proxy = new StallProxy(ProxyPort, BrokerPort, StallForConsumeOk, Log);
proxy.Start();

// ------------------------------------------------------------------------------------ client
var factory = new ConnectionFactory
{
    Uri = new Uri($"amqp://{User}:{Pass}@127.0.0.1:{ProxyPort}/"),
    AutomaticRecoveryEnabled = true,
    TopologyRecoveryEnabled = true,
    ClientProvidedName = "gh-issue-repro",
};

const string QueueName = "gh-issue-repro-qq";
var queueArgs = new Dictionary<string, object?> { ["x-queue-type"] = "quorum" };

var conn = await factory.CreateConnectionAsync();
var channel = await conn.CreateChannelAsync();

await channel.QueueDeclareAsync(QueueName, durable: true, exclusive: false, autoDelete: false, arguments: queueArgs);
await channel.QueuePurgeAsync(QueueName);
await channel.BasicQosAsync(0, prefetchCount: 1, global: false);

int delivered = 0;
var consumer = new AsyncEventingBasicConsumer(channel);
consumer.ReceivedAsync += async (_, ea) =>
{
    Interlocked.Increment(ref delivered);
    Log($"  APP received delivery tag={ea.DeliveryTag} redelivered={ea.Redelivered} -> ack");
    await channel.BasicAckAsync(ea.DeliveryTag, multiple: false);
};
var tag = await channel.BasicConsumeAsync(QueueName, autoAck: false, consumer);
Log($"consuming, consumerTag={tag}");

conn.RecoverySucceededAsync += (_, _) => { Log("EVENT RecoverySucceededAsync"); return Task.CompletedTask; };
conn.ConnectionRecoveryErrorAsync += (_, e) => { Log($"EVENT ConnectionRecoveryErrorAsync {e.Exception.GetType().Name}"); return Task.CompletedTask; };
consumer.RegisteredAsync += (_, e) => { Log($"EVENT consumer RegisteredAsync [{string.Join(",", e.ConsumerTags)}]"); return Task.CompletedTask; };

async Task PublishAsync(int n, string why)
{
    Log($"publishing {n} message(s) ({why})");
    for (int i = 0; i < n; i++)
    {
        await channel.BasicPublishAsync("", QueueName, mandatory: false,
            basicProperties: new BasicProperties { Persistent = true },
            body: Encoding.UTF8.GetBytes($"msg-{why}-{i}"));
    }
}

await PublishAsync(3, "before");
await Task.Delay(2000);
Log($"--- baseline OK: delivered={Volatile.Read(ref delivered)} (expected 3) ---");

// ------------------------------------------------------------- force recovery + stall the RPC
Log("");
Log(">>> severing the TCP connection; the reply to topology recovery's basic.consume will be");
Log($">>> delayed by {StallForConsumeOk.TotalSeconds:F0}s (i.e. longer than the 20s ContinuationTimeout)");
proxy.StallNextConnectionsConsumeOk = true;
proxy.DropAllConnections();

// Recovery retries every 5s; the stalled basic.consume then burns 20s before timing out.
await Task.Delay(TimeSpan.FromSeconds(45));

int before = Volatile.Read(ref delivered);
Log("");
Log($">>> recovery finished. delivered so far = {before}. Publishing 3 more messages...");
await PublishAsync(3, "after");
await Task.Delay(TimeSpan.FromSeconds(15));

int after = Volatile.Read(ref delivered);
Log("");
Log($"=== RESULT: delivered before={before} after={after} (expected after = before + 3) ===");
Log(after == before
    ? "=== BUG REPRODUCED: connection and channel are open, the broker shows an active consumer\n" +
      "===  and an unacked message, but the application never receives another delivery.\n" +
      "===  Look above for: 'Will not retry recovery because of ...TaskCanceledException'\n" +
      "===  followed by 'Unhandled HandleBasicDeliverAsync for tag ...' (the FallbackConsumer).\n" +
      "===  Verify with: rabbitmqctl list_queues name messages messages_unacknowledged consumers"
    : "=== not reproduced this run ===");
Log($"channel.IsOpen={channel.IsOpen} connection.IsOpen={conn.IsOpen}");

// ============================================================================================
sealed class EsListener : EventListener
{
    protected override void OnEventSourceCreated(EventSource s)
    {
        if (s.Name == "rabbitmq-client") EnableEvents(s, EventLevel.LogAlways, EventKeywords.All);
    }

    protected override void OnEventWritten(EventWrittenEventArgs e)
    {
        if (e.EventName == "EventCounters") return;
        var first = e.Payload is { Count: > 0 } ? e.Payload[0]?.ToString() : "";
        Console.WriteLine($"{DateTime.Now:HH:mm:ss.fff}   ESLOG[{e.Level}] {first}");
    }
}

/// <summary>
/// Forwards 127.0.0.1:listenPort -> 127.0.0.1:brokerPort. Can sever all live connections, and can
/// hold the broker->client direction for a while once it sees the client send a basic.consume.
/// </summary>
sealed class StallProxy(int listenPort, int brokerPort, TimeSpan stall, Action<string> log)
{
    private readonly List<Socket> _live = new();
    private readonly object _gate = new();

    public volatile bool StallNextConnectionsConsumeOk;

    public void Start()
    {
        var srv = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
        srv.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.ReuseAddress, true);
        srv.Bind(new IPEndPoint(IPAddress.Loopback, listenPort));
        srv.Listen(16);
        log($"proxy listening on 127.0.0.1:{listenPort} -> 127.0.0.1:{brokerPort}");
        _ = Task.Run(async () =>
        {
            while (true)
            {
                var client = await srv.AcceptAsync();
                _ = Task.Run(() => HandleAsync(client));
            }
        });
    }

    public void DropAllConnections()
    {
        lock (_gate)
        {
            foreach (var s in _live)
            {
                try { s.Close(); } catch { }
            }
            _live.Clear();
        }
    }

    private async Task HandleAsync(Socket client)
    {
        var upstream = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
        await upstream.ConnectAsync(new IPEndPoint(IPAddress.Loopback, brokerPort));
        lock (_gate) { _live.Add(client); _live.Add(upstream); }

        bool armed = StallNextConnectionsConsumeOk;
        // DateTime.MinValue == "no stall in effect"; set to a future instant once armed and triggered.
        var releaseAt = DateTime.MinValue;
        bool triggered = false;

        // client -> broker: watch for basic.consume (class 60, method 20)
        var up = Task.Run(async () =>
        {
            var buf = new byte[65536];
            try
            {
                while (true)
                {
                    int n = await client.ReceiveAsync(buf);
                    if (n <= 0) break;
                    if (armed && !triggered && ContainsBasicConsume(buf, n))
                    {
                        triggered = true;
                        releaseAt = DateTime.UtcNow + stall;
                        log($"  [proxy] saw basic.consume from client -> holding broker replies for {stall.TotalSeconds:F0}s");
                    }
                    await upstream.SendAsync(buf.AsMemory(0, n));
                }
            }
            catch { }
            finally { Close(client, upstream); }
        });

        // broker -> client: withhold once armed
        var down = Task.Run(async () =>
        {
            var buf = new byte[65536];
            try
            {
                while (true)
                {
                    int n = await upstream.ReceiveAsync(buf);
                    if (n <= 0) break;
                    while (DateTime.UtcNow < releaseAt) await Task.Delay(200);
                    await client.SendAsync(buf.AsMemory(0, n));
                }
            }
            catch { }
            finally { Close(client, upstream); }
        });

        await Task.WhenAny(up, down);
    }

    // AMQP frame: type(1) channel(2) size(4) then payload starting with classId(2) methodId(2).
    // basic.consume == class 60 (0x003C), method 20 (0x0014).
    private static bool ContainsBasicConsume(byte[] b, int len)
    {
        for (int i = 0; i + 11 < len; i++)
        {
            if (b[i] == 1 /* method frame */ &&
                b[i + 7] == 0x00 && b[i + 8] == 0x3C &&
                b[i + 9] == 0x00 && b[i + 10] == 0x14)
            {
                return true;
            }
        }
        return false;
    }

    private static void Close(Socket a, Socket b)
    {
        try { a.Close(); } catch { }
        try { b.Close(); } catch { }
    }
}
```
</details>

Output (RabbitMQ 4.2.0, client 7.2.1):

```
13:14:54.268 consuming, consumerTag=amq.ctag-Bt_KTh6cP4sy619NbGJcOg
13:14:54.278   APP received delivery tag=1 redelivered=False -> ack
13:14:54.283   APP received delivery tag=2 redelivered=False -> ack
13:14:54.285   APP received delivery tag=3 redelivered=False -> ack
13:14:56.274 --- baseline OK: delivered=3 (expected 3) ---

>>> severing the TCP connection; the reply to topology recovery's basic.consume will be
>>> delayed by 25s (i.e. longer than the 20s ContinuationTimeout)
13:14:56.296   ESLOG[Error] unexpected connection closure: ... Connection reset by peer.
13:15:01.291   ESLOG[Informational] Performing automatic recovery
13:15:01.349   [proxy] saw basic.consume from client -> holding broker replies for 25s
13:15:21.357   ESLOG[Error] Topology recovery exception                        <-- +20.0s
13:15:21.357   ESLOG[Informational] Will not retry recovery because of System.Threading.Tasks.TaskCanceledException: it's not a known problem with connectivity, ignoring it
13:15:21.358   ESLOG[Informational] Connection recovery completed
13:15:21.358 EVENT RecoverySucceededAsync                                      <-- reports success

>>> recovery finished. delivered so far = 3. Publishing 3 more messages...
13:15:41.295   ESLOG[Informational] Unhandled HandleBasicDeliverAsync for tag amq.ctag-Bt_KTh6cP4sy619NbGJcOg
13:15:56.282 === RESULT: delivered before=3 after=3 (expected after = before + 3) ===
13:15:56.282 channel.IsOpen=True connection.IsOpen=True
```

Note `consumer.RegisteredAsync` never fires after recovery, and only **one** "Unhandled
HandleBasicDeliverAsync" appears because `prefetchCount: 1` means the broker waits forever for the
ack of the message it already dropped.

#### B. Real-world scenario that led us here (3-node cluster, quorum queues)

1. 3-node RabbitMQ 4.2.0 cluster (deployed with the Cluster Operator on Kubernetes).
2. An RPC-style app: one connection consuming from a quorum queue with `BasicQos(0, 1, false)`,
   manual ack, `AutomaticRecoveryEnabled = true`, `TopologyRecoveryEnabled = true`. Steady load
   ~10 msg/s.
3. Kill node A (not hosting the consumer's connection). Quorum retained; the app keeps working.
4. **Within ~2s**, kill node B, the one hosting the consumer's connection. Only 1 of 3 nodes remains,
   so quorum is lost. (Timing matters: if you wait long enough for the first pod to be rescheduled,
   quorum is never actually lost and the bug does not trigger.)
5. Let the nodes come back. Quorum is restored and the queues return to `running`.

Observed: the client reconnects, `RecoverySucceededAsync` fires, but the application never receives
another delivery. `rabbitmqctl` shows the consumer is alive on the *new* connection and holding one
unacked message:

```
name                 messages  messages_ready  messages_unacknowledged  consumers
ClusterTEST_Server   914       913             1                        1
```

Client log for that window:

```
12:49:38.210 Info:  Performing automatic recovery
12:49:38.224 Warn:  Connection delegate was manually closed. Aborted recovery.
12:49:43.224 Info:  Performing automatic recovery
12:50:03.282 Error: Topology recovery exception                      <-- +20.058s = ContinuationTimeout
12:50:03.282 Info:  Will not retry recovery because of System.Threading.Tasks.TaskCanceledException: it's not a known problem with connectivity, ignoring it
12:50:03.291 Info:  Connection recovery completed
12:50:05.827 Info:  Unhandled HandleBasicDeliverAsync for tag amq.ctag-gA_sHk3jDiYaxk157JQGnA
```

Reflecting into the recovered channel while wedged, next to a healthy channel in the same process:

```
stuck : IsOpen=True ActiveDeliveryTagOffset=400 MaxSeenDeliveryTag=1 dispatcherQuiescing=0
        registeredConsumerTags=[]                              consumer.IsRunning=False
healthy: IsOpen=True ActiveDeliveryTagOffset=0  MaxSeenDeliveryTag=400 dispatcherQuiescing=0
        registeredConsumerTags=[amq.ctag-X4EigbhQGbWVeVtChD4AtQ] consumer.IsRunning=True
```

`MaxSeenDeliveryTag=1` proves the channel processed a `basic.deliver`; `registeredConsumerTags=[]`
proves there was no consumer to hand it to.

### Expected behavior

A `basic.consume` that fails or times out during topology recovery must not leave the client in a
state where the broker has an active consumer that the client cannot service. Either:

* the recovery attempt should fail and be retried (so the consumer is genuinely re-established), or
* the client should tear the channel/connection down again, or
* at minimum the condition must be observable — a `basic.deliver` for an unregistered consumer tag is
  a protocol-level inconsistency and should not be swallowed at `Info` level.

Right now `RecoverySucceededAsync` firing is actively misleading.

### Analysis

Line numbers are for `v7.2.1`; the same code is on `main` at c64cc15.

1. `AutorecoveringConnection.Recovery.cs` → `RecoverConsumersAsync` calls
   `consumer.RecoverAsync(channelToUse)` → `IChannel.BasicConsumeAsync(...)`.
2. The RPC continuation times out after `ContinuationTimeout`.
   `AsyncRpcContinuation.HandleContinuationTimeout` does `tcs.TrySetCanceled(cancellationToken)`,
   so the awaiter throws **`TaskCanceledException`**, not `TimeoutException`.
3. `Channel.BasicConsumeAsync`'s `catch` calls
   `_continuationQueue.RpcCanceled(k.HandledProtocolCommandIds)` (the #1802 mechanism) and rethrows.
4. `RecoverConsumersAsync` catches it and calls `HandleTopologyRecoveryException`
   (`AutorecoveringConnection.Recovery.cs:157` in 7.2.1, `:168` on main):

   ```csharp
   private static void HandleTopologyRecoveryException(TopologyRecoveryException e)
   {
       ESLog.Error("Topology recovery exception", e);
       if (e.InnerException is AlreadyClosedException ||
               (e.InnerException is OperationInterruptedException) ||
               (e.InnerException is TimeoutException))
       {
           throw e;
       }
       ESLog.Info($"Will not retry recovery because of {e.InnerException?.GetType().FullName}: it's not a known problem with connectivity, ignoring it", e);
   }
   ```

   `TaskCanceledException` is not in that list, so the exception is **swallowed** and recovery
   continues to `RunRecoveryEventHandlers` → `RecoverySucceededAsync`.
5. Meanwhile the broker completes the `basic.consume` and sends `basic.consume-ok`. Two independent
   guards then prevent the consumer from ever being registered client-side:
   * `Channel.HandleCommandAsync` starts with
     `if (_continuationQueue.ShouldIgnoreCommand(cmd.CommandId)) return;` — the late `basic.consume-ok`
     is dropped outright; and
   * even if it got through, `BasicConsumeAsyncRpcContinuation.DoHandleCommandAsync` passes the
     already-cancelled `CancellationToken` into
     `_consumerDispatcher.HandleBasicConsumeOkAsync(...)`, whose first statement is
     `cancellationToken.ThrowIfCancellationRequested()` — i.e. it throws before `AddConsumer`.
6. From then on `ConsumerDispatcherBase.GetConsumerOrDefault(tag)` misses and returns the
   `FallbackConsumer`, whose `HandleBasicDeliverAsync` logs
   `ESLog.Info($"Unhandled ...")` and returns `Task.CompletedTask` — the delivery is dropped and
   never acked or nacked.

### Additional context

* Also reproduces on `main` @ `c64cc15` ("Merge pull request #1987 …"), i.e. on a build that already
  contains the #1930 fix. Same log signature, same wedged state.
* Raising `ConnectionFactory.ContinuationTimeout` well above the worst-case leader-election time
  narrows the window considerably but does not close it — any outage longer than the timeout
  reproduces the bug. In our cluster scenario the margin is small: with the timeout raised to 120s,
  recovery of the same failure completed in **24.2s**, i.e. only ~4s past the 20s default. That is
  why the failure looks intermittent.
* Broker: RabbitMQ 4.2.0 / Erlang 27.3.4.6. Client: 7.2.1 (and `main`). .NET 8. macOS/Linux.

ansd:
Fixes #1993

Topology recovery silently loses a consumer when the basic.consume it issues does not complete within ContinuationTimeout (20s by default).

An RPC that exceeds ContinuationTimeout completes its continuation as cancelled, so it reaches HandleTopologyRecoveryException as an OperationCanceledException. That type was not in the retry list, so recovery logged "Will not retry recovery because of
System.Threading.Tasks.TaskCanceledException: it's not a known problem with connectivity, ignoring it", skipped the consumer, and went on to report success: "Connection recovery completed" was logged and RecoverySucceededAsync fired.

The broker had not given up, though. basic.consume was already on the wire, so once the queue became available the broker registered an active consumer and started delivering to it. The client had abandoned that RPC, so the late basic.consume-ok was discarded and the consumer was never added to the new channel's ConsumerDispatcher. Every subsequent delivery for that consumer tag then resolved to the FallbackConsumer, which logs at Info level and drops the message without acking it. With a non-zero prefetch the broker waits forever for an acknowledgement that can never come, so the queue stops being consumed altogether - silently, on a connection and channel that both still report IsOpen, with no exception and no event the application can observe.

This is straightforward to hit with quorum queues: while a quorum queue has no leader it cannot answer basic.consume, and electing one can take longer than 20s. It was found on a 3-node cluster where two nodes were killed in quick succession, briefly losing quorum. Measured there, recovery needed 24.2s - only just over the default.

Treat an OperationCanceledException during topology recovery as retryable unless recovery itself was cancelled. A timed-out operation may still be applied by the broker after the client stops waiting, so the client's view of the connection can no longer be trusted and recovery must not be declared successful. Failing the attempt aborts the connection, which clears any consumer the broker did register, and the next attempt re-establishes the topology cleanly.

The exception type is deliberately left alone. The catch (OperationCanceledException) blocks around each RPC are what call RpcContinuationQueue.RpcCanceled, which is what keeps a late reply from being matched to the wrong continuation, and the existing continuation tests assert that a timeout surfaces as an OperationCanceledException.

Verified end to end against RabbitMQ 4.2.0:

  - a 3-node cluster running the reported workload, killing two nodes ~2s apart so quorum is lost. Before: the consumer stops permanently, one message stuck unacked, queue grows without bound. After: the first recovery attempt fails on the timeout, the next succeeds 5s later, and the application resumes and drains the backlog. No deliveries reach the FallbackConsumer.

  - a single-node reproduction that delays the reply to topology recovery's basic.consume past ContinuationTimeout. Before: 3 of 6 messages delivered. After: 6 of 6.

### DIFF
--- a/projects/RabbitMQ.Client/Impl/AutorecoveringConnection.Recovery.cs
+++ b/projects/RabbitMQ.Client/Impl/AutorecoveringConnection.Recovery.cs
@@ -212,6 +212,8 @@ namespace RabbitMQ.Client.Impl
                  * without being acked. With a non-zero prefetch the broker then waits forever for an
                  * acknowledgement that cannot come, and the queue stops being consumed for good -
                  * silently, on a connection and channel that both still report IsOpen.
+                 *
+                 * https://github.com/rabbitmq/rabbitmq-dotnet-client/issues/1993
                  */
                 case OperationCanceledException:
                     return false == recoveryCancellationToken.IsCancellationRequested;
--- a/projects/RabbitMQ.Client/Impl/AutorecoveringConnection.Recovery.cs
+++ b/projects/RabbitMQ.Client/Impl/AutorecoveringConnection.Recovery.cs
@@ -165,16 +165,60 @@ namespace RabbitMQ.Client.Impl
             }
         }
 
-        private static void HandleTopologyRecoveryException(TopologyRecoveryException e)
+        private static void HandleTopologyRecoveryException(TopologyRecoveryException e,
+            CancellationToken recoveryCancellationToken)
         {
             ESLog.Error("Topology recovery exception", e);
-            if (e.InnerException is AlreadyClosedException ||
-                    (e.InnerException is OperationInterruptedException) ||
-                    (e.InnerException is TimeoutException))
+            if (ShouldRetryRecoveryAfter(e.InnerException, recoveryCancellationToken))
             {
                 throw e;
             }
             ESLog.Info($"Will not retry recovery because of {e.InnerException?.GetType().FullName}: it's not a known problem with connectivity, ignoring it", e);
+        }
+
+        /// <summary>
+        /// Decides whether a failure to recover a single topology entity means the whole recovery
+        /// attempt has to be abandoned and retried, rather than skipped over.
+        /// </summary>
+        /// <remarks>
+        /// The distinction that matters is whether the broker may still act on the request after the
+        /// client has given up on it. If it may, the client's view of the connection is no longer
+        /// trustworthy and recovery must not be reported as successful.
+        /// </remarks>
+        internal static bool ShouldRetryRecoveryAfter(Exception? topologyRecoveryInnerException,
+            CancellationToken recoveryCancellationToken)
+        {
+            switch (topologyRecoveryInnerException)
+            {
+                // Known connectivity problems. The connection is gone, so nothing was applied.
+                case AlreadyClosedException:
+                case OperationInterruptedException:
+                case TimeoutException:
+                    return true;
+
+                /*
+                 * A protocol operation that ran past ContinuationTimeout completes its continuation
+                 * as cancelled (AsyncRpcContinuation.HandleContinuationTimeout), so by type alone it
+                 * is indistinguishable from the caller cancelling recovery. Tell them apart with the
+                 * recovery token: if recovery itself was not cancelled, the operation timed out.
+                 *
+                 * A timed-out operation must be retried, because its frame is already on the wire and
+                 * the broker can still apply it after the client has stopped waiting. The case that
+                 * brought this to light is basic.consume against a quorum queue that has lost quorum:
+                 * the queue cannot answer until it elects a new leader, which can take longer than the
+                 * 20s default. Skipping it leaves the broker with a registered, active consumer that
+                 * the client never added to the channel's ConsumerDispatcher, so from then on every
+                 * delivery for that consumer tag resolves to the FallbackConsumer and is discarded
+                 * without being acked. With a non-zero prefetch the broker then waits forever for an
+                 * acknowledgement that cannot come, and the queue stops being consumed for good -
+                 * silently, on a connection and channel that both still report IsOpen.
+                 */
+                case OperationCanceledException:
+                    return false == recoveryCancellationToken.IsCancellationRequested;
+
+                default:
+                    return false;
+            }
         }
 
         private async ValueTask<bool> TryPerformAutomaticRecoveryAsync(CancellationToken cancellationToken)
@@ -341,7 +385,7 @@ namespace RabbitMQ.Client.Impl
                     }
                     else
                     {
-                        HandleTopologyRecoveryException(new TopologyRecoveryException($"Caught an exception while recovering exchange '{recordedExchange}'", ex));
+                        HandleTopologyRecoveryException(new TopologyRecoveryException($"Caught an exception while recovering exchange '{recordedExchange}'", ex), cancellationToken);
                     }
                 }
             }
@@ -432,7 +476,7 @@ namespace RabbitMQ.Client.Impl
                     }
                     else
                     {
-                        HandleTopologyRecoveryException(new TopologyRecoveryException($"Caught an exception while recovering queue '{recordedQueue}'", ex));
+                        HandleTopologyRecoveryException(new TopologyRecoveryException($"Caught an exception while recovering queue '{recordedQueue}'", ex), cancellationToken);
                     }
                 }
 
@@ -507,7 +551,7 @@ namespace RabbitMQ.Client.Impl
                     }
                     else
                     {
-                        HandleTopologyRecoveryException(new TopologyRecoveryException($"Caught an exception while recovering binding between {binding.Source} and {binding.Destination}", ex));
+                        HandleTopologyRecoveryException(new TopologyRecoveryException($"Caught an exception while recovering binding between {binding.Source} and {binding.Destination}", ex), cancellationToken);
                     }
                 }
             }
@@ -587,7 +631,7 @@ namespace RabbitMQ.Client.Impl
                     }
                     else
                     {
-                        HandleTopologyRecoveryException(new TopologyRecoveryException($"Caught an exception while recovering consumer {oldTag} on queue {consumer.Queue}", ex));
+                        HandleTopologyRecoveryException(new TopologyRecoveryException($"Caught an exception while recovering consumer {oldTag} on queue {consumer.Queue}", ex), cancellationToken);
                     }
                 }
             }
--- a/projects/Test/Integration/TestToxiproxy.cs
+++ b/projects/Test/Integration/TestToxiproxy.cs
@@ -37,7 +37,9 @@ using System.Threading;
 using System.Threading.Tasks;
 using Integration;
 using RabbitMQ.Client;
+using RabbitMQ.Client.Events;
 using RabbitMQ.Client.Exceptions;
+using RabbitMQ.Client.Impl;
 using Toxiproxy.Net.Toxics;
 using Xunit;
 using Xunit.Abstractions;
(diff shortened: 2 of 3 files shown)


Zerpet (dismissed):

Zerpet:
The only thing left to consider is whether to back port to `v6.x` branch. I don't have an opinion on that.

lukebakken:
Thanks! I'll get this reviewed today 😸

lukebakken:
> The only thing left to consider is whether to back port to `v6.x` branch

6.x isn't affected, so there's nothing to back port. It's synchronous, and a continuation timeout there surfaces as a real `TimeoutException`: `BlockingCell.WaitForValue(TimeSpan)` throws one at `projects/RabbitMQ.Client/util/BlockingCell.cs:79`, reached through `SimpleBlockingRpcContinuation.GetReply(TimeSpan)`, which is what `ModelBase.BasicConsume` calls with `ContinuationTimeout`. 6.x `HandleTopologyRecoveryException` already lists `TimeoutException` (`AutorecoveringConnection.cs:956`), so it retries.

The mismatch this PR fixes is specific to the async continuation, where `TrySetCanceled` wins and the `TimeoutException` never reaches the awaiter. `AutorecoveringConnection.cs` in 6.x contains no `CancellationToken` at all.

lukebakken (approved):
🥳 thanks!

projects/RabbitMQ.Client/Impl/AutorecoveringConnection.Recovery.cs
```diff
@@ -165,18 +165,64 @@ private async ValueTask StopRecoveryLoopAsync(CancellationToken cancellationToke
...
+                 * the client never added to the channel's ConsumerDispatcher, so from then on every
+                 * delivery for that consumer tag resolves to the FallbackConsumer and is discarded
+                 * without being acked. With a non-zero prefetch the broker then waits forever for an
+                 * acknowledgement that cannot come, and the queue stops being consumed for good -
+                 * silently, on a connection and channel that both still report IsOpen.
+                 *
+                 * https://github.com/rabbitmq/rabbitmq-dotnet-client/issues/1993
+                 */
+                case OperationCanceledException:
+                    return false == recoveryCancellationToken.IsCancellationRequested;
```
gfoidl, line 219: Yoda is greeting 😉.

ansd:
Thank you for reviewing and merging this PR @Zerpet @lukebakken.

The customer is waiting for a new release containing this fix.
Can you please create a new release?

(I see the milestone `7.3.0` got added. This fix can also go into a patch release, but I don't have a strong opinion whether `7.3.0` or `7.2.2` should be cut.)

lukebakken:
OK there's this fix and another that make sense for `7.2.2`. I'll get that today.

projects/RabbitMQ.Client/Impl/AutorecoveringConnection.Recovery.cs
lukebakken: Hah yes well I'm a weirdo that finds it much easier to notice `false == ...` instead of `!` 😸

gfoidl: 😉 
A compromise would be `foo is false` 😃.

lukebakken: I still prefer the `false == ` to be first as it is visually jarring and brings attention to the condition. It's a holdover from my C days that I'm going to continue using in this repo.

lukebakken:
@ansd - done! Published a couple hours ago - https://www.nuget.org/packages/RabbitMQ.Client/7.2.2

ansd:
Thank you, Luke!

lukebakken:
@ansd please ping me directly if the `7.2.2` fix doesn't resolve this issue as expected!

evgenyvalavin:
Thanks everyone for a quick fix! My apps were recently affected by this bug

PR #2000, opened 2026-08-12
Await connection.open-ok before completing connection creation

### LINKED ISSUE #1999: CreateConnectionAsync can complete before connection.open-ok is received
### Describe the bug

With RabbitMQ.Client 7.1.2, `CreateConnectionAsync` can complete after
`connection.open` has only been handed to the transport, before the server's
`connection.open-ok` is received. The same behavior is present on `main` at
`4cdb49044f880c251e35c07d7fec1013a5c3fddf`.

`Connection.OpenAsync` awaits `Channel.ConnectionOpenAsync`, but that method
only awaits `ModelSendAsync`; it does not register or await an RPC continuation
for `ProtocolCommandId.ConnectionOpenOk`. Application code can therefore call
`CreateChannelAsync` and send `channel.open` while the connection and virtual
host are not yet ready.

AMQP 0-9-1 defines connection opening as `C:OPEN S:OPEN-OK`;
`connection.open-ok` signals that the connection is ready for use:

https://github.com/rabbitmq/amqp-0.9.1-spec/blob/main/docs/amqp-0-9-1-reference.md#connectionopen-ok

The 6.x client also treated `connection.open` as an RPC and waited for its reply
using `HandshakeContinuationTimeout`, so this appears to be a regression in the
asynchronous 7.x implementation.

### Reproduction steps

1. Use a fake AMQP 0-9-1 peer that completes Start/Tune, receives
   `connection.open`, but holds `connection.open-ok` behind a gate.
2. Start connection creation:

```csharp
Task<IConnection> connectionTask = factory.CreateConnectionAsync();
await peer.ConnectionOpenReceived;
```

3. Before releasing the Open-Ok gate, observe that `connectionTask` is already
   complete.
4. Await it and immediately create a channel:

```csharp
IConnection connection = await connectionTask;
Task<IChannel> channelTask = connection.CreateChannelAsync();
await peer.ChannelOpenReceived;
```

The captured order is:

```text
C: connection.open
C: channel.open
S: connection.open-ok   # deliberately delayed
```

A peer that requires virtual-host initialization to finish before accepting
channel methods can reject the premature `channel.open` with `connection.close`
reply code 530 (for example, reply text `NotSafeAccessError`).

### Expected behavior

`CreateConnectionAsync` must remain incomplete until `connection.open-ok` is
received. No channel-level method should be sent first.

A server close, disconnect, handshake timeout, or caller cancellation before
Open-Ok should fail the connection attempt.

### Additional context

Affected versions checked:

- RabbitMQ.Client 7.1.2 (`c48e7b08485d07b98d817a71debeba16daff9f9b`)
- `main` (`4cdb49044f880c251e35c07d7fec1013a5c3fddf`)

A deterministic unit test using a controlled `ISession` reproduces the early
completion on the current implementation. A fix and regression tests are ready
in a follow-up pull request.

fuyou001:
## Proposed Changes

Fixes #1999

AMQP 0-9-1 defines connection opening as `C:OPEN S:OPEN-OK`, and
`connection.open-ok` signals that the connection is ready for use. Currently,
`ConnectionOpenAsync` only waits for the write of `connection.open`, allowing
`CreateConnectionAsync` to return before Open-Ok and permitting `channel.open`
to be sent too early.

This change:

- adds a channel-0 RPC continuation that expects
  `ProtocolCommandId.ConnectionOpenOk`;
- acquires the existing RPC semaphore and enqueues the continuation before
  sending `ConnectionOpen`, so an immediate Open-Ok cannot be lost;
- waits for Open-Ok using `HandshakeContinuationTimeout`;
- preserves peer shutdown reasons and uses the existing late-response cleanup
  for timeouts;
- cleans up the continuation without releasing an RPC semaphore permit when a
  caller is cancelled before acquiring it;
- makes initial connection creation and automatic recovery wait until the
  server confirms that the connection is ready.

Protocol reference:

https://github.com/rabbitmq/amqp-0.9.1-spec/blob/main/docs/amqp-0-9-1-reference.md#connectionopen-ok

### Test Plan

- [x] A controlled session delaying Open-Ok keeps `ConnectionOpenAsync`
      incomplete until Open-Ok is delivered.
- [x] Open-Ok delivered from inside `TransmitAsync` is not lost.
- [x] Peer shutdown while waiting preserves the complete shutdown reason in an
      `OperationInterruptedException`.
- [x] `HandshakeContinuationTimeout` cancels the wait, a late Open-Ok is
      absorbed, and the next RPC completes normally.
- [x] Cancellation while waiting for the RPC semaphore sends no frame, releases
      no unowned permit, and does not prevent the next RPC.
- [x] `dotnet test projects/Test/Unit/Unit.csproj --no-restore` (156 passed).
- [x] `dotnet build projects/RabbitMQ.Client/RabbitMQ.Client.csproj --no-restore`
      (`net8.0` and `netstandard2.0`, zero warnings/errors).
- [x] `dotnet format --no-restore --verify-no-changes` for all changed files.
- [x] The integration suite was not run locally because it requires a RabbitMQ
      node plus the repository's TLS/Toxiproxy test environment.

### Compatibility

There are no public API, configuration, persistence, or wire-format changes.
Conforming AMQP 0-9-1 servers already send Open-Ok; connection creation now
includes the final handshake response before exposing the connection. A stalled
or non-conforming peer that never sends Open-Ok will now fail with the configured
handshake timeout instead of returning an unusable connection.

Automatic recovery uses the same `Connection.OpenAsync` path, so takeover,
topology recovery (when enabled), channel recovery, and the recovery-success
event remain gated on a fully opened connection.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #1999)

## Checklist

- [x] I have signed the CLA (see https://github.com/rabbitmq/cla)

## Further Comments

The implementation follows the existing channel RPC continuation pattern. It
intentionally leaves `ConnectionOpenOk` on `DispatchCommandAsync`'s default RPC
response path and keeps the send operation outside the reply-cancellation catch:
a failed send must not register a response as potentially late.

### DIFF
--- a/projects/RabbitMQ.Client/Impl/Channel.cs
+++ b/projects/RabbitMQ.Client/Impl/Channel.cs
@@ -275,11 +275,51 @@ namespace RabbitMQ.Client.Impl
 
         internal async ValueTask ConnectionOpenAsync(string virtualHost, CancellationToken cancellationToken)
         {
-            using var timeoutTokenSource = new CancellationTokenSource(HandshakeContinuationTimeout);
-            using var lts = CancellationTokenSource.CreateLinkedTokenSource(timeoutTokenSource.Token, cancellationToken);
-            var method = new ConnectionOpen(virtualHost);
-            // Note: must be awaited or else the timeoutTokenSource instance will be disposed
-            await ModelSendAsync(in method, lts.Token).ConfigureAwait(false);
+            bool enqueued = false;
+            bool semaphoreAcquired = false;
+            var k = new ConnectionOpenAsyncRpcContinuation(HandshakeContinuationTimeout, cancellationToken);
+
+            try
+            {
+                await _rpcSemaphore.WaitAsync(k.CancellationToken)
+                    .ConfigureAwait(false);
+                semaphoreAcquired = true;
+
+                k.StartTimeout();
+
+                enqueued = Enqueue(k);
+
+                try
+                {
+                    var method = new ConnectionOpen(virtualHost);
+                    await ModelSendAsync(in method, k.CancellationToken)
+                        .ConfigureAwait(false);
+                }
+                catch (AlreadyClosedException)
+                {
+                    // let continuation throw OperationInterruptedException,
+                    // which is a much more suitable exception before connection
+                    // negotiation finishes
+                }
+
+                try
+                {
+                    AssertResultIsTrue(await k);
+                }
+                catch (OperationCanceledException)
+                {
+                    _continuationQueue.RpcCanceled(k.ResponseReceived, k.HandledProtocolCommandIds);
+                    throw;
+                }
+            }
+            finally
+            {
+                MaybeDisposeContinuation(enqueued, k);
+                if (semaphoreAcquired)
+                {
+                    _rpcSemaphore.Release();
+                }
+            }
         }
 
         internal async ValueTask<ConnectionSecureOrTune> ConnectionSecureOkAsync(byte[] response,
--- a/projects/RabbitMQ.Client/Impl/Channel.cs
+++ b/projects/RabbitMQ.Client/Impl/Channel.cs
@@ -280,6 +280,11 @@ namespace RabbitMQ.Client.Impl
             var k = new SimpleAsyncRpcContinuation(ProtocolCommandId.ConnectionOpenOk,
                 HandshakeContinuationTimeout, cancellationToken);
 
+            // Unlike the sibling Connection*Async methods, the semaphore is
+            // acquired inside the try (guarded by semaphoreAcquired) so that a
+            // cancellation of WaitAsync during connection open still reaches the
+            // finally and disposes k. Acquiring it outside the try, as the
+            // siblings do, leaks k's CancellationTokenSource on that path.
             try
             {
                 await _rpcSemaphore.WaitAsync(k.CancellationToken)
--- a/projects/RabbitMQ.Client/Impl/Channel.cs
+++ b/projects/RabbitMQ.Client/Impl/Channel.cs
@@ -277,7 +277,8 @@ namespace RabbitMQ.Client.Impl
         {
             bool continuationEnqueued = false;
             bool semaphoreAcquired = false;
-            var k = new ConnectionOpenAsyncRpcContinuation(HandshakeContinuationTimeout, cancellationToken);
+            var k = new SimpleAsyncRpcContinuation(ProtocolCommandId.ConnectionOpenOk,
+                HandshakeContinuationTimeout, cancellationToken);
 
             try
             {
--- a/projects/RabbitMQ.Client/Impl/Channel.cs
+++ b/projects/RabbitMQ.Client/Impl/Channel.cs
@@ -275,7 +275,7 @@ namespace RabbitMQ.Client.Impl
 
         internal async ValueTask ConnectionOpenAsync(string virtualHost, CancellationToken cancellationToken)
         {
-            bool enqueued = false;
+            bool continuationEnqueued = false;
             bool semaphoreAcquired = false;
             var k = new ConnectionOpenAsyncRpcContinuation(HandshakeContinuationTimeout, cancellationToken);
 
@@ -283,11 +283,11 @@ namespace RabbitMQ.Client.Impl
             {
                 await _rpcSemaphore.WaitAsync(k.CancellationToken)
                     .ConfigureAwait(false);
-                semaphoreAcquired = true;
+                Volatile.Write(ref semaphoreAcquired, true);
 
                 k.StartTimeout();
 
-                enqueued = Enqueue(k);
+                continuationEnqueued = Enqueue(k);
 
                 try
                 {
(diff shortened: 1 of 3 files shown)


michaelklishin:
Thank you. May I ask you to sign our [CLA](https://github.com/rabbitmq/cla)? We'd need it in order to accept a non-trivial set of changes.

fuyou001:
@michaelklishin Thanks! I have emailed the RabbitMQ team to request a CLA signing invitation. I will complete the signing once I receive the Box Sign request and will confirm here afterward.

fuyou001:
CI note: the only failing check is `integration-win32`, where the existing `TestToxiproxy.TestRpcContinuationTimeout_GH1802` failed on .NET Framework while closing a channel after removing the downstream bandwidth toxic. The same test passed in the net8.0 portion of that job, and the complete `sequential-integration-win32` job passed (19m17s).

The failure stack is in `Channel.CloseAsync` from `TestToxiproxy.cs:461`; it does not pass through the new `ConnectionOpenAsync` path. This PR does not change `CloseAsync` or `TestToxiproxy.cs` (the test is identical to `main`). The other Windows build, Ubuntu integration/sequential integration, CodeQL, and OAuth checks passed.

I do not have repository permission to re-run the failed job. Could a maintainer please re-run `integration-win32`?

michaelklishin:
@fuyou001 I've sent you a CLA to sign. Thanks.

fuyou001:
@michaelklishin I have now completed the RabbitMQ CLA signing via Box Sign. Please let me know if any further action is required. Thank you!

fuyou001:
@michaelklishin One additional note for merging: the CLA has been signed, and the only remaining failed check is `integration-win32`. The failure is from the existing `TestToxiproxy.TestRpcContinuationTimeout_GH1802` test on .NET Framework while closing a channel after removing the downstream bandwidth toxic. The same test passed in the .NET 8 portion of that job, and `sequential-integration-win32` passed completely.

The failure occurs in `Channel.CloseAsync` and does not exercise the `ConnectionOpenAsync` path changed by this PR, so it appears unrelated to this change. Could a maintainer please re-run `integration-win32` or treat this failure as non-blocking when merging? Thank you!

projects/RabbitMQ.Client/Impl/Channel.cs
```diff
@@ -275,11 +275,51 @@ await ConsumerDispatcher.WaitForShutdownAsync(cancellationToken)
 
         internal async ValueTask ConnectionOpenAsync(string virtualHost, CancellationToken cancellationToken)
         {
-            using var timeoutTokenSource = new CancellationTokenSource(HandshakeContinuationTimeout);
-            using var lts = CancellationTokenSource.CreateLinkedTokenSource(timeoutTokenSource.Token, cancellationToken);
-            var method = new ConnectionOpen(virtualHost);
-            // Note: must be awaited or else the timeoutTokenSource instance will be disposed
-            await ModelSendAsync(in method, lts.Token).ConfigureAwait(false);
+            bool enqueued = false;
+            bool semaphoreAcquired = false;
```
gfoidl, line 279: I think this should be written to / read from via `Volatile.Write/Read` (or use `volatile` modifier, but for explicitness I'd prefer the `Volatile`-methods).

Due the `await` the remainder can jump to another thread on another CPU and the .NET CLR relies on the CPU's memory model for non-reference types.
Meaning later on the `finally` the `semaphoreAcquired` could be a stale value, so the semaphore won't be released. And this is hard to diagnose 🐛 if it happens.

projects/RabbitMQ.Client/Impl/Channel.cs
```diff
@@ -275,11 +275,51 @@ await ConsumerDispatcher.WaitForShutdownAsync(cancellationToken)
 
         internal async ValueTask ConnectionOpenAsync(string virtualHost, CancellationToken cancellationToken)
         {
-            using var timeoutTokenSource = new CancellationTokenSource(HandshakeContinuationTimeout);
-            using var lts = CancellationTokenSource.CreateLinkedTokenSource(timeoutTokenSource.Token, cancellationToken);
-            var method = new ConnectionOpen(virtualHost);
-            // Note: must be awaited or else the timeoutTokenSource instance will be disposed
-            await ModelSendAsync(in method, lts.Token).ConfigureAwait(false);
+            bool enqueued = false;
```
gfoidl: It's never read from, only assigned to in L288.
So it could be removed, and in L288 use `_ = Enqueue(k);`.

Note: I see this pattern is used all over in this file, but I don't know why 😉.

projects/RabbitMQ.Client/Impl/Channel.cs
fuyou001: Thank you for the careful review! Updated in 7f658dd49 to use Volatile.Write/Read for semaphoreAcquired.

projects/RabbitMQ.Client/Impl/Channel.cs
fuyou001: Thank you for taking a close look! `enqueued` is read in the `finally` block by `MaybeDisposeContinuation(...)`. Its value distinguishes queue ownership from `Enqueue(k)` returning `false` after the channel closes, in which case the continuation still needs to be disposed locally. Removing it would lose that cleanup path. I renamed it to `continuationEnqueued` in 7f658dd49 to make this purpose and the read in `finally` clearer.

gfoidl: Ah, thanks. I didn't spot it before (and browser's search tricked me by including a trailing space).

The rename is 👍🏻.

lukebakken (approved):
Thanks! I have no idea why this hasn't been noticed. It must be due to the timing involved.

PR #2004, opened 2026-08-24
Fix recovery could miss closed connection notification

zvirja:
## Proposed Changes

I have found a race-condition in the connection recovery loop which could lead to connection being never recovered. It happens as a race condition of a few factors:
- we don't start a new recovery task if there is an existing one
- we reset existing recovery task handle some time after the connection was open, adding an unsafe gap
- we run user callbacks after we re-open connection and before we reset the existing recovery task handle, so race-condition window could be large

If newly re-opened connection shortly closes before we reset the handle, we effectively miss it and never recover the connection again.

I fixed it by adding a flag, so we could signal to the existing recovery that a new recovery request has arrived. After long time of playing with lock-free I ended up having a lock, as code is much simpler this way and it's not a hot path anyway. Lock-free is hard as we have to synchronize both `recoveryTask` and `recoveryPendingRequest` in a right way and it becomes tricky. Lock is trivial.

I also did a bit of robustness enhancements:
- cancel recovery token unconditionally, as race could happen between running this loop and a new recovery request coming in the meanwhile
- capture cancellation token early in the method and check for the disposal flag, as disposal loop without closing the connection could lead to _potentially_ infinite recovery loop (token not cancelled, `success = false` always)

Added a unit test which demonstrates the issue (and fails before the fix).

Related issue: #1871 (not marking as a fix as there could be a different root-cause).

## Types of Changes

- [ч] Bug fix (non-breaking change which fixes issue #NNNN)

## Checklist

- [ ] I have signed the CLA (see https://github.com/rabbitmq/cla)

## Further Comments

considered, etc.

### DIFF
--- a/projects/RabbitMQ.Client/Impl/AutorecoveringConnection.Recovery.cs
+++ b/projects/RabbitMQ.Client/Impl/AutorecoveringConnection.Recovery.cs
@@ -43,6 +43,8 @@ namespace RabbitMQ.Client.Impl
 {
     internal sealed partial class AutorecoveringConnection
     {
+        // TODO: Use Lock once update to .NET 9+
+        // https://learn.microsoft.com/en-us/dotnet/api/system.threading.lock?view=net-10.0
         private readonly object _recoverySync = new object();
         private Task? _recoveryTask;
         private bool _recoveryPendingRequest;
--- a/projects/RabbitMQ.Client/Impl/AutorecoveringConnection.Recovery.cs
+++ b/projects/RabbitMQ.Client/Impl/AutorecoveringConnection.Recovery.cs
@@ -37,24 +37,35 @@ using System.Threading;
 using System.Threading.Tasks;
 using RabbitMQ.Client.Events;
 using RabbitMQ.Client.Exceptions;
-using RabbitMQ.Client.Framing;
 using RabbitMQ.Client.Logging;
 
 namespace RabbitMQ.Client.Impl
 {
     internal sealed partial class AutorecoveringConnection
     {
+        private readonly object _recoverySync = new object();
         private Task? _recoveryTask;
+        private bool _recoveryPendingRequest;
         private readonly CancellationTokenSource _recoveryCancellationTokenSource = new CancellationTokenSource();
 
         private Task HandleConnectionShutdownAsync(object? _, ShutdownEventArgs args)
         {
             if (ShouldTriggerConnectionRecovery(args))
             {
-                var recoverTask = new Task<Task>(RecoverConnectionAsync);
-                if (Interlocked.CompareExchange(ref _recoveryTask, recoverTask.Unwrap(), null) is null)
+                lock (_recoverySync)
                 {
-                    recoverTask.Start();
+                    if (_recoveryTask == null)
+                    {
+                        var recoverTask = new Task<Task>(RecoverConnectionAsync);
+                        _recoveryTask = recoverTask.Unwrap();
+                        recoverTask.Start();
+                    }
+                    else
+                    {
+                        // Notify current recovery task about new recovery request,
+                        // as there is no other task to catch it.
+                        _recoveryPendingRequest = true;
+                    }
                 }
             }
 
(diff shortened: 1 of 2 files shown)


projects/RabbitMQ.Client/Impl/AutorecoveringConnection.Recovery.cs
```diff
@@ -37,24 +37,49 @@
...
 using RabbitMQ.Client.Events;
 using RabbitMQ.Client.Exceptions;
-using RabbitMQ.Client.Framing;
 using RabbitMQ.Client.Logging;
 
 namespace RabbitMQ.Client.Impl
 {
     internal sealed partial class AutorecoveringConnection
     {
+        private readonly object _recoverySync = new object();
```
paulomorgado, line 48: Consider using [`Lock`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.lock) instead of `object` for `NET9_0_OR_GREATER`.

To keep the code clean, a pollyfil could be created:

```csharp
#if !NET9_0_OR_GREATER
internal sealed class Lock : object { }
#endif
```

For more information:

- [What's new in C# 13 - New lock object](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-13#new-lock-object)

paulomorgado (changes requested):

lukebakken (dismissed):
Code review findings addressed in c8d4b56 (recovery-loop hardening: token capture moved inside the try to avoid an unobserved throw that could wedge recovery, a `_disposed` guard in the shutdown handler, and `_recoveryTask` read under `_recoverySync` in the stop path). Full CI matrix is green, including integration and sequential-integration on both Ubuntu and Windows.

projects/RabbitMQ.Client/Impl/AutorecoveringConnection.Recovery.cs
zvirja: I would be happy to apply the change - but I noticed that you target only `net8` - so it will never be picked. Maybe it makes sense to do through the whole app once you start supporting .NET 9+

zvirja:
@lukebakken @paulomorgado Thank you guys for a really fast follow up! 🙏  Let me know if you want me to apply that locking thing :)

projects/RabbitMQ.Client/Impl/AutorecoveringConnection.Recovery.cs
lukebakken: Let's add a comment with a link to the .NET 10 docs for `Lock`:

https://learn.microsoft.com/en-us/dotnet/api/system.threading.lock?view=net-10.0

That will alert future-me and other maintainers to look at that when we move to that .NET release.

zvirja: Done! Hope that's what you wanted

lukebakken (dismissed):
Thanks!

projects/RabbitMQ.Client/Impl/AutorecoveringConnection.Recovery.cs
```diff
@@ -37,24 +37,51 @@
...
+                        // still tolerates a disposed source by capturing the token inside its
+                        // try.
+                        return Task.CompletedTask;
+                    }
+
+                    if (_recoveryTask == null)
+                    {
+                        var recoverTask = new Task<Task>(RecoverConnectionAsync);
+                        _recoveryTask = recoverTask.Unwrap();
+                        recoverTask.Start();
```
gfoidl: Is

```suggestion
                        _recoveryTask = RecoverConnectionAsync();
```
not enough? Maybe put a `await Task.Yield()` in `RecoverConnectionAsync` to put it async.
Or just a `Task.Run(RecoverConnectionAsync)`?

zvirja: I just repeated the code which was already there - so I'll leave it up to you.

But I would imagine that the original design was to run it in the background and don't block current method.

lukebakken: > [!NOTE]
> This reply was drafted by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who reviewed everything before posting.

The bare `_recoveryTask = RecoverConnectionAsync();` isn't safe: `RecoverConnectionAsync` can complete synchronously (`IsOpen` already true, or the token cancelled / connection disposed, the close-races-shutdown case). It then runs inline while we hold `_recoverySync`, its `finally` sets `_recoveryTask = null`, and the outer assignment resurrects the completed task into the field, wedging recovery for good.

`Task.Yield()` or `Task.Run` both avoid that by assigning the field before the body runs. Went with `Task.Run(RecoverConnectionAsync)`: one line, no `Unwrap`, and it uses `TaskScheduler.Default` instead of the `TaskScheduler.Current` the bare `.Start()` captured. Thanks!

lukebakken (approved):

lukebakken:
Thank you @zvirja

PR #2005, opened 2026-08-25
Fix potential deadlocks in connection shutdown sequence

zvirja:
### Current design

I'll first start with describing existing library design, so it's more visible what are the issues with the existing behavior.

1. First we create `Connection` which maintains events, in particular `ConnectionShutdownAsync`
2. External consumers and libraries (`EastyNetQ` in particular) are wrapping those object and subscribing to the event to perform lifecycle management. Subscriber is added as a first thing after connection is created (as consumption happens dynamically later).
3. Later we create sessions/channels. Each `SessionBase` subscribes to the `Connection.ConnectionShutdownAsync` to maintain its lifecycle.

With such a design the external subscribers of the `Connection.ConnectionShutdownAsync` are invoked __before__ `SessionBase` (which are subscribing later as we start consuming). There are a few issues with such design:
- Sessions/Channels are not fully closed yet, so external library will deal with half-updated state
- External libraries could introduce substantial delay before sessions are updated
- When external libraries (`EasyNetQ` in particluar) add locking to their handlers, it quickly leads to dead-lock when using connection auto-recovery.
- There is no way for external libraries to subscribe _after_ RabbitMQ is fully uninitialized, so e.g. unsubscribing is behaving correctly.

### Deadlock scenario

I used AI generated project to reproduce how a glitch in the connection reliably leads to a dead-lock: 
[rmq7-deadlock-repro.zip](https://github.com/user-attachments/files/31424647/rmq7-deadlock-repro.zip)

Design is very simple. Create `ConsumerRegistry` which works with RabbitMQ under the lock (to make sure that things are tracked correctly and are in sync). It emulates what `EasyNetQ` library is doing:

<details><summary>ConsumerRegistry</summary>

```c#
sealed class ConsumerRegistry
{
    public async Task StartConsumingAsync(IChannel channel, string queue)
    {
        await _sync.WaitAsync();
        try
        {
            _consumerTags[queue] = await channel.BasicConsumeAsync(queue, autoAck: false, consumer: new AsyncEventingBasicConsumer(channel));
        }
        finally
        {
            _sync.Release();
        }
    }

    public async Task OnConnectionShutdownAsync(object? sender, ShutdownEventArgs e)
    {
        await _sync.WaitAsync();
        try
        {

            _consumerTags.Clear();
        }
        finally
        {
            _sync.Release();
        }
    }
}
```
</details> 

Then just wire it up and use it:

```c#
var registry = new ConsumerRegistry(Log);

// An application that tracks its own consumers subscribes here
connection.ConnectionShutdownAsync += registry.OnConnectionShutdownAsync;

var channel = await connection.CreateChannelAsync();
await channel.QueueDeclareAsync(Queue, durable: false, exclusive: false, autoDelete: true);

var consuming = registry.StartConsumingAsync(channel, Queue);
```

If there is a network glitch inside the consuming request, you will get a reliable dead-lock. 

Thread A

```
ConsumerRegistry.StartConsumingAsync(IChannel, string)                           // holds the semaphore
  RabbitMQ.Client.Impl.AutorecoveringChannel.BasicConsumeAsync(...)
    RabbitMQ.Client.Impl.Channel.BasicConsumeAsync(...)
      await BasicConsumeAsyncRpcContinuation  ->  Task<string>                   // <-- parked here
```

Thread B

```
RabbitMQ.Client.Framing.Connection.MainLoop()
  RabbitMQ.Client.Framing.Connection.HandleMainLoopExceptionAsync(ShutdownEventArgs)
    RabbitMQ.Client.Impl.AsyncEventingWrapper<ShutdownEventArgs>.InternalInvoke(Delegate[], object, ShutdownEventArgs)
      ConsumerRegistry.OnConnectionShutdownAsync(object, ShutdownEventArgs)      // application handler
        System.Threading.SemaphoreSlim.WaitAsync()                               // <-- parked here, waits for semaphore
```

What happens here is that `StartConsumingAsync` is making a request and awaits a confirmation from the server. That confirmation could arrive in the main loop only, but instead of confirmation we got connection exception. Usually RabbitMQ would handle that in `SessionBase` handler (propagating it further down) by failing any outgoing requests - but in our case the `Connection.ConnectionShutdownAsync` is invoked too early. So we are in a dead-lock situation.

There are more complex variations of this happening (e.g. ones involving auto-recovery and re-connection on auto-recovery events), but root-cause is still the same - pending `Channel` operations fired from another thread never return when main loop is blocked.

This issue actively happens with EasyNetQ and it's a hell to figure out what is going on.

If RabbitMQ consumed shutdown in `SessionBase` events _before_ publishing `Connection.ConnectionShutdownAsync` - the dead-lock would not happen, as session/channel would be correctly cancelled by the time external consumers are called.

### Design

One would say that the scenario I demonstrated above is not caused by the library, but is a misuse of the library. You could push back on that. But I would argue that the issue is in the library design:
- The usage with locking is legit and is most likely needed to synchronize with library, especially when there are a lot of concurrent operations. If we start running everything in a background thread, the state would be less deterministic and way harder to synchronize upon.
- It's not expected that library is in a half-closed state when `Connection.ConnectionShutdownAsync` is fired. Usually the semantic of the event is that thing _already happened_, not _is happening_. That's true for the other events.
- There is nothing dictating that sessions/channels shall be disposed after the event is raised. It's a side-effect of the design when we use the same event which external subscribers subscribe to. If we just had a loop disposing channels, it would happen correctly.
- There is no technical possibility to subscribe later at a reliable moment of time when things are settled down. We have to just run background operation with an arbitrary window (e.g. 5 seconds) and hope that things are stable by then.

With the current design we first create connection and then sessions/channel. So the disposal order shall be reverse - first channels are disposed, then connection. And such shall be events order.

### Solution

We need a solution to fix `EasyNetQ` library and I believe that the fix shall be here. My suggestion is to simply introduce a separate internal `Connection.ConnectionShutdown0Async` event and invoke session handlers before. So this way library will dispose sessions/channels and update its state before firing an external event. This change is not breaking in terms of API and is mostly a behavior change.

The change could still be breaking if somebody relied on the event handlers order (that `Session.SessionShutdownAsync` was called _after_ `Connection.ConnectionShutdownAsync`) - but I would say that one shouldn't have relied on that. 

To me the stability of the API is worth the risk, as current design is error-prone and it's hard to reliably work around it.

We used patched RabbitMQ library in our project for a few years where we had this fix and it all was working like a charm. So we want to contribute it back, as investigating those dead-locks is hard.

### Discussion

I only made a change without adding tests - to see what your initial response is. I don't see other nice solutions and would be happy to proceed with it further. But I am totally open if you have other clever ideas or thoughts.

Would be nice to hear what you think.

Thank you 🙏 

## Types of Changes

- [x] Breaking change (POTENTIALLY ???)

## Checklist

- [ ] I have signed the CLA (see https://github.com/rabbitmq/cla)

## Further Comments

considered, etc.

### DIFF
--- a/docs/internal/connection-shutdown-and-cancellation.md
+++ b/docs/internal/connection-shutdown-and-cancellation.md
@@ -8,7 +8,7 @@ Every connection has a hidden channel/session pair on channel number 0 (`_channe
 
 The single most important and least obvious fact about shutdown:
 
-> **Session 0 does not subscribe to `Connection.ConnectionShutdownAsync`.**
+> **Session 0 does not subscribe to either connection-shutdown event.**
 
 See `SessionBase` construction:
 
@@ -20,13 +20,15 @@ protected SessionBase(Connection connection, ushort channelNumber)
     ChannelNumber = channelNumber;
     if (channelNumber != 0)
     {
-        connection.ConnectionShutdownAsync += OnConnectionShutdownAsync;
+        connection.PreConnectionShutdownAsync += OnConnectionShutdownAsync;
     }
     ...
 }
 ```
 
-Application channels (channel number != 0) are shut down automatically when the connection shuts down, because their session is wired to `ConnectionShutdownAsync`. That event handler chain ends in `Channel.OnSessionShutdownAsync -> OnChannelShutdownAsync -> _continuationQueue.HandleChannelShutdown(reason)`, which faults any pending RPC continuation with an `OperationInterruptedException`.
+Application channels (channel number != 0) are shut down automatically when the connection shuts down, because their session is wired to `PreConnectionShutdownAsync`. That event handler chain ends in `Channel.OnSessionShutdownAsync -> OnChannelShutdownAsync -> _continuationQueue.HandleChannelShutdown(reason)`, which faults any pending RPC continuation with an `OperationInterruptedException`.
+
+`PreConnectionShutdownAsync` is an internal event that fires *before* the public `ConnectionShutdownAsync`: `Connection.OnShutdownAsync` awaits every pre-shutdown handler to completion, then awaits the public handlers. The ordering is load-bearing - it guarantees every application session is closed before `AutorecoveringConnection.HandleConnectionShutdownAsync` (a public handler) starts automatic recovery. The subscribe in the `SessionBase` constructor and the unsubscribe in `OnSessionShutdownAsync` must target the same event; a mismatch leaks every closed session's handler onto the connection for its lifetime and re-fires session shutdown when the connection later closes.
 
 Channel 0 has **no such wiring**. Its only shutdown path is `Connection.FinishCloseAsync`:
 
--- a/projects/RabbitMQ.Client/Impl/Connection.cs
+++ b/projects/RabbitMQ.Client/Impl/Connection.cs
@@ -75,8 +75,8 @@ namespace RabbitMQ.Client.Impl
             _connectionUnblockedAsyncWrapper =
                 new AsyncEventingWrapper<AsyncEventArgs>("OnConnectionUnblocked", onExceptionAsync);
 
-            _connectionShutdown0AsyncWrapper =
-                new AsyncEventingWrapper<ShutdownEventArgs>("OnShutdown0", onExceptionAsync);
+            _preConnectionShutdownAsyncWrapper =
+                new AsyncEventingWrapper<ShutdownEventArgs>("OnPreShutdown", onExceptionAsync);
 
             _connectionShutdownAsyncWrapper =
                 new AsyncEventingWrapper<ShutdownEventArgs>("OnShutdown", onExceptionAsync);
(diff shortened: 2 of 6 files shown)


michaelklishin:
@zvirja thank you for taking the time to explain what you are fixing and why.

Can you please make the title less dramatic, though? You are not "fixing the design", in fact, your change is fairly small in scope in the grand scheme of things.

zvirja:
@michaelklishin Didn't mean to be dramatic 🤣 Fair enough, changed the title.

zvirja:
@lukebakken Thanks. Damn, I overlooked unsubscribing from the right pair inside the session 😖  Nice to have code review!

I also thought about pre-shutdown event name and it was my early revision, but it was confusing to me as well, as in reality it's connection shutdown event - it's not like it happens _before_ any shutdown or there is a pre-shutdown phase. Another thought was "Priority" or `ConnectionShutdownPreAsync` (to not indicate that it's a pre-shutdown, but shutdown event fired earlier). But that was ugly as well. Agree that 0 is not nicer either - but that looked the least ugly to me 🤷‍♂️ 

But given that it's internal, I guess the naming doesn't matter much.

lukebakken:
No worries, I had the 🧞 do the review. I'm watching CI now.

lukebakken (approved):
Thanks @zvirja

Issue #1981, opened 2026-07-30
Tracing configuration is process-global and last writer wins (tracing review group C)


lukebakken:
> [!NOTE]
> Claude wrote this issue under the direction of @lukebakken. The defect and the verification behind it come from the review in #1967; the options below are AI-drafted starting points, not settled decisions.

Group C of #1967: tracing configuration is process-global and last-writer-wins. Split out because it is **the only part of this work that changes the public surface of `RabbitMQ.Client`**, which puts it ahead of #1923 in the 7.3.0 ordering.

## The defect

`RabbitMQActivitySource.TracingOptions` is a public settable static holding a mutable object, and `AddRabbitMQInstrumentation` replaces it wholesale while also overwriting both propagation delegates:

```csharp
// projects/RabbitMQ.Client.OpenTelemetry/TraceProviderBuilderExtensions.cs
RabbitMQActivitySource.TracingOptions = options;
RabbitMQActivitySource.ContextExtractor = OpenTelemetryContextExtractor;
RabbitMQActivitySource.ContextInjector = OpenTelemetryContextInjector;
```

**Verified.** Two independent `TracerProvider`s, each calling `AddRabbitMQInstrumentation` with different options: after the second call the first provider's configuration is silently gone, and *both* exporters receive spans shaped by the second. Disposing the second provider restores nothing - `ContextInjector` stays pointed at the OpenTelemetry implementation and `TracingOptions` keeps its values for the life of the process.

This is not a memory-safety problem. 181,425 publishes with a concurrent writer swapping the options object produced zero exceptions, because reference assignment is atomic. The defect is the **ownership model**: per-provider configuration expressed as process-global mutable state.

The statics are also unvalidated, so `ContextInjector = null` makes every subsequent publish throw `NullReferenceException` from inside the client.

## It already bites our own test suite

Not hypothetical. While writing the group A tests (#1978) three of five new tests failed with `Expected: 1 / Actual: 0` and no indication why. Cause: `UseRoutingKeyAsOperationName` defaults to `true`, so publish spans are named `publish <routing-key>`, and no test restores the flag after mutating it - so outcomes depend on execution order.

`TestOpenTelemetry.TestDefaultTracingOptions` asserts the default is `true`, and would fail if an earlier test left it `false`. The group A tests added a `PlainOperationNames` scope guard that restores the prior value, but that is a workaround for the design, not a fix.

[...]

Issue #1979, opened 2026-07-30
OpenTelemetry tracing: 7.3.0 work


lukebakken:
> [!NOTE]
> Claude wrote this issue under the direction of @lukebakken, who asked for a parent issue to track the OpenTelemetry work and chose the structure. The status table reflects work already done or filed; the split of remaining work into groups B and C is from the review in #1967.

Umbrella issue for the OpenTelemetry tracing work landing in 7.3.0. Filed because this has grown past what a single issue tracks well: one review issue has produced three distinct workstreams, a second issue needs a design decision before it can be implemented, and a third gates the release.

## Which package ships what

This distinction keeps getting conflated, so stating it once: nearly all of this work is in **`RabbitMQ.Client`** itself - `RabbitMQActivitySource`, `SessionBase`, `Connection`, `AutorecoveringConnection` - and therefore ships on 7.3.0's schedule.

`RabbitMQ.Client.OpenTelemetry` is versioned separately (its own `otel-` MinVer prefix) and gated by #1728 on the `OpenTelemetry-1.0.0` milestone. Only a small part of the work below touches it: the null-`Headers` extractor fix (done) and the README trace-structure documentation (open).

So "7.3.0" and "OpenTelemetry 1.0.0" are not one bucket, and an item's milestone follows the assembly it changes.

## Children

| Item | What | Status |
|---|---|---|
| #1967 | Full review of the tracing implementation, and group A (behavioural defects) | Review done; group A in #1978, CI green |
| #1980 | Group B: semantic-convention conformance - span kind for `receive`, the `delivery_tag` attribute name, `messaging.destination.name` format, span naming, `fetch (empty)`, Opt-In size attributes | Open; every item changes emitted output, four break existing test assertions |
| #1981 | Group C: per-provider tracing configuration. `TracingOptions`, `ContextInjector`, `ContextExtractor` and `UseRoutingKeyAsOperationName` are public settable statics, so configuration is process-global and last-writer-wins across `TracerProvider`s | Open; blocks #1923 |
| #1970 | Tracing activities for topology and channel recovery | Open, needs a design decision on span granularity and parentage |
| #2008 | Let consumers opt an already-recorded exception out of `SetActivityError` double-recording, via `Exception.Data` | Open, needs a design decision; group A follow-up from #1978 review |
| #1923 | Ship API changes | Open; the actual release gate |
[...]

PR #2009, opened 2026-09-02
Fix process-global OpenTelemetry tracing configuration ownership

lukebakken:
> [!NOTE]
> This pull request was prepared by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who reviewed it before filing. The code was written and verified jointly. The design direction - moving configuration ownership onto the connection, separating that from provider subscription, and making the connection layer an override type rather than reusing the shipped one - came from @danielmarbach's reviews, with @paulomorgado independently arguing for the same shape.

## Summary

Gives tracing configuration an owner: the connection that performs the traced operation. Addresses the remaining half of #1981; group C of the 7.3.0 OpenTelemetry work in #1979. The uncontroversial half - the `ArgumentNullException` guards and honest documentation of process-wide scope - shipped separately in #2026, which is a prerequisite and already merged.

`AddRabbitMQInstrumentation` wrote process-global statics on `RabbitMQActivitySource` inline, so two `TracerProvider`s clobbered each other and disposing one restored nothing.

Per-`TracerProvider` configuration, which is how #1981 was originally framed, cannot exist: one `ActivitySource` produces a single `Activity` shared by every listener, and one publish injects a single set of headers, so neither span shape nor propagated context can differ per provider. The owner is the connection.

## What this adds

```csharp
var factory = new ConnectionFactory();
factory.TracingOptions = new ConnectionTracingOptions
{
    UseRoutingKeyAsOperationName = false,
    Propagator = new OpenTelemetryPropagator(),   // RabbitMQ.Client.OpenTelemetry
};

// or, equivalently, letting the integration package install the bridge
factory.UseOpenTelemetryTracing(options => options.UseRoutingKeyAsOperationName = false);
```

**Every member is nullable, and resolution is per member.** `null` inherits the layer below, so setting one thing leaves the others alone.

**The extension point is a propagator, not a delegate pair.** `ConnectionTracingOptions.Propagator` is a `System.Diagnostics.DistributedContextPropagator`, `null` meaning `DistributedContextPropagator.Current`. `RabbitMQ.Client.OpenTelemetry` ships `OpenTelemetryPropagator` as the bridge to OpenTelemetry's own propagator and `Baggage`.

That is the point @danielmarbach raised and it is also what the OpenTelemetry .NET instrumentation guidance requires of a library: use OpenTelemetry's `TextMapPropagator` "only when the built-in propagator cannot do the job, and only at the application root - never in libraries", because "libraries stay on `DistributedContextPropagator` regardless - bridging is an application-root concern". The core client was already compliant, propagating through `DistributedContextPropagator.Current` with no OpenTelemetry dependency. What was not compliant was exposing `ContextInjector`/`ContextExtractor` as *the* public propagation seam, since those exist precisely so an OpenTelemetry-shaped propagator can be substituted. **The shipped seam was the defect, not just its fallback semantics.**

`RabbitMQActivitySource.ContextInjector` and `ContextExtractor` are therefore now `[Obsolete]` warnings pointing at `Propagator`. They keep working; removal waits for a major. Deprecating them was previously blocked only because no replacement existed.

## Why a new type rather than changing the shipped one

Earlier revisions of this description weighed four shapes, three of which assumed the connection layer had to reuse `RabbitMQTracingOptions` and therefore needed `bool` to become `bool?`. @danielmarbach pointed out the premise was wrong: `v7.2.0`'s `ConnectionFactory` had no tracing members at all, so the connection layer is entirely unshipped and unconstrained by compatibility.

The two layers also want different things. The shipped type is a **defaults** type - complete, standalone, delegates pre-seeded. The connection layer is an **override** type - partial, `null` inherits, not usable standalone. One type playing both roles is exactly what produced the pre-seeding that caused the propagation loss described below. So `RabbitMQTracingOptions` is untouched, and its unshipped instance delegates become `internal` (they remain the storage the deprecated statics read and write). Nothing shipped breaks.

Naming follows house style: `CreateChannelOptions` is already an override type in behaviour - `ushort? ConsumerDispatchConcurrency` means "inherit from the connection" - and does not encode that in its name.

## The defect this replaces

The previous revision of this branch resolved a connection's configuration as `tracing ?? s_tracingOptions` - **whole-object** - and a fresh `RabbitMQTracingOptions` pre-seeds the *client's* built-in delegates rather than inheriting. So this, a span-shaping tweak and exactly what the branch's own documentation told users to write:

```csharp
cf.TracingOptions = new RabbitMQTracingOptions { UsePublisherAsParent = false };
```

switched that connection off OpenTelemetry propagation altogether. Messages went out with `traceparent` and no `baggage`, `Propagators.DefaultTextMapPropagator` was never consulted so a B3 or Jaeger propagator was silently ignored, and the deliver path stopped resetting `Baggage.Current` per message. Spans still exported, so nothing looked wrong. The same cause made the factory and the builder mutually exclusive: configure a factory, and a `configure` action passed to the builder was silently ignored for its connections.

## Verified before choosing this shape

Each of these could have sunk the design, so none is assumed:

| Question | Answer |
|---|---|
| Is `DistributedContextPropagator` subclassable, and `Current` assignable, on **.NET Framework 4.8**? | Yes, at `DiagnosticSource` 9.0.4 and 10.0.0 both |
| Does a bridge reproduce the current OpenTelemetry behaviour end to end? | Yes - header names, baggage value, consumer-side `Baggage.Current` and publish-to-consume trace linkage all identical against a live broker |
| Does `ActivityContext` -> traceparent string -> `TryParse` lose anything? | No, including `tracestate` and the `Recorded` flag |
| Does `Propagator` need a `DiagnosticSource` bump? | No. `CreateW3CPropagator` is 10.x-only, but subclassing is not |
| Is `Propagators.DefaultTextMapPropagator.Fields` safe to read? | **No** - it is a no-op propagator returning `null` when only `OpenTelemetry.Api` is referenced, so the bridge guards it |

`Baggage.Current` is OpenTelemetry state the core client cannot reach, so the bridge sets it inside `ExtractTraceIdAndState` - a call the client already makes per delivery - which is also where the header-less reset #1967 needed lives. A propagator installed here must therefore tolerate a null carrier; that is documented on the member.

## Findings from the previous review, fixed here

- **Resolve once per operation.** The span methods now take an already-resolved `ResolvedTracingOptions`, so resolving twice is not expressible. The publish path used to resolve once for the span name and again for the injector, which lets one span take its name from one configuration and its propagated context from another.
- **No `Session.Connection` walk.** `Channel.TracingOptions` is a readonly property assigned in the constructor from `CreateChannelOptions`, which already copies the values a channel needs out of `ConnectionConfig`. The walk put an unguarded dereference on the deliver path and could read `Session` before the constructor had assigned it.
- **The process-wide install is one reference assignment**, not an instance assignment followed by two more writes into live shared state.
- **`Clone()` is gone** from this path, removing the hand-copied member list and its subclass truncation.

## Coverage

`TestTracingOptionsResolution` pins the resolution matrix, **verified by two mutations rather than one**: the flags and the delegates fall back independently, so no single mutation reaches both. Making the flags whole-object fails `PropagatorOnlyKeepsProcessWideFlags` and `OneFlagDoesNotResetTheOther`; making the delegates whole-object fails `FlagOnlyKeepsProcessWideDelegates`.

`ResolvingTwiceReusesTheSamePropagatorAdapter` earned its place by failing against a first implementation that cached the adapter but still bound a fresh delegate pair on every resolve - which is the publish and delivery path.

`TestOpenTelemetry` now decorates the propagator where it used to wrap the injector delegate, which is the migration a caller actually performs.

Local runs on net8.0: 223/223 unit; 94 passed and 3 skipped across `SequentialIntegration` against a live broker.

Note also that this branch's CI was red before this work, unrelated to the design: the rebase onto `main` left #2036's `TestConsumerDispatchConcurrency` constructing `ConnectionConfig` without the `tracingOptions` argument this branch adds. Fixed here.

## Related

#2040 - installing `RabbitMQ.Client.OpenTelemetry` already lifts an application to `DiagnosticSource` 10.x, which changes the baggage header the core client emits. Independent of this change: no code here causes or fixes it, and it needs the headers characterised plus a release note.

## Not covered

A **per-connection** override does not exist yet - `CreateConnectionOptions.Tracing`, as @danielmarbach sketched. The design composes for it (`connection ?? factory ?? process-wide`, the last term dropping out when the statics are removed), but only the two layers that exist today are implemented and tested.

`IConnectionFactory` gains nothing here. It declares every comparable per-connection knob, but adding to a public interface breaks external implementers and `ConnectionFactory` is `sealed`, so that waits for a major - @danielmarbach's recommendation.

- Tracing configuration can now be owned by a connection. Set `ConnectionFactory.TracingOptions` to a `ConnectionTracingOptions`, whose members are nullable and inherit the process-wide default per member. Propagation is configured with `ConnectionTracingOptions.Propagator`, a `System.Diagnostics.DistributedContextPropagator`; `RabbitMQ.Client.OpenTelemetry` supplies `OpenTelemetryPropagator` for OpenTelemetry propagation and `Baggage`. `RabbitMQActivitySource.ContextInjector` and `ContextExtractor` are deprecated in favour of it and will be removed in a future major version. See `v7-MIGRATION.md`.
- Separately, and not caused by this change: `RabbitMQ.Client.OpenTelemetry` requires `System.Diagnostics.DiagnosticSource` 10.x, whose default propagator emits baggage in the `baggage` header, while the core client alone resolves 9.x and emits the non-standard `Correlation-Context` header. Set `DistributedContextPropagator.Current` explicitly at the application root if the header name matters. Tracked as #2040.


lukebakken:
> [!NOTE]
> This comment was drafted by Claude (Anthropic's Claude Code) under the direction of @lukebakken.

cc @eerhardt @paulomorgado - given your OpenTelemetry .NET work, I would appreciate your eyes on this one. It reworks how the client's tracing configuration is owned: process-wide, reference-counted to `TracerProvider` lifetime, rather than the previous last-writer-wins statics. The reasoning is in the description and in #1981.

gfoidl (dismissed):
LGTM.

danielmarbach (dismissed):
I understand why per-"TracerProvider" configuration is not really achievable here, since the RabbitMQ operation itself has no association with a specific "TracerProvider".

What I am less convinced about is that this necessarily means the configuration itself has to remain static.

It seems to me that the underlying issue is ownership: "AddRabbitMQInstrumentation(...)" is configuring RabbitMQ runtime behavior through "TracerProviderBuilder", even though the actual publish/consume operation belongs to a RabbitMQ connection/channel and has no knowledge of the provider that configured it.

Would it be possible to move the mutable tracing configuration to the RabbitMQ connection or connection factory instead? A connection could snapshot the tracing options, injector, and extractor when it is created and pass that configuration into the instrumentation code.

The existing static APIs could remain functional for compatibility, while a new instance-owned configuration API becomes the preferred path. I would actually be in favor of marking the old API "[Obsolete]" with a warning in a minor release once that replacement exists. That gives users early guidance towards the better API without breaking binary compatibility or removing functionality. The actual removal could then happen at a major-version boundary.

That seems preferable to adding locking, reference counting, and restoration logic around global mutable state if we can instead introduce a configuration model with a more appropriate owner and provide a gradual migration path towards it.

lukebakken:
Thanks @danielmarbach, I didn't really have an opinion so I'm glad you do 😸

danielmarbach:
> Thanks @danielmarbach, I didn't really have an opinion so I'm glad you do 😸

That being said my changes requested is more a friendly I think we should do a detour. Unfortunately my capacity is limited these days so all I can provide is opinions and reviews when I have some time

lukebakken:
> Unfortunately my capacity is limited these days so all I can provide is opinions and reviews when I have some time

Greatly appreciated!

paulomorgado:
I prefer to have what goes into the `Activity` and activity creation site (despite using whatever configuration) over a per-process whatever.

That being said, there could be a process-wide static default, that could be overridden by each connection (or connection factory),

I would keep OTel as much as possible a consumer of logs, traces (`Activity`), and metrics.

In theory, OTel is not necessarily the only consumer.

That being said, even not being fully settled, the semantic naming conventions seem to be the most solid and should be follwed.

lukebakken:
> [!NOTE]
> This comment was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who reviewed it before posting. The implementation and test claims were verified against the branch.

Thanks Daniel, this is the right framing and we adopted it. The reference-counted-statics approach is gone; the configuration now has a per-connection owner.

Mapping to your points:

- **Ownership moved off the statics.** `ConnectionFactory.TracingOptions` (a `RabbitMQTracingOptions`) carries the span-shaping options and the inject/extract delegates. A connection snapshots it at creation into `ConnectionConfig`, and it flows to the `RabbitMQActivitySource` span methods through the channel (lazily via `Session.Connection`). The publish/consume path reads the connection's own configuration, not process-wide state.
- **Statics stay functional for compat.** When a factory sets no `TracingOptions`, operations fall back to the existing process-wide default, read live. No locking, reference counting, or restoration logic; that machinery is removed.
- **`[Obsolete]` on the old path.** The `RabbitMQActivitySource` statics and the two `AddRabbitMQInstrumentation` overloads that configure them are now `[Obsolete]` (warning, not error), pointing at the new `AddRabbitMQInstrumentation(TracerProviderBuilder, ConnectionFactory, ...)` overload. Binary compatibility is preserved; removal can wait for a major bump.

One back-compat subtlety worth flagging: because the delegates now live on the same options object as the flags, assigning the static `TracingOptions` copies the span-shaping options into the existing instance rather than swapping the reference, so a previously configured injector/extractor is preserved (it used to be an independent static). That behavior is pinned by characterization tests.

Reshaped history is pushed. Thanks again for steering this.

danielmarbach (commented):
I really like this direction. Per-connection ownership makes the model much clearer.

One API ergonomics question I still have is how this is expected to work with the generic host / DI pattern.

For example, I can easily imagine the `ConnectionFactory` being registered like this:

```csharp
services.AddSingleton<ConnectionFactory>(sp =>
{
    // create/configure factory using IConfiguration etc.
    return new ConnectionFactory(...);
});
```

while OpenTelemetry is registered separately:

```csharp
services.AddOpenTelemetry()
    .WithTracing(builder =>
    {
        builder.AddRabbitMQInstrumentation(...);
    });
```

At that point the `TracerProviderBuilder` configuration doesn't have access to the `ConnectionFactory` instance because the factory is only resolved later by DI.

That makes me wonder whether there should eventually be a stronger separation between the two responsibilities:

```csharp
factory.UseOpenTelemetryTracing(...);
```

to install the propagation configuration on the thing that owns it, and:

```csharp
builder.AddRabbitMQInstrumentation();
```

to just subscribe the `TracerProvider` to the RabbitMQ `ActivitySource`s.

That would also make the ownership model introduced by this PR very explicit: the factory controls how RabbitMQ creates/propagates tracing information, while OpenTelemetry independently decides whether those activities are observed/exported.

The new overload taking both `TracerProviderBuilder` and `ConnectionFactory` could still be a convenient shortcut when both objects are readily available, but requiring the factory there seems somewhat awkward for the typical HostBuilder + DI scenario.

lukebakken:
> [!NOTE]
> This comment was drafted by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who reviewed it before posting. The measurements and API-surface claims were verified against the branch and the shipped `PublicAPI` files.

@danielmarbach — a compatibility question for you when you have a few minutes. Short version: making per-connection tracing configuration actually work wants a small breaking change to a type that shipped in 7.2.0, and @lukebakken is inclined to treat the current shape as the bug. We'd like your read before committing to that.

## How we got here

I implemented the separation you suggested — `factory.UseOpenTelemetryTracing(...)` to install propagation on the owner, `builder.AddRabbitMQInstrumentation()` to subscribe, with the combined overload kept as a shortcut. An adversarial review then found it silently broken on the path its own documentation recommended, confirmed against a live broker:

```csharp
cf.TracingOptions = new RabbitMQTracingOptions { UsePublisherAsParent = false };
```

That is a span-shaping tweak, but it switched the connection off OpenTelemetry propagation altogether. Messages went out with `traceparent` and no `baggage` header, `Propagators.DefaultTextMapPropagator` was never consulted so a B3 or Jaeger propagator was silently ignored, and the deliver path stopped resetting `Baggage.Current` per message. Spans still exported, so nothing looked wrong from the outside.

The cause is that the fallback from the connection layer to the process-wide layer was **whole-object** (`tracing ?? s_global`), and a fresh `RabbitMQTracingOptions` pre-seeds the client's own built-in delegates rather than inheriting. The same cause makes the two configuration points mutually exclusive: configure a factory, and a `configure` action passed to the builder is silently ignored for that factory's connections.

## What it needs

Per-member fallback, each member `null` meaning "inherit the layer below". The open question is where those nullable members live, and that is where compatibility bites, because `RabbitMQTracingOptions` shipped in 7.2.0 with non-nullable `bool` properties.

The break, measured rather than assumed:

| Usage | Under `bool?` |
|---|---|
| `new RabbitMQTracingOptions { UseRoutingKeyAsOperationName = true }` | compiles |
| `options.UseRoutingKeyAsOperationName = flag;` | compiles |
| `bool x = options.UseRoutingKeyAsOperationName;` | `CS0266` |
| `if (options.UseRoutingKeyAsOperationName)` | `CS0266` |

Writes are unaffected; only reads break at source level, which for a write-mostly options bag should be rare. It is binary-breaking either way, since the getter's return type is part of its signature. `RabbitMQActivitySource.UseRoutingKeyAsOperationName` can stay `bool` by resolving `null` to the default, so that shipped member need not change.

## Four shapes

1. **`bool?` on `RabbitMQTracingOptions`.** One type, one member list, cleanest model. Breaking as above.
2. **A new options type for the connection layer**, shipped type stays as the process-wide default. Non-breaking, but two near-identical types forever and two member lists to keep in step — we already have a finding about exactly that drift in the existing `Clone()`.
3. **Nullable delegates only.** They are unshipped, so they can be `null`-means-inherit for free, while the bools keep whole-object semantics. Non-breaking, and it fixes the severe defect. Residual wart: a factory that sets options purely to add a custom injector silently reverts both span-shaping flags to `true`, even where the process-wide default had set them `false`.
4. **Four nullable properties directly on `ConnectionFactory`**, no options object at that layer. Non-breaking, per-member by construction, and consistent with how the factory exposes every other knob. Costs four more properties on an already-large type and does not compose as neatly for capturing a snapshot.

## The questions

Does (1) seem reasonable for a type that has been public for about ten months across 7.2.0, 7.2.1 and 7.2.2 in the core `RabbitMQ.Client` package, given that only reads break? Or does the binary break argue for (4), which gets the same per-member semantics without touching shipped API?

Second, smaller, if you have a view: `IConnectionFactory` declares every other comparable per-connection knob (`ConsumerDispatchConcurrency`, `ContinuationTimeout`, `HandshakeContinuationTimeout`). Adding one there breaks external implementers, and `ConnectionFactory` is `sealed`, so it may want to wait for a major — but leaving it off means interface-typed consumers cannot reach the new configuration at all.

For context on timing: the ownership model is still targeted at 7.3.0, so this PR is not parked. What we have split out is a smaller prerequisite PR carrying just the `ArgumentNullException` guards and honest documentation of the process-wide scope, so that those land regardless and this PR narrows to the new API alone. No rush on your side though — if it takes a while to settle, the fallback is to ship the prerequisite and let the ownership model follow in the next release rather than rush a third API design.

danielmarbach:
@lukebakken short version: I agree the current shape is the bug, but all four options assume the factory layer has to use the same `RabbitMQTracingOptions`. I don't think it does.                                                                                      
                                                                                                                                                                           
Checked against `v7.2.0`: `ConnectionFactory` had no tracing members at all. What shipped is the five-line options type (two bools, no delegates), the `RabbitMQActivitySource` statics, and the two builder-only overloads. The factory layer is entirely unshipped, so its type is not constrained by compat at all.                        

The two layers also want different things:                                                                                                                              
                                                                                                                                                                           
- The shipped type is a defaults type: complete, standalone, delegates must pre-seed.                                                                                   
- The factory layer is an override type: partial, `null` means inherit below, and it should not be usable standalone.                                                                                                                                             
                                                                        
The pre-seeded delegates that caused the bug exist because one type plays both roles. An override type can't have that failure mode, and `cf.TracingOptions = new RabbitMQTracingOptions { ... }` stops compiling instead of silently killing propagation.
                                     
I'd also question option 2's "two near-identical types forever" costing. The deprecated statics are slated for removal at a major, and the shipped type retires with them. The two types coexist for the deprecation window, not forever.                                                                                                                  
                                                                                                                                                                           
So I'm leaning towards option 2 (or 4): same per-member semantics, nothing shipped breaks, and migration fails at compile time. Option 1 keeps one name, which I'd call a preference rather than a necessity, and it's the only shape that breaks anything.                                                                      

On `IConnectionFactory`: wait for the major. The concrete type dominates real usage, and the break would land on test doubles. Unless you know of external implementers I'm missing?

danielmarbach:
One more design thought while we are shaping this: factory defaults and per-connection overrides are not competing options, they compose.

```csharp
// factory holds the defaults, what this PR adds
factory.TracingOptions = ...;

// a future CreateConnectionOptions can override per connection
await factory.CreateConnectionAsync(new CreateConnectionOptions
{
    Tracing = ..., // null means inherit the factory
    ClientProvidedName = ...,
});
```

Resolution becomes `connection ?? factory ?? process-wide`, and the last term drops out once the statics are removed.

We already have this pattern in the codebase. `CreateChannelOptions.ConsumerDispatchConcurrency` is `ushort?` where null means "use `IConnectionFactory.ConsumerDispatchConcurrency`", and `CreateOrUpdate` merges the connection config at channel creation. So this is house style, not a new mechanism.

If overrides exist, the factory layer still matters:

- Bare `CreateConnectionAsync()` and automatic recovery need a default that is not the process-wide statics. The factory is what lets the statics die at the major.
- `UseOpenTelemetryTracing` needs one writable place to install the delegates.
- Tracing policy gets set once where the factory is registered, not at every call site.

A `CreateConnectionOptions` would also give the six `CreateConnectionAsync` overloads a home, so it earns its place beyond tracing.

danielmarbach:
One more angle, this time about the direction of the API rather than the implementation.

I do not think the ownership model needs another rework. The substrate in this PR looks right. What I would question is whether delegates are the extension point we want to bless on the new factory surface.

There is a strong BCL-first path here: libraries can stay on `System.Diagnostics` and use `DistributedContextPropagator` for propagation, without depending on `OpenTelemetry.Api`. `HttpClient` already works this way.

OpenTelemetry still has its own `Baggage` and `TextMapPropagator` abstractions, so there are effectively two propagation worlds today. [open-telemetry/opentelemetry-dotnet#5667](https://github.com/open-telemetry/opentelemetry-dotnet/issues/5667) calls out that split and proposes bridging the OTel propagator through `DistributedContextPropagator.Current`, which would let instrumentation libraries stay on the BCL abstraction.

That suggests an API like:

```csharp
factory.TracingOptions = new RabbitMQTracingOptions
{
    UseRoutingKeyAsOperationName = true,
    Propagator = myPropagator
};
```

with `null` meaning `DistributedContextPropagator.Current`.

A custom `DistributedContextPropagator` covers the same basic inject/extract scenarios as the delegate pairs, and making it configurable per factory already gives us more control than `HttpClient`, which uses the process-wide `Current`.

I initially thought this would require .NET 10. It does not. The `System.Diagnostics.DiagnosticSource` 10.x package carries the relevant propagator APIs for our existing TFMs as well.

That is the part that makes me hesitate about shipping the delegates now. If `Propagator` is where we think the API should eventually end up, then blessing the delegates first creates another public extension point that we may later want to deprecate.

The OpenTelemetry integration can still work with that model. It can provide a `DistributedContextPropagator` bridge to `Propagators.DefaultTextMapPropagator` and handle the OTel-specific baggage behavior there. That package is also the natural place for the `OpenTelemetry.Api` dependency.

There are still good arguments for keeping the delegates: they are already implemented, they are flexible, and `DistributedContextPropagator` is not especially pleasant to use directly. The two models could also coexist, with delegates remaining the escape hatch.

So I am not arguing that the implementation in this PR is wrong or that we should rework it again. I am only asking whether the public extension point should already be propagator-shaped before we commit to the delegate API.

Separately, the `DiagnosticSource` 10.x upgrade probably deserves its own issue. When the 10.x package is used, `DistributedContextPropagator.Current` [defaults to the W3C propagator rather than the legacy one](https://learn.microsoft.com/mt-mt/dotnet/core/compatibility/core-libraries/10.0/default-trace-context-propagator?), which changes baggage propagation from `Correlation-Context` to `baggage`. Since that implementation is carried by the package across our supported TFMs, I think we should characterize the emitted headers explicitly and call the change out in the release notes.

lukebakken:
> [!NOTE]
> This comment was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who reviewed it before posting. The measurements cited were taken against this branch and a live broker; the .NET Framework results come from running a probe on Windows 11.

@danielmarbach - thank you, all three of your points changed the outcome, and the code is now on the branch rather than a description of intent. The PR body is rewritten; the short version:

**Your propagator point was not a nice-to-have and I had it wrong at first.** The extension point is now `ConnectionTracingOptions.Propagator`, a `DistributedContextPropagator`, with `RabbitMQ.Client.OpenTelemetry` shipping `OpenTelemetryPropagator` as the bridge. The delegates are not part of the new surface, and the two shipped statics are `[Obsolete]` warnings pointing at it. Reading the OpenTelemetry .NET instrumentation guidance properly - "never in libraries", "libraries stay on `DistributedContextPropagator` regardless" - it is not that the delegates were a second-best API, it is that the seam itself was the defect. The core client was already compliant; only the public seam was not.

**You were right that the premise of my four options was wrong.** `v7.2.0`'s `ConnectionFactory` had no tracing members, so the connection layer is unshipped and unconstrained. `ConnectionTracingOptions` is the override type - partial, `null` inherits per member - and `RabbitMQTracingOptions` is untouched as the defaults type. Nothing shipped breaks, and the `bool?` question is moot.

**`IConnectionFactory`: waiting for the major, as you suggested.**

Four things worth your eye, two of which are questions:

1. **The name.** `ConnectionTracingOptions`, chosen to match house style rather than encode merge semantics - `CreateChannelOptions.ConsumerDispatchConcurrency` is already `ushort?`-means-inherit and the type is not called `Overrides`. It does read slightly oddly hanging off the *factory*. Your opinion would be welcome; renaming is cheap now and not later.

2. **`Propagator` needs no `DiagnosticSource` bump.** You noted the 10.x package carries the propagator APIs for our TFMs. Subclassing and assigning `Current` work at **9.0.4**; only `CreateW3CPropagator`/`CreatePreW3CPropagator` are 10.x-only. Verified on net8.0 and on .NET Framework 4.8 at both versions - that last one was the result that could have sunk the design.

3. **Where `Baggage.Current` ended up, because it is the one thing a BCL propagator cannot express.** The bridge sets it inside `ExtractTraceIdAndState`, a call the client already makes per delivery, which is also where the header-less reset from #1967 lives. The consequence is a documented contract: a propagator installed here must tolerate a null carrier. If you would rather that state were handled another way, this is the seam to argue about.

4. **Your composition sketch is not implemented.** `CreateConnectionOptions.Tracing` does not exist, so only the factory and process-wide layers are built and tested. The design composes for `connection ?? factory ?? process-wide` and the resolution is already per member, but I would rather say plainly that the third layer is untested than imply otherwise.

On your separate `DiagnosticSource` 10.x point - it is worse than a future concern, and it wants its own issue as you suggested. `OpenTelemetry.Api` 1.15.3 already depends on `DiagnosticSource >= 10.0.0`, so installing `RabbitMQ.Client.OpenTelemetry` lifts an application from 9.0.4 today. Measured on net8.0 and net472, the emitted baggage changes both header name and encoding:

| resolved `DiagnosticSource` | `Current` | emitted |
|---|---|---|
| 9.0.4 (core client alone) | `LegacyPropagator` | `Correlation-Context=region=eu-west-1, tenant=acme` |
| 10.0.0 (with our OTel package) | `W3CPropagator` | `baggage=region = eu-west-1, tenant = acme` |

CI is running on this push. It was red before this work for an unrelated reason - a rebase left #2036's test calling `ConnectionConfig` without the argument this branch adds - which is fixed here.

No rush from our side, and thank you again for the reviews - the shape is materially better than what it replaced.

lukebakken:
> [!NOTE]

Filed the `DiagnosticSource` 10.x point as #2040, @danielmarbach, with the measured header output for both versions on `net8.0` and .NET Framework 4.8. Linked from this PR's description as related rather than blocking, since nothing here causes or fixes it.

The part that made it worth its own issue rather than a release note: it is already live. `OpenTelemetry.Api` 1.15.3 depends on `DiagnosticSource >= 10.0.0`, so anyone using `RabbitMQ.Client.OpenTelemetry` today already gets 10.x behaviour from the core client, while this repository's own tests resolve 9.0.4 and pass - so nothing here exercises the combination users actually have. #2040 proposes characterising the emitted headers first, independent of whichever version we settle on.

projects/RabbitMQ.Client.OpenTelemetry/OpenTelemetryPropagator.cs
```diff
@@ -0,0 +1,125 @@
...
+            ActivityContext context = parentContext.ActivityContext;
+            if (context == default)
+            {
+                traceId = null;
+                traceState = null;
+                return;
+            }
+
+            // The client reparses this with ActivityContext.TryParse, so emit W3C traceparent form.
+            string flags = (context.TraceFlags & ActivityTraceFlags.Recorded) != 0 ? "01" : "00";
```
danielmarbach: Would it be better to read the enum value to make it slightly more future proof?

projects/RabbitMQ.Client.OpenTelemetry/OpenTelemetryPropagator.cs
```diff
@@ -0,0 +1,125 @@
...
+            traceId = $"00-{context.TraceId}-{context.SpanId}-{flags}";
```
danielmarbach, line 127: If I recall correctly, we don't have `string.Create` available because of Netstandard support, right?

projects/RabbitMQ.Client/RabbitMQ.Client.csproj
```diff
@@ -52,6 +52,9 @@
     <None Remove="icon.png" />
     <Content Include="icon.png" PackagePath="" />
     <None Include="README.md" Pack="true" PackagePath="/" />
+    <!-- First-party companion package: it constructs the process-wide tracing default, which has to
```
danielmarbach: This comment doesn't give me much value because we also don't have any justification for the "Common".

projects/Test/SequentialIntegration/TestOpenTelemetry.cs
```diff
@@ -430,5 +661,39 @@ private void AssertActivityData(bool useRoutingKeyAsOperationName, bool usePubli
                 AssertStringTagEquals(receiveActivity, RabbitMQActivitySource.MessageId, messageId);
             }
         }
+
+        /*
+         * Adds one extra header on inject and defers everything else, including baggage handling, to
+         * the propagator it wraps. This is the shape of a caller extending propagation under the
+         * ConnectionTracingOptions.Propagator API.
+         */
+        private sealed class MarkerAddingPropagator : DistributedContextPropagator
```
danielmarbach, line 670: Nice!

danielmarbach (dismissed):
Looks solid to me

lukebakken:
> [!NOTE]
> This comment was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who reviewed it before posting. Each measurement below names what was run.

Thanks @danielmarbach - all four addressed.

## Trace flags: adopted, with a correction to my own reasoning

You were right, and for a slightly different reason than "future proof". `ActivityContext.TryParse` round-trips the *whole* flags byte - checked at `DiagnosticSource` 9.0.4 with `00`, `01`, `03` and `ff`, all four parsed successfully and all four preserved. So singling out `Recorded` was the bridge discarding something the platform was willing to carry.

The correction, because it narrows this a long way: it is **not** reachable through the default path. `TraceContextPropagator.Extract` masks `03` and `ff` down to `01` before the bridge ever sees them, measured against `OpenTelemetry.Api` 1.15.3. So the loss only appears behind a custom `TextMapPropagator` - which is exactly the case this type exists to serve, so it still seems worth fixing, but I do not want to claim it fixes a live defect on the common path.

On cost, since this is a per-delivery path rather than a cold one: `ExtractTraceIdAndState` runs once per delivery, behind `s_subscriberSource.HasListeners()`, and `Inject` once per publish. My first version formatted the byte unconditionally, which traded an interned literal for a 2-char allocation on every traced delivery - very slightly worse, not better. It now keeps the literal for the two values W3C defines and falls back to formatting only for anything else:

```csharp
string flags = context.TraceFlags switch
{
    ActivityTraceFlags.None => "00",
    ActivityTraceFlags.Recorded => "01",
    _ => ((int)context.TraceFlags).ToString("x2")
};
```

Lossless, and allocation-free for every value reachable today.

## `string.Create`: correct, netstandard2.0 rules it out

Compiled a scratch project against `netstandard2.0;net8.0`: `error CS0117: 'string' does not contain a definition for 'Create'` on netstandard2.0, clean on net8.0. This package targets both, so the interpolation stays. I have not measured a `string.Concat` variant, on the grounds that the `Activity` allocation and the `TryParse` on the other side dominate a single interpolation - say the word if you would rather see numbers.

## The csproj comment: deleted

Your instinct was right, and the duplication makes it worse than merely unhelpful: the same rationale already sits where it is actually enforced, in `AddRabbitMQInstrumentation`. Repeating it in the csproj put a claim about behaviour somewhere nothing can check it, next to entries that carry no justification of their own.

The `InternalsVisibleTo` itself stays. I checked whether the redesign had made it unnecessary - removing it fails the build on both target frameworks, because the package assigns the internal `RabbitMQTracingOptions.ContextInjector` and `ContextExtractor`.

## Your generic host / DI point

Filed as #2042 and assigned to us both, rather than answered here, since it is a public API decision that deserves its own thread. You are right that at `WithTracing(builder => builder.AddRabbitMQInstrumentation(...))` time there is no `ConnectionFactory` to configure, so the two-object overload fits a script better than it fits `HostBuilder`. The split you proposed - `factory.UseOpenTelemetryTracing(...)` to install propagation on the owner, `builder.AddRabbitMQInstrumentation()` to only subscribe the provider - reads as the right shape on top of the ownership model this PR establishes, and the issue records it in your words along with the questions it opens: whether the two-object overload survives, whether any ordering of the two calls can silently do nothing, `IConnectionFactory` versus `ConnectionFactory` as the receiver, and whether the same split should apply to the process-wide default.

PR #2011, opened 2026-09-03
Honor the caller's connection close timeout and bound every abort

### LINKED ISSUE #1973: CloseAsync silently raises a caller-supplied close timeout to 30s, capping Timeout.InfiniteTimeSpan
### Describe the bug

`Connection.CloseAsync` silently raises any caller-supplied non-abort close timeout below 30s to 30s, and any abort timeout below 5s to 5s:

https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/main/projects/RabbitMQ.Client/Impl/Connection.cs#L334-L343

```csharp
if (abort && timeout < InternalConstants.DefaultConnectionAbortTimeout)
{
    timeout = InternalConstants.DefaultConnectionAbortTimeout;
}

if (false == abort && timeout < InternalConstants.DefaultConnectionCloseTimeout)
{
    timeout = InternalConstants.DefaultConnectionCloseTimeout;
}
```

This appears to be unintentional, and it has two consequences beyond ignoring the caller.

### 1. `Timeout.InfiniteTimeSpan` is capped at 30s

`IConnectionExtensions` documents infinite waiting:

https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/main/projects/RabbitMQ.Client/IConnectionExtensions.cs#L59

> To wait infinitely for the close operations to complete use `System.Threading.Timeout.InfiniteTimeSpan`.

`Timeout.InfiniteTimeSpan` is `-1` ticks, so `timeout < DefaultConnectionCloseTimeout` is **true** and it is clamped to 30s. The documented behaviour is currently unreachable.

Before the floor this worked: `new CancellationTokenSource(Timeout.InfiniteTimeSpan)` is accepted and its token never fires, which is exactly "wait forever".

### 2. It silently defeated the fix for #1759

#1759 ("Disposing Connection after closing it with timeout causes deadlock") was fixed in a3c4ec557, whose commit message reads:

> When passing a timeout of 0, `DisposeAsync` would block forever after closing a connection.

Its regression test calls `CloseAsync(TimeSpan.Zero)`:

https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/main/projects/Test/Integration/GH/TestGitHubIssues.cs#L145

The floor turns that `TimeSpan.Zero` into 30s, so the test no longer exercises the zero-timeout path it was written to protect.

Note the test still **passes** rather than failing, which is why this went unnoticed: a healthy connection closes in ~175ms, so no close timeout is ever reached and the clamp is invisible. It is vacuous with respect to its own scenario.

### Why this looks unintended

The two constants were introduced in 4042bc148 (Aug 2021) purely as **defaults for the parameterless overloads** - `Close(connection)` passed 30s, `Abort(connection)` passed 5s. A caller supplying an explicit timeout got exactly that value; there was no clamping.

Both floor clauses were added later, in 9a8d57e18 (PR #1809, "Address `ObjectDisposedException`", fixes #1802). That change was about an `ObjectDisposedException` on a `SemaphoreSlim` during shutdown. Neither the issue, the PR body, the commit message, nor the PR discussion mentions close timeouts, clamping, or a minimum.

The same diff changed `_mainLoopTask.WaitAsync(timeout, cts.Token)` to `WaitAsync(cts.Token)`, moving the timeout into the linked `CancellationTokenSource`. The floors read as defensive scaffolding for that refactor rather than a deliberate policy that callers may not close faster than 30s.

The no-timeout overloads already pass the 30s constant explicitly, so the floor is redundant for the default path and only ever affects callers who deliberately chose a smaller value.

### Note: the floor cannot simply be deleted

It does incidentally prevent a crash, so removing it needs explicit handling of the edge cases rather than a straight revert:

* `new CancellationTokenSource(negative)` throws `ArgumentOutOfRangeException`, except for `InfiniteTimeSpan` which is accepted and never fires.
* `TimeSpan.Zero` produces an already-cancelled token.

So `InfiniteTimeSpan` should mean no timeout, and zero/negative values need a defined behaviour (either honoured as "expire immediately" or rejected as an argument error). Worth deciding whether `TimeSpan.Zero` should throw `ArgumentOutOfRangeException` or close immediately - #1759's test implies the latter was intended.

### Relationship to #1968

This is why the flaky test in #1968 waits ~30s despite passing a 6s `_waitSpan`. The floor is not the cause of that flake - a bare `OperationCanceledException` from `TaskExtensions.DoWaitAsync` is - but it widens the window considerably. Filing separately because the two problems above are independent of that net472 flake, and the `InfiniteTimeSpan` cap is a public-API contract break.

### Reproduction

The clamp is arithmetic and needs no broker:

| requested | effective |
| --- | --- |
| `Timeout.InfiniteTimeSpan` | 30s |
| `TimeSpan.Zero` | 30s |
| 6s | 30s |
| 30s | 30s |
| 60s | 60s |

---

*Investigated with the assistance of Claude Code. The git history and exception behaviour above were verified against the repository and by running the cases.*

lukebakken:
> [!NOTE]
> This pull request was prepared by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who reviewed it before filing. The code was written and verified jointly. The decision to cap every abort, and to keep an over-large timeout bounded rather than unbounded, were @lukebakken's calls.

## Summary

Fixes #1973. `Connection.CloseAsync` raised any graceful close timeout below 30 seconds to 30 seconds, and any abort timeout below 5 seconds to 5 seconds. As #1973 establishes, those floors were not policy: they arrived in `9a8d57e18` (PR #1809) as incidental scaffolding while fixing an unrelated `ObjectDisposedException`, and neither that issue, PR, commit message nor discussion mentions close timeouts.

A caller's timeout is now honoured, and an abort is always bounded.

## Behaviour

| Call | Before | After |
|---|---|---|
| `CloseAsync(TimeSpan.Zero)` | 30s | 1s (the minimum) |
| `CloseAsync(TimeSpan.FromSeconds(6))` | 30s | 6s |
| `CloseAsync(Timeout.InfiniteTimeSpan)` | 30s | unbounded |
| `CloseAsync(TimeSpan.FromDays(60))` | `ArgumentOutOfRangeException` | clamped to ~24.86 days |
| `AbortAsync(Timeout.InfiniteTimeSpan)` | 5s | 10s |
| `AbortAsync(TimeSpan.FromSeconds(30))` | 30s | 10s |
| `AbortAsync(TimeSpan.MaxValue)` | `ArgumentOutOfRangeException` | 10s |
| `CloseAsync(TimeSpan.FromMilliseconds(-2))` | 30s | 1s (the minimum) |

### Why an abort is capped

An abort is best-effort teardown that never throws, so its entire value to a caller is that it returns promptly. Honouring an arbitrarily large abort timeout defeats that, and the previous shape produced a five-order-of-magnitude split between two spellings of "as long as it takes": `AbortAsync(TimeSpan.MaxValue)` would have waited roughly 49.7 days while `AbortAsync(Timeout.InfiniteTimeSpan)` waited 5 seconds. An abort now resolves into `[5s, 10s]`, honouring anything inside that band.

### Why an over-large timeout is clamped, not unbounded

Only an explicit `Timeout.InfiniteTimeSpan` produces a wait with no local bound, because the escape hatches from such a wait are thinner than they appear. All three were checked:

- `CloseAsync`'s `finally` calls `MaybeTerminateMainloopAndStopHeartbeatTimers(cancelMainLoop: false)`, which stops **both** heartbeat timers before the wait begins — so the read timer that would otherwise call `FinishCloseAsync` after three missed beats is already gone.
- `NetworkStream` ignores `ReadTimeout` for asynchronous reads, and the main loop uses `PipeReader.ReadAsync`.
- What remains is the broker: with our heartbeat writer stopped it stops hearing from us and drops the TCP connection after ~2× the negotiated interval, the read fails, and the main loop unwinds. That is roughly 120–180s at the default 60s heartbeat — **and it never happens with `RequestedHeartbeat = TimeSpan.Zero`.**

So `TimeSpan.FromDays(60)` is clamped: someone writing that has not opted into a wait only the broker can end. Capping the abort path is what makes clamping monotonic, since the two paths no longer share the branch.

## Simplification

Because exceeding the ceiling is no longer an error, the per-target-framework ceiling is gone. `s_maxCancellationTokenSourceDelay` is now the lower of the two runtime limits unconditionally — no `#if NETSTANDARD`, and no dependence on which build the application resolved. A caller asking for 30 days gets ~24.86 days on every build instead of 24.86 or 49.7 days depending on the target, which removed roughly twenty lines of documentation across five files. The constant also moved beside the two sibling timeouts `ResolveCloseTimeout` compares against, which incidentally stops `Connection` gaining a static constructor it did not previously have.

## #1759's regression test is no longer vacuous

This is the second consequence #1973 reports, and the concrete payoff. `DisposeWhileCatchingTimeoutDeadlocksRepro_GH1759` closes with `TimeSpan.Zero`; the floor turned that into 30 seconds, and since a healthy connection closes in ~20ms the zero-timeout path it guards was never taken. It passed for the wrong reason.

Asserting on the exception or on `IsOpen` does not fix that — `IsOpen` is `CloseReason is null`, set before the first cancellable await, so it reports `false` even when nothing was torn down. It now asserts the **shutdown-event sequence**, which is what separates a real close from a half-done one: exactly one event, `Application/200/Goodbye`. On a build without the minimum it fails, because the escaped close leaves channel 0 with a null `CloseReason` and `DisposeAsync` then transmits `channel.close` on channel number 0 — a connection-level violation the broker answers with a 504, arriving as a second shutdown event:

```
[Application/200/Goodbye | Peer/504/CHANNEL_ERROR - unexpected method in connection state running]
```

### Why there is still a minimum

Honouring `TimeSpan.Zero` literally leaked the connection. The timeout is linked into the tokens passed to `SetSessionClosingAsync` and the `connection.close` transmit, not only the wait for the peer's reply, and the `if (false == abort) throw;` that handles a cancelled handshake sits inside the `else` block — before the teardown `try`. So the main loop was never awaited, the socket never closed, and the broker kept the connection until process exit, while `IsOpen` already reported `false`. Measured against a live broker, holding the process open without disposing:

| call | threw | elapsed | broker connection after |
|---|---|---|---|
| `CloseAsync(30s)` | — | 18ms | gone |
| `CloseAsync(6s)` | — | 19ms | gone |
| `CloseAsync(TimeSpan.Zero)` | `TaskCanceledException` | 6ms | **still listed** |
| `CloseAsync(TimeSpan.Zero)`, with the minimum | — | 19ms | gone |

`InternalConstants.MinConnectionCloseTimeout` is 1 second, and it is not the old floor in miniature: at 30 seconds the floor hid every caller's intent, while this one only stops a close from cancelling itself. #1973's headline complaint — a caller's 6 seconds being ignored — was already fixed and is unaffected.

## Public API

No signature changes; `PublicAPI.*.txt` is untouched. The XML documentation on `IConnection.CloseAsync` and the five `IConnectionExtensions` overloads was rewritten to describe what the code does.

## Testing

Solution builds clean on net8.0 and netstandard2.0, zero warnings. 211 `Unit` tests and 20 close/shutdown integration tests pass against a live broker.

`TestConnectionCloseTimeout` (15 tests) asserts against the resolution directly, because the difference is otherwise unobservable on a healthy connection — which is exactly why the floors went unnoticed. Two of them exist specifically because earlier versions could not fail:

- The ceiling test previously compared only against the ceiling constant, so halving it left the whole suite green. It now asserts the ceiling is *maximal* — one millisecond above it is rejected by `CancellationTokenSource` — which fails under that mutation.
- The negative-timeout theory described the wrong hazard. Measured: `(long)TotalMilliseconds` truncates toward zero, so `(-2ms, -1ms]` is **accepted and never cancels** (a silent unbounded close) while only `(-1ms, 0)` cancels immediately, and both original `InlineData` values *threw*. It is now parameterised in ticks with a case in each regime; the `-10001` tick case fails if the infinite exemption is loosened to `(long)TotalMilliseconds == -1`.

- **A graceful close timeout is now honoured as given, down to a one second minimum.** Values below 30 seconds were silently raised to 30 seconds; `CloseAsync(TimeSpan.FromSeconds(6))` now takes about 6 seconds. The minimum exists because the timeout also bounds the close handshake itself, not only the wait for the peer's reply: a smaller value cancels the close before it transmits anything, which leaves the connection open on the broker even though the client reports it closed. `CloseAsync(TimeSpan.Zero)` therefore resolves to one second rather than being taken literally.
- **`Timeout.InfiniteTimeSpan` is honoured as a graceful close timeout.** It is documented as the way to wait without a bound, but being negative it compared below the floor and was silently lowered to 30 seconds, so the documented behaviour was unreachable. Note what it means: the heartbeat timers are stopped before the wait begins and the socket read timeout does not apply to asynchronous reads, so in practice such a wait ends only when the broker notices our silence and drops the connection — which does not happen at all if heartbeats are disabled.
- **An abort is now always bounded, between 5 and 10 seconds.** It previously honoured any finite value however large, so `AbortAsync(TimeSpan.FromMinutes(2))` waited two minutes; it now waits at most 10 seconds. An abort is best-effort teardown that never throws, so returning promptly is the point.
- **A close timeout too large for the timer, including `TimeSpan.MaxValue`, is clamped to roughly 24.86 days** instead of throwing `ArgumentOutOfRangeException`. That exception was previously raised before the connection began shutting down, leaving it fully open. The bound no longer depends on which build of the library your application resolves.
- **A negative timeout other than `Timeout.InfiniteTimeSpan` takes the same one second minimum.** Such values are not durations, and passing them through reached a `CancellationTokenSource` that rejects most of them while silently treating anything in `(-2ms, -1ms]` as unbounded.


projects/RabbitMQ.Client/Impl/Connection.cs
```diff
@@ -509,6 +501,84 @@ await _frameHandler.CloseAsync(cts.Token)
...
+            }
+
+            return timeout;
+        }
+
+        /// <summary>
+        /// The largest delay <see cref="CancellationTokenSource"/> accepts; anything above
+        /// this throws <see cref="ArgumentOutOfRangeException"/>. Roughly 49.7 days.
+        /// </summary>
+        private static readonly TimeSpan MaxCancellationTokenSourceDelay =
```
gfoidl: Having this as (private) property is cheaper in this case.
* construction the TimeSpan on the fly is cheap
* avoids a memory access for the static readonly field (which potentially must evict something from the cache)

paulomorgado: There are still some conversions and calculation involved here.

There's no overload of [`TimeSpan.FromMilliseconds`](https://learn.microsoft.com/en-us/dotnet/api/system.timespan.frommilliseconds). Being compile-time constants, the C# compiler can take care of this. If they weren't, they would definitely be run-time constants, and the JIT compiler would take care of it.

However, creating a `TimeSpan` instance always ends up in [`TimeSpan(long ticks)`](https://learn.microsoft.com/en-us/dotnet/api/system.timespan.-ctor#system-timespan-ctor(system-int64)).

Why not just do it ourselves?

```csharp
TimeSpan.FromTicks((long.MaxValue & 0xFFFFFFFF) - 1 * TimeSpan.TicksPerMillisecond)
```

lukebakken: > [!NOTE]
> This reply was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who reviewed it before posting.

Thanks both. Leaving this as the static readonly field. It is computed once at type init and read once per connection close, a rare and I/O-bound path, so a property that recomputes `FromMilliseconds` on each call would do marginally more work rather than less, and the cache-eviction cost is not really reachable here.

lukebakken: > [!NOTE]
> This reply was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who reviewed it before posting.

Following up: a later review pass caught that the limit was miscalibrated across target frameworks, independent of the field-vs-property and `FromTicks` points above. `CancellationTokenSource(TimeSpan)` caps at `int.MaxValue` ms (roughly 24.86 days) on .NET Framework (the netstandard2.0 target), lower than modern .NET's roughly 49.7 days, so `uint.MaxValue - 1` ms let a 25 to 49 day timeout past the guard into an `ArgumentOutOfRangeException`. Corrected to `TimeSpan.FromMilliseconds(int.MaxValue)`, which both runtimes accept, with a roughly 30-day test case added. The field and `FromMilliseconds` shape is otherwise unchanged.

gfoidl: > read once per connection close

Even more an argument pro re-computing 😄 and not caching it in a field.

> cache-eviction cost is not really reachable here

Well, only under the assumption that "this thing" is the only "thing" that the machine does. A CPU has other things also to do, so viewing that from a bigger altitude it has a cost.

Further the static readonly field needs to get boxed in order to live on the heap, and adds to the memory footprint.

Sure all points won't move the needle in the one or other direction very much, but they still can sum up.

projects/RabbitMQ.Client/Impl/Connection.cs
```diff
@@ -573,11 +573,22 @@ internal static TimeSpan ResolveCloseTimeout(TimeSpan timeout, bool abort)
...
+        /// with <see cref="ArgumentOutOfRangeException"/>, so it uses that floor. A runtime that
+        /// accepts more - .NET Core and later allow up to <c>uint.MaxValue - 1</c> milliseconds,
+        /// roughly 49.7 days - still treats a delay in the gap as unbounded, a harmless
+        /// over-approximation for so extreme a value. The net8.0 build uses the larger limit.
         /// </summary>
+#if NETSTANDARD
+        private static readonly TimeSpan MaxCancellationTokenSourceDelay =
+            TimeSpan.FromMilliseconds(int.MaxValue);
+#else
         private static readonly TimeSpan MaxCancellationTokenSourceDelay =
```
gfoidl: Opposed to Claude I still think this should be a property, not a static readonly field. 
Cf. https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/2011#discussion_r3922743163

lukebakken: This is noted, thank you, but the intent of `static readonly` makes it clear that `MaxCancellationTokenSourceDelay` is a constant. If this code were on the hot path, or had demonstrable negative performance or impact, I would reconsider.

projects/RabbitMQ.Client/Impl/Connection.cs
lukebakken: > [!NOTE]
> This comment was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who reviewed it before posting.

Correcting my follow-up above: the ceiling is not a single `TimeSpan.FromMilliseconds(int.MaxValue)` for both runtimes. It is selected per target framework — `int.MaxValue` ms (~24.86 days) under `#if NETSTANDARD`, `uint.MaxValue - 1` ms (~49.7 days) for `net8.0` — because only the netstandard2.0 build can load on .NET Framework, which is what caps at the lower value. The code, the XML docs and the release note all describe it that way; only my summary of the change was wrong.

Issue #1995, opened 2026-08-04
Custom TopologyRecoveryExceptionHandler bypasses the topology recovery retry classification


lukebakken:
> [!NOTE]
> This issue was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken. The code references were verified against the #1994 branch; the options listed below are AI-drafted starting points, not settled decisions.

A custom `TopologyRecoveryExceptionHandler` bypasses the retry classification added in #1994, so the #1993 bug persists for anyone who installs one.

In `RecoverConsumersAsync` (`projects/RabbitMQ.Client/Impl/AutorecoveringConnection.Recovery.cs:617-638`) the catch block picks one of two paths:

```csharp
if (_config.TopologyRecoveryExceptionHandler.ConsumerRecoveryExceptionHandlerAsync != null
    && _config.TopologyRecoveryExceptionHandler.ConsumerRecoveryExceptionCondition(consumer, ex))
{
    // custom handler runs; exception is considered handled
}
else
{
    HandleTopologyRecoveryException(...);   // where ShouldRetryRecoveryAfter is consulted
}
```

`ShouldRetryRecoveryAfter` is only reached on the `else` branch. When a handler is installed the exception is swallowed and recovery continues to report success, which is exactly the #1993 failure mode: the broker has registered the consumer, the client never wired it into the new channel's `ConsumerDispatcher`, and deliveries go to the `FallbackConsumer` and are dropped unacked.

The default conditions make this maximally broad. All four in `projects/RabbitMQ.Client/TopologyRecoveryExceptionHandler.cs:11-14` are `(x, ex) => true`:

```csharp
private static readonly Func<IRecordedExchange, Exception, bool> s_defaultExchangeExceptionCondition = (e, ex) => true;
private static readonly Func<IRecordedQueue, Exception, bool> s_defaultQueueExceptionCondition = (q, ex) => true;
private static readonly Func<IRecordedBinding, Exception, bool> s_defaultBindingExceptionCondition = (b, ex) => true;
private static readonly Func<IRecordedConsumer, Exception, bool> s_defaultConsumerExceptionCondition = (c, ex) => true;
```

So installing a handler without also narrowing its condition routes *every* exception to it, including continuation timeouts that the client should retry. A handler written to deal with, say, a queue that fails an equivalence check now also intercepts timeouts it has no way to fix.

[...]

PR #2012, opened 2026-09-03
Pass the recovery cancellation token when recovering a consumer

### LINKED ISSUE #1997: RecordedConsumer.RecoverAsync does not pass a cancellation token
> [!NOTE]
> This issue was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken. The code references were verified against the #1994 branch; the suggested direction below is an AI-drafted starting point, not a settled decision.

`RecordedConsumer.RecoverAsync` (`projects/RabbitMQ.Client/Impl/RecordedConsumer.cs:115-118`) does not pass a cancellation token:

```csharp
public Task<string> RecoverAsync(IChannel channel)
{
    return channel.BasicConsumeAsync(Queue, AutoAck, ConsumerTag, false, Exclusive, Arguments, Consumer);
}
```

`IChannel.BasicConsumeAsync` accepts `CancellationToken cancellationToken = default` (`projects/RabbitMQ.Client/IChannel.cs:181-183`), so the default is `CancellationToken.None`.

`RecoverConsumersAsync` has a `cancellationToken` in scope and uses it elsewhere in the same method, for `_recordedEntitiesSemaphore.WaitAsync` and for the `RecoveringConsumerEventArgs` it constructs. The call at line 597 is the one place in that loop that does not receive it.

The consequence is bounded: the `basic.consume` still ends after `ContinuationTimeout` because the continuation has its own timeout token, so nothing hangs indefinitely. But when recovery is being torn down (connection closing), the in-flight `basic.consume` is not cancelled promptly and shutdown waits out the continuation timeout instead.

This is also what makes the `ShouldRetryRecoveryAfter` discriminator in #1994 necessary in its current form: since the recovery token is not plumbed into the RPC, the code has to consult `recoveryCancellationToken.IsCancellationRequested` separately to tell "recovery is shutting down" from "this operation timed out".

Threading the token through looks straightforward, but it changes which exception arrives on the shutdown path, so the #1994 classification and its unit tests in `projects/Test/Unit/TestTopologyRecoveryExceptionClassification.cs` should be re-checked alongside it. Worth verifying whether other `Recorded*.RecoverAsync` methods have the same gap.

lukebakken:
> [!NOTE]
> This pull request was prepared by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who reviewed it before filing. The code was written and verified jointly.

## Summary

`RecordedConsumer.RecoverAsync` did not take a cancellation token, so its `basic.consume` ran with `CancellationToken.None`, while the three other recorded entities (exchange, queue, binding) all received the recovery token. When recovery is being torn down the in-flight `basic.consume` was therefore not cancelled promptly, and shutdown waited out the full `ContinuationTimeout` instead. Fixes #1997.

Verified that `RecordedConsumer` was the only gap of the four, and that the only call site is `RecoverConsumersAsync`, which already had the token in scope and used it for the surrounding semaphore waits and event arguments. The token is `_recoveryCancellationTokenSource.Token`, cancelled only by `CloseAsync` / `DisposeAsync`, so the new cancellability cannot fire during a normal recovery.

## The #1994 retry classification is unaffected

The issue asked for this to be re-checked, so to be explicit: `ShouldRetryRecoveryAfter` still needs its token-based discriminator, and its unit tests pass unchanged. A protocol operation that runs past `ContinuationTimeout` still completes its continuation as cancelled, so an `OperationCanceledException` remains ambiguous by type alone. Threading the token only makes a torn-down recovery reach that check promptly instead of after the continuation timeout. Both branches stay correct: recovery cancelled means do not retry, recovery not cancelled means the operation timed out and must be retried.

## Testing

The consumer fix itself is enforced by the compiler, since dropping the parameter again breaks its only call site. What nothing enforced is a *newly added* entity repeating the mistake, so `TestRecordedEntityRecovery` pins the set of recovery methods and requires each to accept the token, in the same style as the existing `SemaphoreSlim` field test that makes adding a field a deliberate act. It fails against the unfixed code. That limit - it cannot detect a method that accepts the token and then ignores it - is now closed by a second test; see the review follow-up below.

Local runs on net8.0: 225/225 unit, 47/47 integration across `ConnectionRecovery`.

## Related hazard - now fixed here, contrary to what this section originally said

All four entity-recovery loops release `_recordedEntitiesSemaphore`, await application code, then re-acquire it in a `finally`, and every re-acquire passed the recovery token. If that token is already cancelled the re-acquire throws, the `finally` completes without holding the semaphore, and the outer `finally` in `TryPerformAutomaticRecoveryAsync` releases one it does not hold, throwing `SemaphoreFullException` on a `SemaphoreSlim(1, 1)` - which replaces the real cancellation and turns an orderly shutdown into a 541.

This section originally said the hazard was pre-existing, out of scope, and belonged with #1995. **It is fixed on this branch**, at all seven re-acquire sites, which pass `CancellationToken.None`. Fixing one and leaving six would have read as a fix while leaving the same failure reachable from the other entity types.

- Closing a connection while automatic recovery is invoking one of your topology-recovery callbacks no longer reports the shutdown as an internal error caused by `SemaphoreFullException`. Recovery releases an internal semaphore around each callback and re-acquired it with the recovery token, so once a close had begun the re-acquire threw, the semaphore was released twice, and the resulting `SemaphoreFullException` replaced the real cancellation. This affected the `RecoveringConsumerAsync` event and every `TopologyRecoveryExceptionHandler` callback. Note that a connection closed mid-recovery can still be reported with reply code 541 for a separate reason, tracked as **#2037**; this change removes the semaphore path only.

## Review follow-up (2026-09-12)

Two max-effort passes over this branch. Both found the production changes correct and my descriptions of them wrong, which is why the two sections above carry corrections rather than just additions.

**The signature test's stated gap is closed.** `RecordedConsumerActuallyPassesTheTokenToBasicConsume_GH1997` drives `RecordedConsumer.RecoverAsync` against a `DispatchProxy` `IChannel` that captures the token `basic.consume` was actually handed, so keeping the parameter and passing `CancellationToken.None` in the body - a one-word change that left the whole suite green before this test existed - now fails. It asserts `basic.consume` was reached at all, so it cannot pass vacuously, and any other `IChannel` member throws rather than returning a default. `net8.0` only: `DispatchProxy` does not exist on `net472`.

That test originally used `Assert.Equal(cts.Token, recorder.CapturedToken)`, whose failure said nothing - `CancellationToken` does not override `ToString()`, so xUnit printed `System.Threading.CancellationToken` for both expected and actual. Measured, then replaced with an `Equals` assertion whose message names the guarded mutation and prints `CanBeCanceled` and whether the captured token is `CancellationToken.None`. Under the mutation it now reports `CanBeCanceled=False, is CancellationToken.None=True`.

**A pre-existing deadlock was found while proving the untokenised waits safe, and is filed as #2038.** `TryPerformAutomaticRecoveryAsync` holds `_recordedEntitiesSemaphore` across `RecoverChannelsAndItsConsumersAsync`, which reaches `Channel.RunRecoveryEventHandlers` and invokes user channel-recovery handlers **while the semaphore is held**. A handler that declares an exchange - or calls `BasicConsumeAsync`, which also records - waits on that same semaphore with no token and never returns. Reproduced on `main`: the handler stayed blocked past 25 s, the connection kept reporting `IsOpen == true`, `RecoverySucceededAsync` never fired, and application topology calls on unrelated channels blocked indefinitely. It cannot deadlock this branch's re-acquires, because it is all one sequential task.

**Prose corrections, all in comments and docs rather than code.** The internal doc still said the semaphore hazard was "not fixed by #1997" while this branch fixes all seven sites - the sentence a maintainer could have cited to revert it. The comment justifying an untokenised wait said "the only other holder is this same recovery loop"; there are ten others, all in `AutorecoveringConnection.Recording.cs`, and the conclusion survives only because each holds the semaphore across in-memory dictionary work alone. "Two waits are deliberately still cancellable" then listed three, and called cancelling the initial acquisition "correct and cheap" when it reaches the same catch-all and the same 541. A claim that the handler deadlock "runs strictly after the consumer loop" was false, since the channel loop runs each channel's handlers before the next channel's consumers. And `AutorecoveringChannel.RunRecoveryEventHandlers` does not exist - it is declared in `Impl/Channel.cs`.


lukebakken:
> [!NOTE]

Splitting out the residual half of this, as #2037.

This PR removes one of two independent routes to a misreported close. A cancelled re-acquire of `_recordedEntitiesSemaphore` raised `SemaphoreFullException`, which replaced the real cancellation and surfaced as a 541 `FailedAutoRecovery`; all seven re-acquire sites now pass `CancellationToken.None`.

The second route is untouched here and is #2037: the two `_channelsSemaphore.WaitAsync(cancellationToken)` calls in `RecoverChannelsAndItsConsumersAsync` throw on the same cancellation, and `TryPerformAutomaticRecoveryAsync`'s catch-all cannot tell that from a recovery failure, so it logs an error and aborts with `InternalError`. Fixing that needs a decision about how cancellation should be reported rather than a token swap, which is why it is not in this PR.

Two consequences worth recording here:

- **The semaphore fix in this PR has no regression test, and #2037 is where it gets one.** A deterministic test is achievable - the recovery callbacks are handed the recovery token itself, so a handler can block on it until cancellation, with no sleeps or races - but the clean assertion is "the close reports 200, not 541", and that still fails while the second route exists. #2037 therefore carries a single test covering both routes.
- **#2037 must land after this PR** for the same reason. Its code change is independent and in a different region, so there is no conflict either way; only the test depends on both.

PR #2041, opened 2026-09-15
Run channel recovery handlers outside the entities semaphore

### LINKED ISSUE #2038: Recovery handler that touches topology deadlocks recovery permanently
> [!NOTE]
> This issue was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who reviewed it before filing. The deadlock and its blast radius were reproduced against `main` at `d039f07b5` with a live broker, not inferred; every line reference below was read in that tree. The remediation options are AI-drafted starting points, not settled decisions.

A channel-recovery handler that declares topology or registers a consumer deadlocks automatic recovery permanently. The connection then reports `IsOpen == true` while every topology operation on it blocks forever, with no token, no timeout and no exception.

## Mechanism

`TryPerformAutomaticRecoveryAsync` acquires `_recordedEntitiesSemaphore` (`Impl/AutorecoveringConnection.Recovery.cs:327`) and holds it across the whole of topology recovery, including the call to `RecoverChannelsAndItsConsumersAsync(recordedEntitiesSemaphoreHeld: true, ...)` at `:348`.

That loops over the recovered channels calling `AutorecoveringChannel.AutomaticallyRecoverAsync`, which reaches `await _innerChannel.RunRecoveryEventHandlers(this, cancellationToken)` (`Impl/AutorecoveringChannel.cs:214`; the method is declared at `Impl/Channel.cs:159`). So **user recovery handlers are invoked with `_recordedEntitiesSemaphore` held.** The handler is reached at all because `Channel.TakeOver` carries the subscription across the channel swap (`Impl/Channel.cs:199`, `_recoveryAsyncWrapper.Takeover(...)`).

A handler that declares an exchange then calls `AutorecoveringChannel.ExchangeDeclareAsync`, which does the broker round trip first (`:432`) and afterwards records the entity:

```csharp
// Impl/AutorecoveringChannel.cs:437
await _connection.RecordExchangeAsync(recordedExchange, recordedEntitiesSemaphoreHeld: false)
```

`recordedEntitiesSemaphoreHeld: false` sends it to `Impl/AutorecoveringConnection.Recording.cs:67`:

```csharp
await _recordedEntitiesSemaphore.WaitAsync()
```

**No token and no timeout.** The holder is the recovery loop that is currently awaiting this very handler, so the wait never completes. Note the AMQP `exchange.declare` has already succeeded - the broker is fine; the client wedges on its own bookkeeping afterwards.

**This is not specific to exchanges, and it is not limited to topology declaration.** `AutorecoveringChannel` has 14 call sites passing `recordedEntitiesSemaphoreHeld: false`, covering exchange, queue and binding declare and delete, the auto-delete cleanups - and `BasicConsumeAsync` (`:336`) and `BasicCancelAsync` (`:321`). **Registering a consumer from a recovery handler deadlocks in exactly the same way**, verified below, and that is a far more common thing to do from `RecoveryAsync` than re-declaring an exchange.

## Reproduction

Register a channel `RecoveryAsync` handler that declares an exchange on the sender, then force a drop with `rabbitmqctl close_all_connections`:

```csharp
((IRecoverable)ch).RecoveryAsync += async (sender, ea) =>
{
    // Never returns.
    await ((IChannel)sender).ExchangeDeclareAsync("probe-from-handler", ExchangeType.Direct, false, true);
};
```

Observed (`NetworkRecoveryInterval` 500 ms, so nothing here is a slow-recovery artefact):

```
[handler] entered channel RecoveryAsync; declaring an exchange
RESULT: DEADLOCK CONFIRMED - handler still blocked after 00:00:25.49
```

Replacing the body with a consumer registration on a pre-existing durable queue reproduces identically:

```csharp
((IRecoverable)ch).RecoveryAsync += async (sender, ea) =>
{
    var c = new AsyncEventingBasicConsumer((IChannel)sender);
    c.ReceivedAsync += (a, b) => Task.CompletedTask;
    await ((IChannel)sender).BasicConsumeAsync(qname, true, c);   // never returns
};
```

```
[handler] entered channel RecoveryAsync; calling BasicConsumeAsync
RESULT: DEADLOCK CONFIRMED - handler still blocked after 00:00:15.51
```

## Blast radius: worse than a stalled recovery

Because the recovery task never returns, the semaphore is never released, and it guards all recorded-entity access - not just recovery. Measured after the handler wedged:

| observation | result |
|---|---|
| `conn.IsOpen` | **`true`** |
| `RecoverySucceededAsync` fired | **never** |
| application `ExchangeDeclareAsync` on a *new* channel | **blocked, >15 s** |
| application `QueueDeclareAsync` on a *new* channel | **blocked, >15 s** |

So the failure is silent and total: the connection presents as healthy, no recovery-failure event is raised, nothing times out, and the application's own topology calls hang indefinitely on unrelated channels. There is no way for a caller to detect or escape it, because the wait accepts no token. Only tearing down the process clears it.

## What is and is not affected

Safe from a recovery handler: anything that does not record or un-record an entity - publishing, acking, `BasicQosAsync`, reading properties, logging, signalling other application code.

Deadlocks: any of the 14 recording operations. That is every exchange, queue and binding declare or delete, plus `BasicConsumeAsync` and `BasicCancelAsync`. Both of the two most plausible uses of this event - re-declare my topology, re-attach my consumer - fall inside that set.

So the reason this has gone unnoticed is probably not that the trigger is exotic. It is that a handler which hangs silently, while `IsOpen` stays `true` and no failure event fires, is easy to misread as "recovery is slow" or to blame on the broker.

## Relationship to other issues

- **#1997 / PR #2012** touches the same semaphore, changing the seven *re-acquire* sites in the recovery loops to `CancellationToken.None`. That is a different hazard (a cancelled re-acquire escaping without the semaphore and causing a `SemaphoreFullException`) and neither causes nor fixes this. This one is pre-existing on `main` and independent.
- **#2020** tracks the dispose-versus-recovery race on `_disposed` / `_innerChannel`. Related in that both are about recovery running without synchronization against another actor, but a different defect.
- **#2016** tracks unsynchronised state on `AsyncDefaultBasicConsumer`. Unrelated mechanism.

## Remediation options

1. **Do not hold the semaphore across user code.** Release `_recordedEntitiesSemaphore` before `RunRecoveryEventHandlers` and re-acquire after, which is the pattern the entity loops already use for their own callbacks. Consistent with existing practice and fixes the root cause, but it widens the window in which recorded state can change mid-recovery, and PR #2012 exists precisely because that release/re-acquire pattern has its own sharp edges.
2. **Make the recording waits reentrant for the recovery task.** An `AsyncLocal<bool>` or an explicit "already held" flag threaded through, so a call originating inside the recovery task takes the synchronous branch that already exists for `recordedEntitiesSemaphoreHeld: true`. Narrow and preserves the current locking, but adds ambient state, and a handler that hands work to another thread would defeat it.
3. **Invoke recovery handlers after recovery completes**, outside the semaphore. Arguably the more correct contract anyway - a handler told "this channel recovered" reasonably expects recovery to be finished - but it changes observable ordering for existing handlers.
4. **Bound the wait.** Give the `Recording.cs` waits a timeout and throw. Turns a permanent silent hang into a diagnosable error, which is a real improvement, but leaves the design defect and makes a legitimate handler fail.

Option 1 or 3 addresses the cause; 3 is the smaller change and the better contract, so it looks the most promising, but the ordering change needs a compatibility call that is @lukebakken's rather than mine.

## Testing

A regression test is cheap and deterministic here, unlike most recovery races: register a handler that calls `BasicConsumeAsync` - the likelier real-world trigger - drop the connection, and assert the handler returns within a bounded time. It fails reliably today by hanging, so the assertion needs a timeout rather than relying on the test framework. `projects/Test/Integration/ConnectionRecovery/` already builds the drop-and-recover harness. Worth also asserting that `RecoverySucceededAsync` fires, since the current failure suppresses it.

lukebakken:
> [!NOTE]
> This pull request was written by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who reviewed it before it was opened. Every measurement below was taken against a live broker in this branch, not inferred; the mutation results are individual test runs.

## Proposed Changes

Fixes #2038. A channel `RecoveryAsync` handler that declared topology, registered or cancelled a consumer, or closed its channel deadlocked automatic recovery **permanently**, in every 7.x release from `v7.0.0` to `v7.2.2`.

`TryPerformAutomaticRecoveryAsync` holds `_recordedEntitiesSemaphore` across topology recovery, and the handler invocation sat two frames below it, inside `AutorecoveringChannel.AutomaticallyRecoverAsync`. Recording an entity re-acquires that semaphore, and **no caller can supply a token to that wait**: five of the recording methods take none at all, and the other five receive one from the public channel method the handler called, where it defaults to `default`. So the handler waited on a holder that was waiting on the handler, with nothing to interrupt it.

The failure was silent and total. Measured after a handler wedged: `IsOpen` still `true`, `RecoverySucceededAsync` never fired, and `ExchangeDeclareAsync` on a brand-new channel still blocked when the probe gave up - that semaphore guards all recorded-entity access, not just recovery.

**The fix:** `RecoverChannelsAndItsConsumersAsync` returns the channels that recovered, and `TryPerformAutomaticRecoveryAsync` invokes their handlers after the semaphore is released and before `RecoverySucceededAsync`. Every *asynchronous* callback `Recovery.cs` invokes itself already followed that rule, so this restores the file's own pattern rather than inventing one. 23 lines of production code.

## Behaviour changes

- Handlers fire once **every** channel has recovered, rather than interleaved per channel. See the review request below.
- A later channel failing with an exception defers the earlier channels' handlers to the attempt that succeeds.
- A channel closed but not disposed during the loop still gets its handler; its topology calls throw `AlreadyClosedException` into `CallbackExceptionAsync`. Newly reachable, since closing another channel from a handler used to deadlock.
- Unchanged: a channel's consumers are still recovered before its handler runs, and handler exceptions still reach `CallbackExceptionAsync` rather than aborting recovery.

## Evidence

Each row is a test run against a live broker:

| variant | consumer-in-handler test | all-channels-recovered test | pre-existing `_Called` test |
|---|---|---|---|
| this branch | pass | pass | pass |
| handler loop moved back inside the semaphore | FAIL (30 s timeout) | pass | pass |
| original design restored | FAIL | FAIL | pass |

The two mutations fail for different reasons, so neither test is a restatement of the other. The pre-existing `TestRecoveryEventHandlers_Called` passes under both mutations - it registers an empty handler and counts invocations, which is why this survived four releases.

Also run: `FullyQualifiedName~Recovery` 61 passed / 1 skipped (Toxiproxy container not started), Unit 223 passed, both target frameworks compile with 0 warnings. No `PublicAPI` churn - the new member is internal.

## Review request: the ordering change

@zvirja - you may want to look at this one, since it is the same class of bug you fixed in #2005: an external handler invoked while the client holds a lock. Your argument there that handler *ordering* was not something callers should have relied on is the argument this change needs too, so your view on the ordering change specifically would be valuable.

Concretely: channel `RecoveryAsync` handlers now fire once every channel has recovered, instead of one at a time as each channel recovers. EasyNetQ is the one dependent found on that event - `PersistentChannel.cs` pattern-tests every channel against `IRecoverable` and throws `NotSupportedException` if the test fails, then publishes `ChannelRecoveredEvent` from inside the handler. Nothing in that handler touches the semaphore, so EasyNetQ itself never deadlocked; the question is whether anything downstream of `ChannelRecoveredEvent` assumes it fires mid-recovery rather than after it.

## Notes for reviewers

`docs/internal/recovery-event-handler-invocation.md` carries the rule and the reasoning, so the call site stays short.

**Two things deliberately left out of scope**, both pre-existing:

- Eight *synchronous* user predicates (the four `TopologyRecoveryFilter` funcs and four `TopologyRecoveryExceptionHandler` conditions) still run with the semaphore held. They can only deadlock via sync-over-async, so this is narrower, but it is the same hang. Named in the internal doc.
- Calling the *connection* `CloseAsync` from a handler stalls for `RequestedConnectionTimeout` and then completes, because the handler runs on the recovery task that `StopRecoveryLoopAsync` waits on. Unrelated to the semaphore.

`git merge-tree` against #2012 is clean and the hunks are disjoint. Whichever merges second: #2012's comment describing #2038 as "a pre-existing deadlock that does not reach it" will want a light touch-up.

## Release notes

Fixed a permanent deadlock in automatic connection recovery: a channel recovery handler that declared topology, registered or cancelled a consumer, or closed its channel would wedge recovery forever, while the connection still reported itself open and no recovery-failure event fired.


projects/RabbitMQ.Client/Impl/AutorecoveringConnection.Recovery.cs
```diff
@@ -761,7 +773,7 @@ void UpdateConsumer(string oldTag, string newTag, in RecordedConsumer consumer)
             }
         }
 
-        private async ValueTask RecoverChannelsAndItsConsumersAsync(bool recordedEntitiesSemaphoreHeld, CancellationToken cancellationToken)
+        private async ValueTask<List<AutorecoveringChannel>> RecoverChannelsAndItsConsumersAsync(bool recordedEntitiesSemaphoreHeld, CancellationToken cancellationToken)
```
danielmarbach: Given this is not perf sensitive code would it be better to return IReadOnlyCollection?

projects/RabbitMQ.Client/Impl/AutorecoveringConnection.Recovery.cs
```diff
@@ -761,7 +773,7 @@ void UpdateConsumer(string oldTag, string newTag, in RecordedConsumer consumer)
...
         {
             if (false == recordedEntitiesSemaphoreHeld)
```
danielmarbach, line 809: Given the recorded semaphore must be held when entering this method, what benefit does the `_channelsSemaphore` provide given it is only ever used in this method?

Also, if that holds why do we then need `notRecoveredChannels` given we could simply remove them inline directly?

But I might not be seeing this clearly...

lukebakken:
> [!NOTE]

Thanks @danielmarbach.

- `IReadOnlyCollection`: done.
- `_channelsSemaphore` is load-bearing - `CreateChannelAsync` reaches `RecordChannelAsync` (`AutorecoveringConnection.cs:294`) without ever taking `_recordedEntitiesSemaphore`, so an application thread can `List.Add` to `_channels` while recovery snapshots it.
- `notRecoveredChannels`: you are right it is not necessary. But since the semaphore *is* needed, inline removal costs an acquire/release per failure inside the loop, so it is not actually simpler. Left as-is; happy to change it if you disagree.
- Your question found a real gap: nothing tested that removal. Added `TestChannelDisposedWhileConnectionIsDownIsDroppedFromRecovery` - a channel disposed while the connection is down stays recorded, and recovery is the only thing that drops it. Fails in 10s if the removal is suppressed.

danielmarbach (dismissed):

